PR #133, opened 2013-03-12
Add augeas() function

raphink:
This PR adds a new `augeas()` function, which does on the master side pretty much what the `augeas` type does on the client side.

_Example:_

```
augeas("proc        /proc   proc    nodev,noexec,nosuid     0       0\n", 'Fstab.lns', ['rm ./1/opt[3]'])
```

Would result in:

```
"proc        /proc   proc    nodev,noexec     0       0\n"
```
- _Type_: rvalue

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -60,6 +60,21 @@ a single integer and float value as an argument.
 
 - *Type*: rvalue
 
+augeas
+------
+Modifies a string using Augeas.
+
+*Example:*
+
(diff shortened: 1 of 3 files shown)


ColinHebert:
@raphink, it's an old issue, but I wonder whether it is worth reopening now that augeas is part of the puppet distribution.

cc @jeffmccune @adrienthebo

raphink:
If you find the function useful, sure.

raphink:
Also, lambda functions could probably make this much nicer now.

PR #223, opened 2014-02-26
MODULES-467 Function file_join for smart path joining

k8r-io:


DimanNe:
It was not "*a little too specific and obscure for most people*". - I have been using puppet for 1.5 days and this PR **is one of the first things I found**.

Without a function like this you will either get paths like `/a//////b` (because "every step" will add an `/` just in case), or paths like `/ab` (if someone wanted to avoid redundant `/`).

DimanNe:
Nice! Thanks!

PR #249, opened 2014-05-07
Set arity on all functions

dalen:
Use the function arity feature to remove a lot of boilerplate code from
the functions.

take 2, now that puppet 2.x support is deprecated (again kinda)


hunner:
Given that the puppet 4 function api dispatches take care of arity and puppet 3 will eventually be EOL'd, I'm going to just close this... when we drop puppet 3 we'll convert all functions to the 4 api.

PR #259, opened 2014-05-17
to_yaml function added.

rdark:
* Makes use of ZAML.dump


cyberious:
@rdark please update to use YAML instead of ZAML.

mhaskel:
Hi @rdark, thanks for the contribution! I'm going to close this PR for now since we haven't heard from you in a while. If this is still something you're interested in having added to the module please address the concerns in the comments and resubmit. Thanks again!

PR #260, opened 2014-05-19
added a truncate function

mburger:


mhaskel:
Hi @mburger, thanks for the contribution! Since it looks like this can be accomplished with `regsubst()`, and we haven't heard from you in a while, I'm going to close this PR for now. If you'd still like to get the `truncate()` function added to stdlib and you want to address the issues in the comments, we'd be happy to review it again. Thanks again!

PR #272, opened 2014-06-16
add a filter_hash function

roidelapluie:


hunner:
I think this can be done in puppet 4 https://docs.puppetlabs.com/references/latest/function.html#filter

PR #275, opened 2014-06-17
Support hashes in prefix() function

roidelapluie:


hunner:
Is there a reason that this does keys instead of values? It seems to me that prefixing values would be way more common.

Perhaps hashes should have their own way of prefixing things separate from the `prefix()` function

roidelapluie:
Because I needed it.

everestx:
I specifically need it to prefix keys when I am using create_resources() on a common hash data structure that gets applied to multiple different hosts. As an example consider a hash of nagios_service() resources that we define in hiera to apply to all hosts. Then those resources get exported to a monitoring host. At the time of export all is good. However at the time of collection if we don't prefix all the resource titles with the exported hostname, we'll get duplicate resource errors.
Note that future parse might alleviate this concern with new map functions. However, we just implemented our own custom function for the time being.

mhaskel:
@roidelapluie @everestx ok, we merged this in #420. Thanks!

PR #304, opened 2014-07-21
Add new function `deep_merge`.

riton:
* Add rspec code to test new functionnality (didn't test the raise of an error if no _deep_merge_ module is available)
* Add documentation to README
* Add example in tests/deep_merge.pp

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -20,6 +20,7 @@ group :development, :test do
   gem 'simplecov',               :require => false
   gem 'beaker',                  :require => false
   gem 'beaker-rspec',            :require => false
+  gem 'deep_merge',              :require => false
 end
 
 ENV['GEM_PUPPET_VERSION'] ||= ENV['PUPPET_GEM_VERSION']
--- a/README.markdown
+++ b/README.markdown
@@ -162,6 +162,22 @@ If called with only an array it counts the number of elements that are not nil/u
 
 - *Type*: rvalue
 
+deep_merge
+----------
+Deep merges two or more hashes together using gem *deep_merge* and returns the resulting hash.
+
+For example:
+
+    $hash1 = {'one' => 1, 'two' => [1]}
+    $hash2 = {'two' => [2], 'three' => 'tres'}
+    $merged_hash = merge($hash1, $hash2)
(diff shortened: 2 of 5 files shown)


cyberious:
@riton I am not a fan of requiring a gem to be installed in order for the functionality to work, especially as in a masterless scenario this will fail over on Windows, I understand the use case and reasons but not the implementation to fail hard when the gem is unable to be loaded

DavidS:
1) this needs rebasing against a current version of stdlib.
2) given the massive change in behavior, it would be much better if this were a separate function, so people can make their own decisions/tradeoffs. I would suggest `deep_merge_gem` or similar.
3) a new function can significantly reduce complexity and test effort by just passing through all arguments to the gem and just checking for the right function to be called.

mhaskel:
@riton ping?

riton:
Here I'm guys, sorry for the delay !

riton:
So I'm a little confused here.
@cyberious said that having a dependency on an external _gem_ wasn't a good thing.
I understood that you wasn't interested in this _pull request_ anymore.

And 2 weeks ago, @DavidS seems to say that this needs work, but seems a good idea after all.

Are you still interested in this function ?

DavidS:
@riton both issues need to be addressed before the functionality can be merged. I pointed out a way to do so: create a new function, thereby not breaking backwards compatibility, and creating a place where it can be described in detail what needs to be done to use this function.

DavidS:
In the "worst" case, this might be a candidate for https://github.com/puppet-community/puppet-extlib

nibalizer:
We welcome speculative functions into extlib so that they can be collected and perused and used. We've not had any functions promote out of extlib into stdlib, but that would be a huge success I think.

igalic:
:+1:

cyberious:
@riton I am not a fan of gem dependencies but if it is called out explicitly from the function name at least it is a little more clear that this gem MUST be installed.

bmjen:
Thanks for the contribution @riton. I think the consensus here is that this function would fit best in https://github.com/puppet-community/puppet-extlib

PR #305, opened 2014-07-31
new rvalue function: create_password. Please read embeded doc

cjeanneret:


cyberious:
Based on the comments I would like to see this function but would appreciate an acceptance test in regards to idempotent. Also requires rebase

elyscape:
This has some other issues. Will comment with details when I'm at a computer.

hunner:
Please generate your passwords outside of puppet, then add the generated hashes to puppet or hiera.

Also, there is no place to pass an actual password, and if you use a hash for an unknown password then it very well could be a hash for "password"... you never know. Security is hard.

PR #307, opened 2014-08-04
Add string_rand() function

yasn77:
`string_rand()` behaves mostly like `fqdn_rand(`). The difference being the
`fqdn` is not used as a seed to generate a random number, rather a string
(or integer) argument is used.

Since `fqdn_rand()` provides a random number for a single host, this function can be useful in providing a random number that can be used as an identifier for clustered services (across multiple hosts).


mhaskel:
Hi @yasn77,

Thanks for the contribution! We've been discussing this contribution, but we don't like the implementation. We don't see the value between having deterministic random numbers (rather than generate a random number you could just use the seed), and being deterministic there's no added security through having it be 'random'.

Thanks again though!

PR #314, opened 2014-09-02
Check if file exists before loading with loadyaml. If not, return nil

amateo:
In ```loadyaml``` function, check if file exists before trying to import. If not exists, return nil.

### DIFF
--- a/lib/puppet/parser/functions/loadyaml.rb
+++ b/lib/puppet/parser/functions/loadyaml.rb
@@ -13,7 +13,12 @@ module Puppet::Parser::Functions
       raise Puppet::ParseError, ("loadyaml(): wrong number of arguments (#{args.length}; must be 1)")
     end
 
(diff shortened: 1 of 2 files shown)


mhaskel:
Hi @amateo, thanks for the contribution, and apologies it took so long for us to review this. Do you think you could update tests in https://github.com/puppetlabs/puppetlabs-stdlib/blob/master/spec/functions/loadyaml_spec.rb to check the case where the file exists? Thanks!

amateo:
I have added a test for the case when the file does not exist. The case in when it exists, it is already included in existent tests

elyscape:
> I have added a test for the case when the file does not exist.

That doesn't seem to be included in this PR.

amateo:
Sorry, I forgot to push the changes to github repository. I think they are now.

mhaskel:
Great, thanks @amateo !

PR #318, opened 2014-09-12
jstringtohash function

WhatsARanjit:
Before structured facts are adopted, some facts in EC2 are returning "JSON" objects which are actually Strings.  This function allows you to safely convert them into actual data objects.


cyberious:
Closing PR for the above mentioned comments.  Also the functions do not meet the naming conventions.  I would have suggested convert_json_to_hash

PR #325, opened 2014-09-23
loadyaml function should be able to read files from module path

amuraru:


tedivm:
Obviously this isn't passing tests, but this really should be how this works.

nibalizer:
I have a hard time groking exactly what this is doing. However it seems like what it's trying to do is fall back to opening a file existing in the current module's directory. Couldn't this be accomplished by composing get_module_path() and load_yaml()?

tedivm:
That's pretty much how I handled it, but I feel this is inconsistent with the rest of the api.

daenney:
I can't find any precedent for how this would be inconsistent with the (which) rest of (what) API?

cyberious:
Also tests need to be updated for the new workflow as it is still mocking out `File.expects(:exists?)` https://github.com/puppetlabs/puppetlabs-stdlib/blob/master/spec/functions/loadyaml_spec.rb#L10 and should now instead be mocking out the Puppet::Util.absolute_path?

bmjen:
Thanks @amuraru for the contribution. Due to lack of activity, we are going to close this PR for now. If there is interest in adding this functionality, please fix failing tests and re-submit! Thanks!

PR #326, opened 2014-09-23
Added function for possibility of cutting variables

ruslanloman:


DavidS:
Hi,

given that this functionality is covered by puppet 4 and this PR has no specs, I'd rather not merge it.

PR #329, opened 2014-09-26
Make file_line accept white spaces in path

sulaweyo:
Took over the validation code from validate_absolute_path.rb. I did not use the function itself as it throws another type of error.


DavidS:
I've tested this agains the current head and cannot reproduce this behaviour. But I'll keep a spec in my WIP to avoid regressions.

PR #332, opened 2014-10-04
Adding a function random() to generate a random float. Updated README

WhatsARanjit:
Adding a function random() to generate a random float.


hunner:
This doesn't return floats or ranges; it just returns integers. Sorry.

PR #338, opened 2014-10-24
Clarifying behaviour of attributes and adding an extra example.

gjngeldenhuis:


tphoney:
Thanks for the PR, could you possibly squash these commits, and also fix the space changes that @igalic mentioned. 
Thanks again !

mhaskel:
Rebased/squashed in #430

PR #341, opened 2014-10-27
Add an append_file_content function

raphink:
Add content to a file resource.

This allows to generate concatenated files on the master side, without managing multiple file resources.

The first argument is the file path. You are responsible for creating the file resource yourself prior to calling the function. This way, you can finely control the file's parameters.

The second argument is the content to be added to the file.

The third (optional) argument is the order, specifying the position in which the content should be added. The default order is '10'. If you specified a content in the original file resource, this content is kept with order '00'

Examples:

    file { '/tmp/target':
      content => "Orig content\n",
    }

    append_file_content('/tmp/target', "hello, ")
    append_file_content('/tmp/target', "world\n")
    append_file_content('/tmp/target', "This is a multi
    line statement
    ", '01')
    append_file_content('/tmp/target', template('foobar/example.erb'), '05')

- *Type* : statement

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -80,6 +80,8 @@ If you want to use a standardized set of run stages for Puppet, `include stdlib`
 
 * `any2array`: This converts any object to an array containing that object. Empty argument lists are converted to an empty array. Arrays are left untouched. Hashes are converted to arrays of alternating keys and values. *Type*: rvalue
 
+* `append_file_content`: Add content to a file resource. This allows to generate concatenated files on the master side, without managing multiple file resources.
+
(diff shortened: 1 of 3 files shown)


nibalizer:
I think this is a bit too 'out there' for stdlib, can we submit this to puppet-community/extlib instead?

hunner:
Yeah, this is a bit too extravagant for stdlib. If a bunch of defined resource tries to use this and they are evaluated in random order, the fragments will be in random order.

We would prefer that you either move it to puppet-community/extlib as mentioned, or make it available in a standalone module. Thanks!

PR #342, opened 2014-10-27
Ability to pin beaker versions by ENV variable

cyberious:


mhaskel:
This is against a very old branch at this point, and I think this change should be in modulesync, closing.

PR #351, opened 2014-10-30
Added netmask2cidr function

hkumarmk:
This function help to convert network mask to cidr notation.

### DIFF
--- a/lib/puppet/parser/functions/netmask2cidr.rb
+++ b/lib/puppet/parser/functions/netmask2cidr.rb
@@ -1,0 +1,27 @@
+##
+## netmask2cidr
+##
+module Puppet::Parser::Functions
+  newfunction(
+    :netmask2cidr,
+    :type  => :rvalue,
+    :arity => 1,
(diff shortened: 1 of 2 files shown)


tphoney:
ping @hkumarmk Could you please rebase this. A few more tests would also be good.

elyscape:
It might also be a good idea to validate that it's a valid network mask, maybe by doing something like this:
```ruby
mask = IPAddr.new(args[0]).to_i.to_s(2)
is_valid = mask.match(/\A1*0*\z/)
raise Puppet::ParseError, "Invalid mask: #{args[0]}" unless is_valid
mask.count("1")
```

elyscape:
Also, `require` is not something that requires `end`.

mhaskel:
Hi @hkumarmk, thanks for the contribution. We haven't heard from you in a while, so I'm going to close this PR for now. If this is still something you're interested in having added to the module, please address the comments, above, and also update the README with documentation on the new function and resubmit. Thanks again!

PR #375, opened 2014-12-04
(MODULES-560) Add new functions validate_numeric() and validate_integer().

poikilotherm:
I saw MODULES-560 on Jira (https://tickets.puppetlabs.com/browse/MODULES-560) and had a need for it (check ports easily in a manifest), so I implemented it.

Added rspec unit tests, too.
Added necessary stuff to README.markdown, too.

Would be glad, if this gets merged... :) Thanks!

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -579,6 +579,76 @@ The first argument of this function should be the string to test, and the second
   
   *Type*: statement
 
+* `validate_integer`: Validate that the first argument is an integer (or an array of integers). Abort catalog compilation if any of the checks fail.
+    
+  The second argument is optional and passes a maximum. (All elements of) the first argument has to be less or equal to this max.
+
+  The third argument is optional and passes a minimum.  (All elements of) the first argument has to be greater or equal to this min.
+  If, and only if, a minimum is given, the second argument may be an empty string or undef, which will be handled to just check
+  if (all elements of) the first argument are greater or equal to the given minimum.
+
+  It will fail if the first argument is not an integer or array of integers, and if arg 2 and arg 3 are not convertable to an integer.
+
+  The following values will pass:
+
+  ```
+  validate_integer(1)
+  validate_integer(1, 2)
+  validate_integer(1, 1)
+  validate_integer(1, 2, 0)
+  validate_integer(2, 2, 2)
+  validate_integer(2, '', 0)
+  validate_integer(2, undef, 0)
+  $foo = undef
+  validate_integer(2, $foo, 0)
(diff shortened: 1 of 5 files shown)


cyberious:
With the future parse and types in the next version, what is this accomplishing that you can't with validate_re and a regex like '^\-?\d+\.?*\d*$'
Just trying to understand use case.

hlindberg:
With the new type system in Puppet 4.0 (and in future parser in 3.x) there is support for checking not only type but also the values of types. As an example, for integers (and floats) the type can be expressed as a range and the type can be combined with Optional (if an undef value is acceptable). Further you can use the scanf function to convert string representations to values (and handle failure).  Types can be combined to express things like - Array of integers in a range etc. Type checking is automatic if parameter are declared with type, and there is an assert_type function that can be used for manual type checking.

IMO we already have far too many "is_xxx" and "validate_xxx" functions in standard lib and we are about to deprecate and update these functions for puppet 4. Suggest sharing this validate_numeric function in a separate module for older puppet versions.

poikilotherm:
Right, didn't have the new Puppet 4 stuff on the radar, but wanted to be able to do things now, still using Puppet 3.x

Is there already such a module like you suggested? This seems to be a good plan...

hlindberg:
There is no such module (collection of 3x, new, but soon to be deprecated functions) - suggest you create your own.

README.markdown
```diff
@@ -579,6 +579,76 @@ The first argument of this function should be the string to test, and the second
...
+
+  The following values will pass:
+
+  ```
+  validate_integer(1)
+  validate_integer(1, 2)
+  validate_integer(1, 1)
+  validate_integer(1, 2, 0)
+  validate_integer(2, 2, 2)
+  validate_integer(2, '', 0)
```
daenney: How does `''` get converted to an integer?

README.markdown
```diff
@@ -579,6 +579,76 @@ The first argument of this function should be the string to test, and the second
...
+  validate_integer(2, undef, 0)
```
daenney: Same with `undef`?

hlindberg:
I really object to this having been added - a function called validate_integer() that accepts undef and empty string - really? And it accepts an Array. An array is not an integer.

@Hunner I strongly suggest reverting this. It will just trick people and make them write buggy code.

hlindberg:
I had a knee-jerk reaction and posted something that was not right. I deleted that comment. Sorry about the noise, should have read more carefully.

poikilotherm:
Ego te absolvum.

PR #382, opened 2014-12-16
(MODULES-1473) Deprecate type() function for new parser

hunner:
The `type()` function will cease to work on the new parser because 'type' is a reserved keyword. The `type3x()` function may be used to continue similar functionality, but will be deprecated in favor of the built-in typing system.

The `type_of()` function has been included to introspect types in the new parser.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -464,7 +464,12 @@ manifests as a valid password attribute. *Type*: rvalue
 * `to_bytes`: Converts the argument into bytes, for example 4 kB becomes 4096.
 Takes a single string value as an argument. *Type*: rvalue
 
-* `type`: Returns the type when passed a variable. Type can be a string, array, hash, float, integer, or boolean. *Type*: rvalue
+* `type3x`: Returns a string description of the type when passed a value. Type can be a string, array, hash, float, integer, or boolean. This function will be removed when puppet 3 support is dropped and the new type system may be used. *Type*: rvalue
+
+* `type_of`: Returns the literal type when passed a value. Requires the new
+  parser. Useful for comparison of types with `<=` such as in `if
+  type_of($some_value) <= Array[String] { ... }` (which is equivalent to `if
+  $some_value =~ Array[String] { ... }`) *Type*: rvalue
 
 * `union`: This function returns a union of two arrays. For example, `union(["a","b","c"],["b","c","d"])` returns ["a","b","c","d"].
 
--- a/lib/puppet/functions/type_of.rb
+++ b/lib/puppet/functions/type_of.rb
@@ -1,0 +1,17 @@
+# Returns the type when passed a value.
+#
+# @example how to compare values' types
+#   # compare the types of two values
+#   if type_of($first_value) != type_of($second_value) { fail("first_value and second_value are different types") }
+# @example how to compare against an abstract type
+#   unless type_of($first_value) <= Numeric { fail("first_value must be Numeric") }
+#   unless type_of{$first_value) <= Collection[1] { fail("first_value must be an Array or Hash, and contain at least one element") }
+#
+# See the documentation for "The Puppet Type System" for more information about types.
+# See the `assert_type()` function for flexible ways to assert the type of a value.
+#
+Puppet::Functions.create_function(:type_of) do
+  def type_of(value)
+    Puppet::Pops::Types::TypeCalculator.infer_set(value)
+  end
+end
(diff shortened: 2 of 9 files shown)


hunner:
I can't seem to get the unit test to work at all :(. It can't load the function, and digging through the puppet source code doesn't seem to highlight anything.

Help @hlindberg or @cyberious ?

hunner:
Tests pushed.

Someone labeled this "backwards incompatible" though it is backwards compat.

spec/unit/puppet/functions/type_of_spec.rb
```diff
@@ -0,0 +1,32 @@
+#! /usr/bin/env ruby -S rspec
+
+require 'spec_helper'
+require 'puppet/pops'
```
cyberious: needs to be wrapped within the ENV['FUTURE_PARSER']

lib/puppet/functions/type_of.rb
```diff
@@ -0,0 +1,13 @@
+# Returns the type when passed a value.
+#
+# @example how to compare values' types
+#   # compare the types of two values
+#   if type_of($first_value) != type_of($second_value) { fail("first_value and second_value are different types") }
+#
```
hlindberg: Can add this as well:

@example how to compare against a type
  unless type_of($first_value) <= Numeric { fail("first_value must be Numeric") } 
  unless type_of{$first_value) <= Collection[1] { fail("first_value must be an Array or Hash, and contain at least one element") }

lib/puppet/functions/type_of.rb
```diff
@@ -0,0 +1,13 @@
...
+# See the documentation for "The Puppet Type System" for more information about types.
+#
```
hlindberg: Can also add:

Also see the assert_type function for flexible ways to assert the type of a value.

lib/puppet/functions/type_of.rb
hunner: Sure.

README.markdown
```diff
@@ -464,7 +464,11 @@ manifests as a valid password attribute. *Type*: rvalue
 * `to_bytes`: Converts the argument into bytes, for example 4 kB becomes 4096.
 Takes a single string value as an argument. *Type*: rvalue
 
-* `type`: Returns the type when passed a variable. Type can be a string, array, hash, float, integer, or boolean. *Type*: rvalue
+* `type3x`: Returns a string description of the type when passed a variable. Type can be a string, array, hash, float, integer, or boolean. This function will be removed when puppet 3 support is dropped and the new type system may be used. *Type*: rvalue
```
hlindberg: Same change here as I commented on in an outdated diff, it is not just "a variable", it is any expression e.g. type3x(1+2)

README.markdown
```diff
@@ -464,7 +464,11 @@ manifests as a valid password attribute. *Type*: rvalue
...
+
+* `type_of`: Returns the literal type when passed a variable. Requires the new
```
hlindberg: Same here (not just a variable)

README.markdown
```diff
@@ -464,7 +464,11 @@ manifests as a valid password attribute. *Type*: rvalue
...
+  parser. The output is meant for human use; for programatic inspection of types
```
hlindberg: While it is true that interpolating the resulting type (or passing it to notice) will produce a string representation of the type, it is also very useful to get the type of something and using that result multiple times (as opposed to doing multiple =~ where the type has to be inferred each time.) Note that `$value =~ Numeric` is the same as doing `type_of($value) <= Numeric`.

hlindberg:
I think my last comment was made as you pushed new changes...

README.markdown
hunner: I think I like the look of `=~ Numeric` over `<= Numeric`. And yeah, I see your point, now. I wrote that before I knew what it was for.

README.markdown
```diff
@@ -464,7 +464,11 @@ manifests as a valid password attribute. *Type*: rvalue
 * `to_bytes`: Converts the argument into bytes, for example 4 kB becomes 4096.
 Takes a single string value as an argument. *Type*: rvalue
 
-* `type`: Returns the type when passed a variable. Type can be a string, array, hash, float, integer, or boolean. *Type*: rvalue
+* `type3x`: Returns a string description of the type when passed a value. Type can be a string, array, hash, float, integer, or boolean. This function will be removed when puppet 3 support is dropped and the new type system may be used. *Type*: rvalue
+
+* `type_of`: Returns the literal type when passed a value. Requires the new
+  parser. Useful for comparison of types with `=~` such as in `if $some_value
+  =~ Array[String] {` *Type*: rvalue
```
hlindberg: I think you meant this to say:

as in `if type_of($some_value) <= Array[String] { ... }`

hlindberg:
+1 on merging this

PR #390, opened 2014-12-19
Rename private() to assert_private()

fpletz:
As mentioned in #270, private is a reserved keyword in the future parser which is to be released with Puppet 4. As it stands, this function is not useable with the future parser so it needs to be renamed.

This is a breaking change.

This should've been fixed right after merging #270. But it wasn't.


cyberious:
@fpletz see [PR #382 |https://github.com/puppetlabs/puppetlabs-stdlib/pull/382] we should deprecate and forward on until we can do a 4.0 release.

cyberious:
Closing with merged code in #422

fpletz:
Thanks!

PR #404, opened 2015-01-15
(MODULES-1670) Do not match dotted-quad IP address as domain name

roderickm:
A valid host name can never have the dotted-decimal form #.#.#.#,
since at least the highest-level component label will be alphabetic.
See RFC 1123, Section 2.1
http://tools.ietf.org/html/rfc1123#section-2
Fixes PUP-3856

### DIFF
--- a/lib/puppet/parser/functions/is_domain_name.rb
+++ b/lib/puppet/parser/functions/is_domain_name.rb
@@ -33,6 +33,10 @@ Returns true if the string passed to this function is a syntactically correct do
     return false if domain.empty?
     return false if domain.length > domain_max_length
 
+    # The top level domain must be alphabetic if there are multiple labels.
+    # See rfc1123, 2.1
+    return false if domain.include? '.' and not /\.[A-Za-z]+$/.match(domain)
+
     # Check each label in the domain
     labels = domain.split('.')
     vlabels = labels.each do |label|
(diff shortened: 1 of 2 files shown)


hlindberg:
As I read the RFCs for domains, the highest order segment must be alphabetical which makes a domain such as 'a.b.c.42' not legal. 

Tests are needed.

lib/puppet/parser/functions/is_domain_name.rb
```diff
@@ -32,6 +32,7 @@ module Puppet::Parser::Functions
     # Check the whole domain
     return false if domain.empty?
     return false if domain.length > domain_max_length
+    return false if /^\d+\.\d+\.\d+\.\d+$/.match(domain)
```
hunner: Would it be appropriate to use `Resolv::IPv4::Regex` and `Resolv::IPv6::Regex` here to match both 4 and 6 ip addresses without having to write our own regexes?

Thanks.

roderickm: I agree that the match should only consider the highest-order label (aka top-level domain) because digits are not permitted there, which would also excluded dotted-quads. Using IPv4 or IPv6 tests would be too strict, I think, because there are dotted quads that would be invalid according to Resolv::IPv4::Regex (such as '192.168.1.500') but should still fail is_domain_name().

mhaskel:
@roderickm Thanks for the contribution! Could you rebase this and squash to a single commit?

roderickm:
@mhaskel Done!

mhaskel:
@roderickm Did you fetch current master and rebase against that? This is still showing as unmergeable. Thanks!

mhaskel:
Great, thanks @roderickm !

PR #405, opened 2015-01-27
(MODULES-1715) Add FQDN-based random string generator

elyscape:
As requested in [MODULES-1715](https://tickets.puppetlabs.com/browse/MODULES-1715), this PR provides `fqdn_rand_string()`, which generates random alphanumeric strings of the desired length.  It also provides `fqdn_rand_base64()`, which generates random base64 strings of the desired length.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -244,6 +244,23 @@ This function flattens any deeply nested arrays and returns a single flat array 
 Returns the largest integer less than or equal to the argument.
 Takes a single numeric value as an argument. *Type*: rvalue
 
+#### `fqdn_rand_string`
+
+Generates a random alphanumeric string using an optionally-specified character set (default is alphanumeric), combining the `$fqdn` fact and an optional seed for repeatable randomness.
+
+*Usage:*
+```
+fqdn_rand_string(LENGTH, [CHARSET], [SEED])
+```
+*Examples:*
+```
+fqdn_rand_string(10)
+fqdn_rand_string(10, 'ABCDEF!@#$%^')
+fqdn_rand_string(10, '', 'custom seed')
+```
+
+*Type*: rvalue
+
 #### `fqdn_rotate`
 
 Rotates an array a random number of times based on a node's fqdn. *Type*: rvalue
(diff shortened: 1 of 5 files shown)


elyscape:
As a side note, the performance of these could be improved by bringing the contents of `fqdn_rand()` and, more importantly, `Puppet::Util.deterministic_rand()` inline, as that would allow reuse of the same random generator for successive characters/integers, but that would decrease maintainability and I wasn't sure if the tradeoff was worth it.  If you think that it would be a worthwhile change, let me know and I'll make it happen.

elyscape:
I have added acceptance tests for the new functions.  Let me know if I should squash some/all of these commits.

elyscape:
It occurred to me that I didn't document the functions in the README, so I fixed that.  I squashed it into the first commit and squashed the two test commits together as well.

elyscape:
The test failures are due to the release of rspec 3.2.0, which breaks some things.

elyscape:

elyscape:
I went ahead and squashed this down into one commit.

hunner:
I can see random numbers and strings... but why base64?

elyscape:
Among other things, it can be used to generate password salts.  The POSIX spec says that the valid character set for password hash salts is a slightly-modified base64 (`+` replaced with `.`).

daenney:
But now you're asking us to include a base64 implementation which cannot be used for the salts according to the POSIX spec. You're countering your own point here.

hunner:
I'd actually rather have `fqdn_rand_string()` take an argument of the character set from which to generate a random string, then base64, or base64 with `.` instead of `+`, or random UTF-8 emoticons... etc. Something generic.

elyscape:
Good points.  I'll modify `fqdn_rand_string()` to accept a charset as a string.

elyscape:
Custom character sets added.  Pinging @daenney and @hunner for review.  If the review is positive, I'll squash it down.

elyscape:
Rebased onto master and squashed.  Pinging @daenney and @hunner for review.

PR #408, opened 2015-02-03
(MODULES-1737) Add pw_hash() function

elyscape:
As per [MODULES-1737](https://tickets.puppetlabs.com/browse/MODULES-1737), this PR provides a method of generating password hashes.  Tests coming.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -500,6 +500,24 @@ Calling the class or definition from outside the current module will fail. For e
 
   *Type*: statement
 
+#### `pw_hash`
+
+Hashes a password using the crypt function. Provides a hash usable on most POSIX systems.
+
+The first argument to this function is the password to hash. If it is undef or an empty string, this function returns undef.
+
+The second argument to this function is which type of hash to use. It will be converted into the appropriate crypt(3) hash specifier. Valid hash types are:
+
+|Hash type            |Specifier|
+|---------------------|---------|
+|MD5                  |1        |
+|SHA-256              |5        |
+|SHA-512 (recommended)|6        |
+
+The third argument to this function is the salt to use.
+
+Note: this uses the Puppet Master's implementation of crypt(3). If your environment contains several different operating systems, ensure that they are compatible before using this function.
+
 #### `range`
 
 When given range in the form of '(start, stop)', `range` extrapolates a range as an array. For example, `range("0", "9")` returns [0,1,2,3,4,5,6,7,8,9]. Zero-padded strings are converted to integers automatically, so `range("00", "09")` returns [0,1,2,3,4,5,6,7,8,9].
--- a/lib/puppet/parser/functions/pw_hash.rb
+++ b/lib/puppet/parser/functions/pw_hash.rb
@@ -1,0 +1,56 @@
+Puppet::Parser::Functions::newfunction(
+  :pw_hash,
+  :type => :rvalue,
+  :arity => 3,
+  :doc => "Hashes a password using the crypt function. Provides a hash
+  usable on most POSIX systems.
+
+  The first argument to this function is the password to hash. If it is
+  undef or an empty string, this function returns undef.
+
+  The second argument to this function is which type of hash to use. It
+  will be converted into the appropriate crypt(3) hash specifier. Valid
+  hash types are:
+
+  |Hash type            |Specifier|
+  |---------------------|---------|
+  |MD5                  |1        |
+  |SHA-256              |5        |
+  |SHA-512 (recommended)|6        |
+
+  The third argument to this function is the salt to use.
+
+  Note: this uses the Puppet Master's implementation of crypt(3). If your
+  environment contains several different operating systems, ensure that they
+  are compatible before using this function.") do |args|
+    raise ArgumentError, "pw_hash(): wrong number of arguments (#{args.size} for 3)" if args.size != 3
+    raise ArgumentError, "pw_hash(): first argument must be a string" unless args[0].is_a? String or args[0].nil?
+    raise ArgumentError, "pw_hash(): second argument must be a string" unless args[1].is_a? String
+    hashes = { 'md5'     => '1',
+               'sha-256' => '5',
+               'sha-512' => '6' }
+    hash_type = hashes[args[1].downcase]
+    raise ArgumentError, "pw_hash(): #{args[1]} is not a valid hash type" if hash_type.nil?
+    raise ArgumentError, "pw_hash(): third argument must be a string" unless args[2].is_a? String
+    raise ArgumentError, "pw_hash(): third argument must not be empty" if args[2].empty?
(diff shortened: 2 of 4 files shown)


elyscape:
Tests added.

elyscape:

cyberious:
Please update to have graceful failure on unsupported platforms for this function.   They way it is currently written if attempted on Windows it will fail hard.

lib/puppet/parser/functions/pw_hash.rb
```diff
@@ -0,0 +1,43 @@
...
+    raise ArgumentError, "pw_hash(): third argument must be a string" unless args[2].is_a? String
+    raise ArgumentError, "pw_hash(): third argument must not be empty" if args[2].empty?
+    raise ArgumentError, "pw_hash(): characters in salt must be in the set [a-zA-Z0-9./]" unless args[2].match(/\A[a-zA-Z0-9.\/]+\z/)
+
+    password = args[0].to_s
+    return nil if password.empty?
+
+    # work around JRuby bug in String::crypt for JRuby < 1.7.17
+    if RUBY_PLATFORM == 'java' and 'test'.crypt('$1$1') != '$1$1$Bp8CU9Oujr9SSEw53WV6G.'
+      def password.crypt(salt)
```
hunner: Why is this not just `module String def crypt ...` or something?

elyscape: I could change it to that.  It was based on some code I found elsewhere to work around this issue.

elyscape:
@cyberious I'm not sure what you mean. Ruby provides a basic implementation of `String#crypt()` on Windows, albeit one that only offers the default DES hashing scheme.

lib/puppet/parser/functions/pw_hash.rb
```diff
@@ -0,0 +1,43 @@
...
+    raise ArgumentError, "pw_hash(): #{args[1]} is not a valid hash type" unless ['1', '5', '6'].include? args[1].to_s
+    raise ArgumentError, "pw_hash(): third argument must be a string" unless args[2].is_a? String
+    raise ArgumentError, "pw_hash(): third argument must not be empty" if args[2].empty?
+    raise ArgumentError, "pw_hash(): characters in salt must be in the set [a-zA-Z0-9./]" unless args[2].match(/\A[a-zA-Z0-9.\/]+\z/)
+
+    password = args[0].to_s
+    return nil if password.empty?
+
+    # work around JRuby bug in String#crypt for JRuby < 1.7.17
+    if RUBY_PLATFORM == 'java' and 'test'.crypt('$1$1') != '$1$1$Bp8CU9Oujr9SSEw53WV6G.'
```
daenney: Why are we only salting on the JVM?

lib/puppet/parser/functions/pw_hash.rb
```diff
@@ -0,0 +1,43 @@
...
+      def password.crypt(salt)
```
daenney: This function is still masking the password variable.

lib/puppet/parser/functions/pw_hash.rb
elyscape: We salt everywhere.  See [line 42](https://github.com/puppetlabs/puppetlabs-stdlib/pull/408/files#diff-dfa59737972d9adbac64e202167f383eR42).  This just checks for a faulty `String#crypt` implementation that doesn't support extended hashes, which is known to exist in older versions of JRuby, and overrides it with a proper Java implementation of `crypt()`.  This weak implementation also exists on MRI on Windows, but we don't have a proper Java implementation under that scenario.  I could change this to fail if the implementation is weak and we're not on Java, though.  That seems like it would probably be a good idea.

lib/puppet/parser/functions/pw_hash.rb
elyscape: It actually overrides the `String#crypt` function on the `password` variable.  I'll add some documentation to clarify this.

elyscape:
Documentation and checking for unsupported platforms has been added.  Working on a spec test for unsupported platforms now.

elyscape:
Spec test is in.  Pinging @daenney, @cyberious, and @hunner for review.

README.markdown
```diff
@@ -377,6 +377,22 @@ Calling the class or definition from outside the current module will fail. For e
...
+
+  The first argument to this function is the password to hash. If it is undef or an empty string, this function returns undef.
+
+  The second argument to this function is a specifier of which hash type to use. Valid hash types are:
+
+  |Specifier|Hash type            |
+  |---------|---------------------|
+  |1        |MD5                  |
+  |5        |SHA-256              |
+  |6        |SHA-512 (recommended)|
```
igalic: why do we even implement hashes which we don't recommend?
why not make this the third, optional argument?

README.markdown
```diff
@@ -377,6 +377,22 @@ Calling the class or definition from outside the current module will fail. For e
...
+
+  The third argument to this function is the sale to use.
```
igalic: s/sale/salt/

README.markdown
daenney: There's a difference between "you shouldn't use this" and "we recommend you to use the strongest one".

README.markdown
elyscape: Good catch.  Fixed.

elyscape:
Rebased onto master and squashed.

lib/puppet/parser/functions/pw_hash.rb
```diff
@@ -0,0 +1,50 @@
...
+  |6        |SHA-512 (recommended)|
+
+  The third argument to this function is the salt to use.
+
+  Note: this uses the Puppet Master's implementation of crypt(3). If your
+  environment contains several different operating systems, ensure that they
+  are compatible before using this function.") do |args|
+    raise ArgumentError, "pw_hash(): wrong number of arguments (#{args.size} for 3)" if args.size != 3
+    raise ArgumentError, "pw_hash(): first argument must be a string" unless args[0].is_a? String
+    raise ArgumentError, "pw_hash(): #{args[1]} is not a valid hash type" unless ['1', '5', '6'].include? args[1].to_s
```
daenney: Can we do this the other way around?

User-understanding wise I much prefer: `pw_hash('lalalala', 'sha-256', 'oseatnoirsentaiore')` over `pw_hash('lalalala', '5', 'oseatnoirsentaiore')`.

The first case I can immediately see what's going on. The second case I need to look up some implementation detail around specifiers to understand what type of hash this function call will be computing.

elyscape: Sure.  Give me a few minutes to make the change.

elyscape:
@daenney How's it look now?

daenney:
Much better, but tests are red now.

elyscape:
@daenney I knew I forgot to do something!  Fixed.

daenney:
@hunner @cyberious I'm okay with this. Up to you now.

PR #409, opened 2015-02-06
Added dirs and files exist check functions

jordiclariana:
Add 4 functions:

- file_exists
- file_exists_array
- dir_exists
- dir_exists_array

They are documented in the README.markdown and I've also created their specs.


hunner:
Due to the fact that functions only run on the master, and even in puppet standalone functions run before application, this would be highly confusing to most of our userbase. Sorry.

PR #412, opened 2015-02-14
(MODULES-1771) Don't modify input to is_domain_name()

seanmil:
Fix is_domain_name() so it dup's its incoming argument
to avoid changing the original with a later chomp!


mhaskel:
Thanks @seanmil !!

PR #416, opened 2015-02-24
Add new function 'value_at' that returns value from hash, with optional default value

viranch:


mhaskel:
Hi @viranch, 

Thanks for the contribution! We're going to close this PR though, because there's an easy way to get this functionality with existing functions.

Here's how to do it:

```
$my_hash = { 'a' => 'b', 'b' => 'c', 'c' => 'd' }
if has_key($my_hash, 'z') {
  $_val = $my_hash['z']
} else {
  $_val = 'a'
}
```

Thanks again!

PR #418, opened 2015-02-26
Fix issue with 1.8.7 and upcase

cyberious:


lib/puppet/parser/functions/upcase.rb
```diff
@@ -31,8 +31,8 @@ module Puppet::Parser::Functions
       result = value.collect { |i| i.is_a?(String) ? i.upcase : i }
     elsif value.is_a?(Hash)
       result = {}
-      result << value.each_pair do |k, v|
-        return {k.upcase => v.collect! { |p| p.upcase }}
+      value.each_pair do |k, v|
+        result.merge!({k.upcase => v.collect! { |p| p.upcase }})
```
elyscape: Why not instead do
```ruby
result[k.upcase] = v.collect { |p| p.upcase}
```
This results in significantly object creations.

Additionally, this will fail if v is a nested array or a hash.

PR #419, opened 2015-02-28
Loosen the restrictions of upcase and allow for recursion of the objects...

cyberious:
... and only worry if the object responds to upcase

### DIFF
--- a/lib/puppet/parser/functions/upcase.rb
+++ b/lib/puppet/parser/functions/upcase.rb
@@ -17,22 +17,22 @@ Will return:
   ) do |arguments|
 
     raise(Puppet::ParseError, "upcase(): Wrong number of arguments " +
-                                "given (#{arguments.size} for 1)") if arguments.size < 1
+                                "given (#{arguments.size} for 1)") if arguments.size != 1
 
     value = arguments[0]
 
-    unless value.is_a?(Array) || value.is_a?(String) || value.is_a?(Hash)
+    unless value.is_a?(Array) || value.is_a?(Hash) || value.respond_to?(:upcase)
       raise(Puppet::ParseError, 'upcase(): Requires an ' +
-                                  'array, string or hash to work with')
+                                  'array, hash or object that responds to upcase in order to work')
     end
 
     if value.is_a?(Array)
       # Numbers in Puppet are often string-encoded which is troublesome ...
-      result = value.collect { |i| i.is_a?(String) ? i.upcase : i }
+      result = value.collect { |i| function_upcase([i]) }
     elsif value.is_a?(Hash)
       result = {}
       value.each_pair do |k, v|
-        result.merge!({k.upcase => v.collect! { |p| p.upcase }})
+        result[function_upcase([k])] = function_upcase([v])
       end
     else
       result = value.upcase
--- a/README.markdown
+++ b/README.markdown
@@ -475,7 +475,7 @@ Takes a single string value as an argument. *Type*: rvalue
 
 You can also use this with arrays. For example, `unique(["a","a","b","b","c","c"])` returns ["a","b","c"]. *Type*: rvalue
 
-* `upcase`: Converts a string or an array of strings to uppercase. For example, `upcase("abcd")` returns 'ABCD'. *Type*: rvalue
+* `upcase`: Converts an object, array or hash of objects that respond to upcase to uppercase. For example, `upcase("abcd")` returns 'ABCD'.  *Type*: rvalue
 
 * `uriescape`: Urlencodes a string or array of strings. Requires either a single string or an array as an input. *Type*: rvalue
 
--- a/spec/functions/upcase_spec.rb
+++ b/spec/functions/upcase_spec.rb
@@ -36,4 +36,23 @@ describe "the upcase function" do
         scope.function_upcase([{'test' => %w(this that and other thing)}])
     ).to eq({'TEST' => %w(THIS THAT AND OTHER THING)})
   end
+
+  if :test.respond_to?(:upcase)
+    it 'should accept hashes of symbols' do
+      expect(
+          scope.function_upcase([{:test => [:this, :that, :other]}])
+      ).to eq({:TEST => [:THIS, :THAT, :OTHER]})
+    end
+    it 'should return upcase symbol' do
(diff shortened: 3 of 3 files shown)


cyberious:
Hey @elyscape, thanks for your feedback on the latest upcase enhancement, I know there was talk about why we needed it.  The reason is so that we can get the keys however the user passes in the object whether it is 'grant', 'Grant' or 'GRANT' or any type of permutation.  Hope this captures some of your thoughts.  Also I had some of my own, like do we care what kind of object it is if it responds to upcase, symbols don't in ruby 1.8.7 but they do in 1.9.3 and 2.x

elyscape:
This will fail if you pass in a hash that contains a numeric type anywhere inside it.  I think the best thing to do would be to validate that the initial object passed in as either an Array, a Hash, or something that responds to `upcase` and then hand off the recursion to a helper function that does the actual work.

lib/puppet/parser/functions/upcase.rb
```diff
@@ -21,18 +21,18 @@ module Puppet::Parser::Functions
 
     value = arguments[0]
 
-    unless value.is_a?(Array) || value.is_a?(String) || value.is_a?(Hash)
+    unless value.is_a?(Array)  || value.is_a?(Hash) ||  value.respond_to?(:upcase)
```
elyscape: Why the spacing changes here?

lib/puppet/parser/functions/upcase.rb
```diff
@@ -21,18 +21,18 @@ module Puppet::Parser::Functions
...
 
     if value.is_a?(Array)
       # Numbers in Puppet are often string-encoded which is troublesome ...
-      result = value.collect { |i| i.is_a?(String) ? i.upcase : i }
+      result = value.collect { |i| function_upcase([i]) }
     elsif value.is_a?(Hash)
       result = {}
       value.each_pair do |k, v|
-        result.merge!({k.upcase => v.collect! { |p| p.upcase }})
+        result.merge!({function_upcase([k]) => function_upcase([v])})
```
elyscape: This results in way more object creations than `result[function_upcase([k])] = function_upcase([v])` would and, as a result, will require more memory and be less performant.  More importantly, it's also less clear, IMO.

lib/puppet/parser/functions/upcase.rb
cyberious: IDE, generated white spaces

elyscape:
As a side note, why are we checking that the number of arguments passed in is at least 1?  Shouldn't we be checking that it's exactly 1, considering we only use the first argument?

lib/puppet/parser/functions/upcase.rb
cyberious: except in the case where a user passes 'grant' => {'select'}, 'Grant' => {'Update'} this would merge the two whereas the proposed change would only take the last occurrence.

cyberious:
I agree we should check to see if we are being passed at least one argument, didn't notice it was missing.

elyscape:
No, we are checking to see if we're being passed at least one argument, but I'm asking why we're not instead checking to see if we're being passed exactly one argument.  That is, [where we currently have](https://github.com/cyberious/puppetlabs-stdlib/blob/master/lib/puppet/parser/functions/upcase.rb#L19-L20):
```ruby
    raise(Puppet::ParseError, "upcase(): Wrong number of arguments " +
                                "given (#{arguments.size} for 1)") if arguments.size < 1
```
it would make more sense if we had:
```ruby
    raise(Puppet::ParseError, "upcase(): Wrong number of arguments " +
                                "given (#{arguments.size} for 1)") if arguments.size != 1
```

lib/puppet/parser/functions/upcase.rb
elyscape: Do you mean `{'grant' => ['select'], 'Grant' => ['Update']}`?  Because so far as I can tell `{'grant' => {'select'}}` is valid in neither Ruby nor Puppet.

elyscape: If you do mean `{'grant' => ['select'], 'Grant' => ['Update']}`, I have some bad news:
```
irb(main):009:0> {'grant'=>'select'}.merge({'grant'=> 'update'})
=> {"grant"=>"update"}
irb(main):010:0> {'grant'=>['select']}.merge({'grant'=> ['update']})
=> {"grant"=>["update"]}
irb(main):011:0> {'grant'=>{1=>'b'}}.merge({'grant'=> {2=>'c'}})
=> {"grant"=>{2=>"c"}}
```
It looks like `Hash#merge` overwrites old values with new ones too.

cyberious:
Instead of this check we use arity and ensure that we are only ever passed one parameter.  This PR actual does this for all of the functions but waiting for Puppet 4 to merge and slate for Stdlib 5

lib/puppet/parser/functions/upcase.rb
cyberious: Fair point on the deep merge. And I did actually mean an array as in `{'grant' => ['select'], 'Grant' => ['Update']}` if that is the case then we will need a param or new function for that.

lib/puppet/parser/functions/upcase.rb
```diff
@@ -17,22 +17,22 @@ module Puppet::Parser::Functions
...
 
-    unless value.is_a?(Array) || value.is_a?(String) || value.is_a?(Hash)
+    unless value.is_a?(Array) || value.is_a?(Hash) || value.respond_to?(:upcase)
       raise(Puppet::ParseError, 'upcase(): Requires an ' +
-                                  'array, string or hash to work with')
+                                  'array, hash or object that responds to upcase in order to work')
     end
 
     if value.is_a?(Array)
       # Numbers in Puppet are often string-encoded which is troublesome ...
```
mhaskel: The code to deal with the fact that numbers are often string-encoded seems to have been removed? Will this work with numbers?

cyberious: As long as it is passed as a String type it will work as it always has, that being said, with future parser it will fail if it is passed a Numeric type, but this was always going to be the case.

cyberious:
@mhaskel docs added.

PR #421, opened 2015-03-04
Add hash2json function

cleung2010:
This function helps convert a hash into a JSON string

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -286,6 +286,24 @@ returns the value of the resource's parameter. For example, the following code r
 
 * `hash`: This function converts an array into a hash. For example, `hash(['a',1,'b',2,'c',3])` returns {'a'=>1,'b'=>2,'c'=>3}. *Type*: rvalue
 
+* `hash2json`: Converts a hash to a sorted json string. This function supports nested hashes.  *Type*: rvalue
+
+  *Example*:
+
+  ```
+  $myhash = {
+    b_key => 'val1',
+    a_key => 'val2'
+  }
+
+  notice( hash2json($myhash) )
+  # The resulting json output will be
+  # {
+  #   "a_key": "val2",
+  #   "b_key": "val1"
+  # }
+  ```
+
 * `intersection`: This function returns an array an intersection of two. For example, `intersection(["a","b","c"],["b","c","d"])` returns ["b","c"].
 
 * `is_array`: Returns 'true' if the variable passed to this function is an array. *Type*: rvalue
@@ -736,7 +754,3 @@ To report or research a bug with any part of this module, please go to
 ##Contributors
 
 The list of contributors can be found at: https://github.com/puppetlabs/puppetlabs-stdlib/graphs/contributors
-
-
-
-
--- a/lib/puppet/parser/functions/hash2json.rb
+++ b/lib/puppet/parser/functions/hash2json.rb
@@ -1,0 +1,60 @@
+# Recursive function that builds json from a hash
+def hash_to_json (hash, json, depth)
+  # Base case: If the hash has been completely parsed, return the json string
+  return json if !hash.is_a?(Hash)
+
+  # Since puppet generates the hash with random KV pairs ordering every time,
+  # we want to maintain the same json output to avoid triggering refresh events
+  # so we have to sort the hash in order
+  hash = hash.sort
+  json = "{\n"
+  hash.each_with_index do |(key, value), index|
+    # Convert the key to a quoted string
+    key = key.inspect
+    # Check for value type and convert to quoted string if it is a string
+    value = value.inspect if value.is_a?(String)
+    # If the value pair is an array build a new json string recursively
+    # and concatenate it with brackets as a string
+    if value.is_a?(Array)
+      json2 = ""
+      value.each_with_index do |val, i|
+        json2 += val.inspect if val.is_a?(String)
+        json2 += hash_to_json(val, "", depth + 1) if val.is_a?(Hash)
+        json2 += "," unless i == value.length - 1
+      end
+      value = "[" + json2 + "]"
+    end
+    # Recurse hash_to_json with one more depth level to build any hashes
+    # that exists as a value on the KV pair
+    value = hash_to_json(value, json, depth + 1) if value.is_a?(Hash)
+    # Depth is used here to correctly indent the KV output line
+    depth.times do
+      json += "\t"
+    end
+    # Concatenate the key and value pair in the correct json format
+    json += "#{key}: #{value}"
+    # Add trailing comma unless the it is the last KV pair
+    json += "," unless (index + 1) == hash.length
+    json += "\n"
+  end
+  # Depth is used here to correctly indent the closing bracket
+  (depth-1).times do
+    json += "\t"
+  end
+  json += "}"
+end
+
+module Puppet::Parser::Functions
+  newfunction(:hash2json, :type => :rvalue, :doc => <<-EOS
+Converts a hash to a sorted json string. This function supports nested hashes.
+    EOS
+  ) do |args|
+
+    raise(Puppet::ParseError, "hash2json(): Wrong number of arguments given (#{args.size} for 1)") unless args.size == 1
+    raise(Puppet::ParseError, "hash2json(): First argument must be a Hash") unless args[0].is_a?(Hash)
+
+    hash = args[0]
+    json = hash_to_json(hash, "", 1)
+    return json
+  end
+end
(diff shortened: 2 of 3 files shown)


elyscape:
Instead of writing your own `hash_to_json()` function, why not use the Ruby-provided solution?
```ruby
module Puppet::Parser::Functions
  newfunction(:hash2json, :type => :rvalue, :doc => <<-EOS
Converts a hash to a sorted json string. This function supports nested hashes.
    EOS
  ) do |args|

    raise(Puppet::ParseError, "hash2json(): Wrong number of arguments given (#{args.size} for 1)") unless args.size == 1
    raise(Puppet::ParseError, "hash2json(): First argument must be a Hash") unless args[0].is_a?(Hash)

    require 'json'
    args[0].to_json
    # If you want it to be sorted, replace that line with:
    # args[0].sort.to_h.to_json
  end
end
```

cleung2010:
I have actually tried using that beforehand, but the json gem would not work on Windows. It would give the following error:
`ERROR: Failed to build gem native extension.`
Puppet on Windows uses its own version of Ruby (I believe it's 1.8.7). You could install Ruby devkit inside in Puppet's Ruby but that's not always possible, especially when you don't have permissions to the machine in a production environment.

hunner:
If you need json but don't have ruby dev available for native gems, use `json-pure`

cleung2010:
I remember switching to using `json-pure`, but don't recall what was the reason I didn't choose to use it. Let me do a quick run on on Windows to verify that there is no issue using that.

elyscape:
Puppet on Windows currently ships with Ruby 2.0 for x64 and I believe 1.9.3 for x32.  Regardless, from [the Puppet system requirements](https://docs.puppetlabs.com/puppet/latest/reference/system_requirements.html#mandatory-libraries):
> Mandatory Libraries
> * The `json` gem (any modern version)

hunner:
Also, puppet already has json listed as a mandatory library, so I wouldn't feel bad about merging a function that uses ruby's json: https://docs.puppetlabs.com/puppet/latest/reference/system_requirements.html#mandatory-libraries

On the other hand, what is the use case for this PR?

cleung2010:
This wouldn't apply for machines with Puppet version < 3.5 though. Puppet 3.4.x is relatively new and there is a possibility that machines (especially production ones) are still using that version of Puppet.

elyscape:
Hmm, good point.  Looks like the requirement of having the `json` or `json_pure` gem was added with 3.5.0.  Though the gem only needs to be present on the master.  I'll need to think about that.

In the meantime, what's the use case for this PR?

cleung2010:
The error I am getting is:
`Error: Could not run: no such file to load -- json_pure`

I did a gem install in to the system-wide Ruby as well as Puppet's Ruby 
`C:\Puppet\sys\ruby\bin\gem.bat install json_pure`
Verified the gem is in the path `C:\Puppet\sys\ruby\lib\ruby\gems\1.8\gems`

```
    Directory: C:\Puppet\sys\ruby\lib\ruby\gems\1.8\gems


Mode                LastWriteTime     Length Name
----                -------------     ------ ----
d----         10/8/2014  10:50 AM            bundler-1.7.3
d----         10/8/2014  10:35 AM            CFPropertyList-2.2.8
d----         10/9/2014   1:12 PM            ffi-1.9.3-x86-mingw32
d----         10/8/2014  10:35 AM            i18n-0.6.11
d----          3/4/2015   4:37 PM            json_pure-1.8.2
d----         10/8/2014  10:35 AM            rake-10.3.2
d----         10/7/2014   4:25 PM            sys-admin-1.5.6-x86-mingw32
d----        10/27/2014   3:38 PM            tilt-1.4.1
d----         10/7/2014   4:25 PM            win32-api-1.4.8-x86-mingw32
d----         10/7/2014   4:25 PM            win32-dir-0.3.7
d----         10/7/2014   4:25 PM            win32-eventlog-0.5.3
d----         10/7/2014   4:25 PM            win32-process-0.6.5
d----         10/7/2014   4:25 PM            win32-security-0.1.4
d----         10/7/2014   4:25 PM            win32-service-0.7.2-x86-mingw32
d----         10/7/2014   4:25 PM            win32-taskscheduler-0.2.2
d----         10/7/2014   4:25 PM            win32console-1.3.2-x86-mingw32
d----         10/7/2014   4:25 PM            windows-api-0.4.0
d----         10/7/2014   4:25 PM            windows-pr-1.2.2
```

elyscape:
Even when using the `json_pure` gem, you still `require 'json'`.

cleung2010:
The use case is for modules like puppet-consul and puppet-consul_replicate which would need to generate a JSON-formatted config file from a hash class parameter that the user can pass in. 

Basically it's handy when a JSON config file is needed and Puppet can be used to generate it on the fly via a template and/or hiera. I had this code in several other private repos, but it was getting repetitive and though I should consolidate in to a single module so my first thought was to add it to stdlib.

elyscape:
Instead of having this as a function in stdlib and passing the results to a file resource's content parameter, wouldn't it make more sense to put the conversion into a template and use that?

cleung2010:
I am dubious on having to require `json` when there is `json_pure` (I thought the whole point was to have a variant that would not need `ruby-dev`)

In any case I would get this error if I require `json`:
`ERROR: Failed to build gem native extension.`

elyscape:
To clarify, `json_pure` is a gem provided by the same people behind the `json` gem, and you use it exactly like you would the `json` gem.  This means that you require it with `require 'json'`, even though the installed gem is called `json_pure`.

cleung2010:
Generating JSON in the template would not be as clean, specially when you don't know the structure of the JSON. Sure, it can be built within the template if it's something simple, but dynamically generating JSON from a hash would be more effective in a function (unless you wrap almost everything in `<%= %>`).

elyscape:
I meant more having a template that's something like:
```erb
<% require 'json' %>
<%= @somevar.to_json %>
```

hunner:
Templates only return strings, whereas custom functions can return arrays. So `inline_template` probably wouldn't work.

elyscape:
But the whole point is to get JSON string output, right?

cleung2010:
@elyscape Thanks for clarifying, I did not know that. :) 
However, I am still getting `Error: Could not run: no such file to load -- json_pure`

elyscape:
@cleung2010 What is the code you are using?

cleung2010:
I'm using it in a private module, but essentially this is what I'm trying to do in `config.json.erb` under your suggestion:

```erb
<% require 'json' %>
<%= scope.lookupvar('consul::populate_hash::merged_config_hash').to_json %>
```

elyscape:
Okay.  You'll need to install the `json_pure` gem on your Puppet master (or agent, if you're running standalone or `puppet apply`).

cleung2010:
Yep I got json_pure installed in both Puppet's Ruby and system-wide Ruby to ensure that it was there. https://github.com/puppetlabs/puppetlabs-stdlib/pull/421#issuecomment-77255126

elyscape:
So you're doing this via `puppet apply`?

cleung2010:
That's correct.

daenney:
So, manually making JSON is not a good idea.

The only thing you need to do is check if the hash responds to `.to_json` and if not fail.

cleung2010:
I don't think it makes sense to say that Puppet <= 3.5 is ancient when the documentation on https://github.com/puppetlabs/puppetlabs-stdlib#version-compatibility says that it will support Puppet 3.x all the way until stdlib 5.x. Also, failing when .to_json call fails means that Windows systems will not work if they are running Puppet < 3.4.x, which again is likely to be the case in production servers.

daenney:
In that case it's highly unlikely we'll merge this. We've had some discussion about this with @mhaskel @hunner @cyberious and @cmurphy and we're all uncomfortable with a solution that manually tries to build up some JSON.

cleung2010:
@elyscape So I ran the `puppet apply` again on a fresh machine and it worked. However, I did have to `gem install json_pure` in Puppet's Rubygem. Thanks for pointing out `require 'json'`!

cleung2010:
No problem, perfectly understandable. I was using this function on a few
private modules so I thought of contributing back and consolidating it to
one place. It's good to know that json_pure is working, so I'll generate it
directly in the template and find a way of installing the gem on the
machines.

cleung2010:
After some more thought, I don't think installing the gem on machines that I don't own is possible, especially since I am just the module provider and do not manage their deployment nor am I in the position to enforce installing it separately. Nevertheless, it's good to know that a solution exists, just not applicable in our use case. You can go ahead and close this PR :)

mhaskel:
Ok, great, thanks @cleung2010 . I'll close this.

PR #424, opened 2015-03-08
(#MODULES-865) validate_re cannot match numeric regex in all cases

ripzura:
The validate_re function now tries to convert it's first argument to
a string. This will ensure that numeric, boolean, symbols, arrays and hashes can be matched
by a regex since regex matching can only happen on strings.

validate_re(1, '^\d+$') would previously fail because the first argument is a Fixnum.
This patch ensures the first argument is cast to a string so this would match as one would expect.

### DIFF
--- a/lib/puppet/parser/functions/validate_re.rb
+++ b/lib/puppet/parser/functions/validate_re.rb
@@ -28,12 +28,18 @@ module Puppet::Parser::Functions
       raise Puppet::ParseError, ("validate_re(): wrong number of arguments (#{args.length}; must be 2 or 3)")
     end
 
-    msg = args[2] || "validate_re(): #{args[0].inspect} does not match #{args[1].inspect}"
+    input = args[0]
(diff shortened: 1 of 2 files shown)


ripzura:
Test seems to fail for arrays and hashes on puppet 2.7.x. So my fix does not work for Arrays and Hashes on 2.7.x (function reverts to it's old behavior) It does work for symbols, integers and floats however.

Since I cannot test myslf on 2.7.x I would like some comments on how to proceed. Is this critical?
I thought the current stdlib wasn't supposed to be compatible with 2.7.x anyway so any pointers on how to get this patch merged are welcome.

DavidS:
I'm currently working on a major cleanup for the next major stdlib release. I'll see to it that this will be fixed there.

DavidS:
Actually looking through the code, this change would break backwards compatibility in exactly the case where it's needed the most: when the input is unexpected. I've replaced this with #526, clarifying the docs and the error message.

PR #425, opened 2015-03-10
Fix off-by-one error in validate_augeas_spec.rb that was causing rspec failure

jeffcoat:
This fixes the test "Nicer Error Messages" in puppetlabs-stdlib/spec/functions/validate_augeas_spec.rb. Without it, 'rake spec' will always fail with RegexpError: empty char-class: /[]/ as Ruby tries and fails to build a pattern off the wrong entry in the inputs array.

I do not understand why I'm the only one to have run into this. But I believe this trivial fix is what the original author intended.

### DIFF
--- a/spec/functions/validate_augeas_spec.rb
+++ b/spec/functions/validate_augeas_spec.rb
@@ -60,7 +60,7 @@ describe Puppet::Parser::Functions.function(:validate_augeas), :if => Puppet.fea
     end
 
     describe "Nicer Error Messages" do
-      # The intent here is to make sure the function returns the 3rd argument
+      # The intent here is to make sure the function returns the 4th argument
       # in the exception thrown
       inputs = [
         [ "root:x:0:0:root\n", 'Passwd.lns', [], 'Failed to validate passwd content' ],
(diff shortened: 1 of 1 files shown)


jeffcoat:
I see now why this hasn't been noticed before: validate_augeas_spec.rb runs :if => Puppet.features.augeas?. I was running the tests out of my Puppet Enterprise install, so I'm checking 890 examples, while the  travis-ci build is only checking 892.

hunner:
Thanks! Could you add the augeas gem to the Gemfile so that this test can run in Travis CI?

DavidS:
I've just opened a pull request on the augeas gem to make it compile on more recent installations.

DavidS:
Having a working augeas gem, I got the tests running on a Ubuntu 14.04 machine. 10 out of 16 added tests failed with this:

    expected Exception with message matching /testing path/, got #<NoMethodError: private method `open' called for Augeas:Class> with backtrace:
      # ./lib/puppet/parser/functions/validate_augeas.rb:44:in `block in <module:Functions>'

That does not look healthy to me.

DavidS:
augh, the gem we actually want is 'ruby-augeas ~> 0.5.0'. With that the tests run and this PR in fact fixes two failing test cases. It contains a native extension that won't compile everywhere though :-(

PR #429, opened 2015-04-08
Update Travis CI job from current modulesync_configs

DavidS:


.travis.yml
```diff
@@ -6,13 +6,19 @@ script: "bundle exec rake validate && bundle exec rake lint && bundle exec rake
...
+  - rvm: 1.9.3
+    env: PUPPET_GEM_VERSION="~> 3.4.0"
   - rvm: 1.8.7
-    env: PUPPET_GEM_VERSION="~> 2.7.0" FACTER_GEM_VERSION="~> 1.6.0"
-  - rvm: 1.8.7
-    env: PUPPET_GEM_VERSION="~> 2.7.0" FACTER_GEM_VERSION="~> 1.7.0"
+    env: PUPPET_GEM_VERSION="~> 3.0"
   - rvm: 1.9.3
     env: PUPPET_GEM_VERSION="~> 3.0"
-  - rvm: 2.0.0
```
cyberious: Why are we removing 2.0 we still use this explicitly with PE 3.7 on Windows. There is enough difference with 2.1.5 and 2.0 that I would rather see both

DavidS:
That change was done here: https://github.com/puppetlabs/modulesync_configs/commit/f3e1c6d78a4fbf3d97124849d9c20da6cee13cc9

.travis.yml
DavidS: That change was done here: https://github.com/puppetlabs/modulesync_configs/commit/f3e1c6d78a4fbf3d97124849d9c20da6cee13cc9

PR #435, opened 2015-04-10
Added an second argument to getvar for a optional default

solarkennedy:
In a lot of our puppet code we do stuff like this:

```puppet
$something = getvar('maybe_this_exists')
if $something {
  $something_real = $something
} else {
  $something_real = 'some_default'
}
```

This PR makes it so we can say this instead:
```puppet
$something = getvar('maybe_this_exists', 'some_default')
```

This mimics the python behavior (which I like) 
https://docs.python.org/2/library/stdtypes.html#dict.get

I'm pretty sure I got the tests right?

cc @hashbrowncipher @bobtfish


hashbrowncipher:
Yep, nice feature idea @solarkennedy.

igalic:
i've been told it's problematic to make a getter function which also influences the returning value…

so, for a second i thought we could just use a selector here with a `default` label, but then, we'd have to call getvar() multiple times
```puppet
$something = getvar('maybe_this_exists') ? {
  default => oh wow this is annoying…
```

i don't think we're caching (pure?) function calls currently, nor do we have a Perl style `$_` variable, which would be holding such a value…

solarkennedy:
thanks @bobtfish for the failing test. I'll incorporate it and try to get it to pass.

solarkennedy:
Ok. How about now?

cyberious:
@solarkennedy Any reason why we couldn't use `pick(getvar('maybe_this_exists'),'default value')` instead? I would like to know the use case that would not allow use to use the aforementioned?

solarkennedy:
@cyberious yes, I supposed that achieves the same effect!

PR #436, opened 2015-04-14
Can use a captured string in match parameter

melkor:
Hi, In my case I need to add strings in middle of an existing line. 
For exemple, in this existing file applied on all hosts managed by my puppet master, I want add another subnet for specific rules:
- from "sshd: 10.10.10.0/255.255.255.0"
-     to "sshd: 10.10.10.0/255.255.255.0 92.156.27.32"

I think File_line is the good resource for doing this job.
     file_line { 'hosts.allow':
          ensure    => present,
          path      => '/etc/hosts.allow',
          line      => "sshd: \\1 92.126.27.32",
          match     => '^sshd: (.*)^',
        }


igalic:
the example you provide us super confusing (because of the order of line and match ;) and a bit wrong (because you use ^ as terminator)

melkor:
I'm sorry, I wrote this example too quickly  :)
This is another exemple. 
I used it in my office to add a custom cmdline parameter in /etc/default/grub:
        file_line { "add ${cmdline} in /etc/default/grub":
            match     => '^GRUB_CMDLINE_LINUX_DEFAULT="(.*)"',
            line      => "GRUB_CMDLINE_LINUX_DEFAULT=\"\\1 ${cmdline}\"",
            no_append => true,
            path      => '/etc/default/grub',
        }

It produces : 
GRUB_CMDLINE_LINUX_DEFAULT="quiet eth_order=08:00:27:55:a8:04"

bmjen:
Thanks for the contribution @melkor! However, we don't think this capability is within the scope of the file_line function.

PR #442, opened 2015-04-20
Adding a respond_to function

danzilio:
This PR adds a `respond_to` function to check if a resource type has a particular parameter, similar (in concept) to the `respond_to?` method in Ruby. This could be useful in a number of circumstances, not sure if I'm the only one that finds this function interesting.

### DIFF
--- a/lib/puppet/parser/functions/respond_to.rb
+++ b/lib/puppet/parser/functions/respond_to.rb
@@ -17,15 +17,15 @@ EOS
   if args[0] =~ /^(?i:class)$/
     type, title, params = args
     raise ArgumentError, 'Must specify a class name' unless title
-    raise ArgumentError, 'Must specify a parameter' unless params
     resource = find_hostclass(title)
   else
     title, params = args
     type = 'resource'
-    raise ArgumentError, 'Must specify a parameter' unless params
+    raise ArgumentError, 'Must specify a resource type' unless title
     resource = find_resource_type(title) || find_definition(title)
   end
 
+  raise ArgumentError, 'Must specify parameter(s)' unless params
   raise ArgumentError, "The #{title} #{type} could not be found" if resource.nil?
 
   params = [params] unless params.is_a?(Array)
--- a/lib/puppet/parser/functions/respond_to.rb
+++ b/lib/puppet/parser/functions/respond_to.rb
@@ -8,23 +8,34 @@ returns true if that resource responds to the given parameter.
     respond_to('class', 'apt', 'always_update_apt')
     respond_to('file', 'path')
     respond_to('apt::key', 'key')
+    respond_to('archive', ['checksum_type', 'checksum_file'])
 EOS
 ) do |args|
   args = [args] unless args.is_a?(Array)
+  valid_params = []
 
   if args[0] =~ /^(?i:class)$/
-    type, title, param = args
+    type, title, params = args
     raise ArgumentError, 'Must specify a class name' unless title
-    raise ArgumentError, 'Must specify a parameter' unless param
+    raise ArgumentError, 'Must specify a parameter' unless params
     resource = find_hostclass(title)
   else
-    title, param = args
+    title, params = args
     type = 'resource'
-    raise ArgumentError, 'Must specify a parameter' unless param
+    raise ArgumentError, 'Must specify a parameter' unless params
     resource = find_resource_type(title) || find_definition(title)
   end
 
   raise ArgumentError, "The #{title} #{type} could not be found" if resource.nil?
 
-  resource.valid_parameter?(param)
+  params = [params] unless params.is_a?(Array)
+  params.each do |param|
+    valid_params << resource.valid_parameter?(param)
+  end
+
+  if valid_params.include?(false)
+    return false
+  else
+    return true
+  end
 end
--- a/lib/puppet/parser/functions/respond_to.rb
+++ b/lib/puppet/parser/functions/respond_to.rb
@@ -1,0 +1,30 @@
+# Test whether a given class or resource type responds to the given parameter.
+Puppet::Parser::Functions.newfunction(:respond_to,
+                                      :type => :rvalue,
+                                      :doc => <<-EOS
+Takes a resource type, title (only if type == class), and parameter name and
+returns true if that resource responds to the given parameter.
+
+    respond_to('class', 'apt', 'always_update_apt')
+    respond_to('file', 'path')
+    respond_to('apt::key', 'key')
+EOS
+) do |args|
+  args = [args] unless args.is_a?(Array)
+
+  if args[0] =~ /^(?i:class)$/
+    type, title, param = args
+    raise ArgumentError, 'Must specify a class name' unless title
+    raise ArgumentError, 'Must specify a parameter' unless param
+    resource = find_hostclass(title)
+  else
+    title, param = args
+    type = 'resource'
+    raise ArgumentError, 'Must specify a parameter' unless param
+    resource = find_resource_type(title) || find_definition(title)
+  end
+
+  raise ArgumentError, "The #{title} #{type} could not be found" if resource.nil?
+
+  resource.valid_parameter?(param)
+end
--- a/spec/functions/respond_to_spec.rb
+++ b/spec/functions/respond_to_spec.rb
@@ -1,0 +1,45 @@
+#! /usr/bin/env ruby -S rspec
+
+require 'spec_helper'
+
+describe 'respond_to' do
+  describe 'with no arguments' do
+    it { is_expected.to run.with_params().and_raise_error(ArgumentError, /Must specify a resource type/) }
+  end
+
+  describe 'when looking up a class' do
+    let(:pre_condition) do
+      'class mock ($mocked_var1, $mocked_var2) { }'
+    end
+
+    it { is_expected.to run.with_params('class', 'mock', 'mocked_var1').and_return(true) }
+    it { is_expected.to run.with_params('class', 'mock', ['mocked_var1']).and_return(true) }
+    it { is_expected.to run.with_params('class', 'mock', 'nonexistent').and_return(false) }
+    it { is_expected.to run.with_params('class', 'mock', ['mocked_var1', 'mocked_var2']).and_return(true) }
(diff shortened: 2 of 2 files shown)


elyscape:
> This could be useful in a number of circumstances

Could you provide some examples?  I'm not really sure I see the use case at present.

danzilio:
Sure!! The most useful (not contrived) example I can think of is supporting different module interfaces. For example, the `camptocamp/archive`  and `nanliu/archive` modules all do basically the same thing, but have different interfaces. If I want to be able to support both modules, I need to adjust the parameters I'm calling based on the interface available to me.

To pull a file with `nanliu/archive` I would use:

```
archive { 'apache-tomcat-6.0.26':
  ensure        => present,
  source        => 'http://archive.apache.org/dist/tomcat/tomcat-6/v6.0.26/bin/apache-tomcat-6.0.26.tar.gz',
  extract_path  => '/opt',
}
```

But with `camptocamp/archive` I would use:

```
archive { 'apache-tomcat-6.0.26':
  ensure => present,
  url    => 'http://archive.apache.org/dist/tomcat/tomcat-6/v6.0.26/bin/apache-tomcat-6.0.26.tar.gz',
  target => '/opt',
}
```

To abstract this away and allow my module users to pick whichever `archive` module they wish, I need to check the 'signature' of the `archive` resource. Here's how I would implement that using this `respond_to` function:

```
class fetch (
  $filename = 'apache-tomcat-6.0.26',
  $url = 'http://archive.apache.org/dist/tomcat/tomcat-6/v6.0.26/bin/apache-tomcat-6.0.26.tar.gz',
  $target = '/opt',
) {

  if respond_to('archive', 'url') and respond_to('archive', 'target') {
    archive { $filename:
      ensure => present,
      url    => $url,
      target => $target,
    }
  } 
  elsif respond_to('archive', 'source') and respond_to('archive', 'extract_path') {
    archive { $filename:
      ensure        => present,
      source        => $url,
      extract_path  => $target,
    }
  }
  else {
    fail('No supported archive module installed. Please install camptocamp/archive or nanliu/archive.')
  }
}
```

I could be the only one who wants to do something like this (I could also be nuts to want to do this). I just thought I'd submit it and see what the discussion was. If there's interest, I'll write some tests and docs for this PR. Also, if there are better ways to implement this, I'd love to hear how!

danzilio:
Based on some feedback, I just added the ability to pass an array of parameters. The example above would now look like this:

```
class fetch (
  $filename = 'apache-tomcat-6.0.26',
  $url = 'http://archive.apache.org/dist/tomcat/tomcat-6/v6.0.26/bin/apache-tomcat-6.0.26.tar.gz',
  $target = '/opt',
) {

  if respond_to('archive', ['url', 'target']) {
    archive { $filename:
      ensure => present,
      url    => $url,
      target => $target,
    }
  } 
  elsif respond_to('archive', ['source', 'extract_path']) {
    archive { $filename:
      ensure        => present,
      source        => $url,
      extract_path  => $target,
    }
  }
  else {
    fail('No supported archive module installed. Please install camptocamp/archive or nanliu/archive.')
  }
}
```

purpleidea:
@danzilio Cool idea, but (and maybe I'm misunderstanding) but what if the other module implements the parameters that you're checking for in the first... There's no guarantee this won't break!

IOW, cool idea, but this might end up being more trouble than it's worth.
Now if modules were fully qualified, eg: danzilio::archive, then that would solve your problem I think.

Since that might not change, perhaps each module could support a modulename::namespace class value which would equal danzilio or whatever the module namespace uuid is... Then you could test based on that.

HTH

danzilio:
That's a good point. This function ultimately requires some knowledge of the other classes/resources you're using. I think this would work great with some kind of soft dependency mechanism in the module metadata. @garethr talked about some kind of interface system in a talk a few years ago. The example from his slides:

```
interface packageandservice {
  $version
  $enable
  $package_name
  $template
}
```

With the invocation looking something like this:

```
class diamond (
  $version,
  $enable,
  $package_name,
  $template,
) {
  implements packageandservice
```

This kind of functionality would obviate the need for this function because I could (I assume) just check to see what interface the class is implementing. I'd love to see something like that.

hlindberg:
While I think it is useful to have reflection on types (i.e. being able to get information about their structure, attributes, etc.) I do not think that using such mechanisms to achieve Duck Typing (Ruby style - checking one 'responds_to', or Go style - checking multiple parameters) is going to be very successful in terms of maintainability. 

I would want the check to be based on the type - either the implementation type, or an interface that an implementation type has declared that it implements. I can imagine a function `module_of(type)` that returns the module name of the type. Then the decision can be based on that. Unfortunately this requires using weak (i.e. unlisted) dependencies since module meta data cannot describe "requires one of ..." or depend on "a module that defines the resource type 'archive'" (and what defines archive BTW, and how do you depend on that?).

To me `module_of` makes it possible to reference information that should have been available and part of the name/identifier in the first place. (This function would return names like 'camptocamp::archive', 'nanliu::archive')

When adding type reflection, I think the method in question is better named `has_attr` since `respond_to` is really asking about the type's behavior, not its structure. And if we start using reflection we probably want to be able to get all attributes, their names and types. An alternative implementation would be to implement `attributes(type)` which produces a hash of name => type mappings. It is then possible to check if that hash has a specific key, iterate over it etc. While more powerful it is still not something I recommend for reasons of maintainability - reflective code is best suited for conduits, error checking, logging, and other similar aspects. 

I do not want generic functions like these in stdlib as they would squat on behavior that I would like the puppet type system to have. I would prefer if they where implemented using the 4.x function API and name-spaced so they do not clash in the future. I also think such functions, if not designed to handle everything related to type reflection, they should be in a separate module, not in stdlib.

danzilio:
@hlindberg you've articulated exactly what was in my head at the time I wrote this. There's definitely an inherent maintainability hit with this, but I think this is a rational design decision in many cases. Obviously, there are a number of ways to make this more maintainable. 

With the `module_of` implementation, the metadata would need to describe the types implemented in that module, as well as the classes inside the module (I think listing the classes inside the module would be very helpful in general). With this, we could implement type-based 'soft' dependencies (i.e. my module needs `example` class that implements `example` interface). 

```json
"dependencies": [
  { "class": "archive", "type_requirement": "somerequiredinterface" },
  { "name": "puppetlabs/stdlib", "version_requirement": ">=3.2.0 <5.0.0" }
]
```

I think 'soft' dependencies are desirable but, as of now, there's no way for a module to tell what implementation is available to it. If there was some `optional_dependency` object in `metadata.json` that listed prioritized optional dependencies (one must exist, if none exist, the first in the list is installed by `puppet module`). 

```json
"optional_dependencies": [
  { "name": "archive", "options": [
    { "name": "nanliu/archive", "version_requirement": ">=0.1.6 <0.2.2" },
    { "name": "camptocamp/archive", "version_requirement": ">=0.3.1 <0.5.2" }
  ]}
]
```

This would definitely make `puppet module` a more robust package management solution, but I'm not sure how many people are interested in this functionality.

I really like the idea of the `attributes(type)` implementation. I'm going to take a stab at it, but it sounds like this is something you'd rather keep out of stdlib. Until something like this is implemented in the type system, perhaps this would be a good candidate for extlib?

igalic:
my thought was that when modules were name-spaced in puppet/environments, then we could "simply" fall-back on the dependencies from metadata.json

so, if puppetlabs/tomcat requires nanliu/archive, we could just use that.
now, lesaux/kibana4 can still depend on camptocamp/archive, and we can use both modules without too much headache.

of course, as @hlindberg says, this is "weak", and the question remains what to do with modules which don't even list dependencies, but just use them.

hlindberg:
The new loader system bases visibility on the module dependencies; a module can only see what it has declared dependencies on. We have made modules that have no dependencies what so ever exempt from this rule (the see everything else). The intent is to deprecate the exemption and make this strict in a future version (i.e. the configuration gradually gets more reliable). In 4.0.0 (and 3.x with future parser), only the 4x function API makes use of the new loaders. The intent is to move the rest of the loaded entities over to the new loaders.

I am all for being able to describe everything in module metadata. Basically, it requires two things: modules declare what they require and what the provide. Each such element is a "capability" (borrowing a term from a system that works this way) that has a namespaced type, a name within that type, and a version range. The namespaced type follows the principles of "puppet" that is reserved for puppet core, and "puppet_x::author" is reserved per author. 

Flexibility is needed since it may not be possible to update all modules; i.e. I could know something about a module that I would like to make available as meta data and I want to re-export it (make it visible to others even if there was no declaration of this in the original module) - say that I invent an interface and want to describe that I know that X and Y comply with this interface. (OTOH, features like this makes it a lot more complex).

Most of this metadata can be derived from the content in a module when a module is built and this can take into account if something is private to the module (and then not publish it).

Naturally we only need a very simple scheme at the start, but I think it is a big mistake in nor providing mechanisms for proper name spacing - we are still only at the beginning of the evolution of the puppet echo-system (and see how much pain the lack of proper naming has caused to date).

wrt. to the function `attributes` (or whatever it ends up being), I can imagine that a set of well designed functions for type reflection end up in puppet core; this includes relection on types themselves and ability to get the type parameters (attributes being one such, the length of an array, the element type of an array examples of other attributes of types).

... ok, I am rambling...

hlindberg:
why is this robot pinging this thread over and over again?

nibalizer:
Sorry! Thats 'my' bot. We're gonna fix it to not spam but haven't done that yet. This is where we are tracking it: https://github.com/puppet-community/pcci/issues/4

purpleidea:
On Wed, Jun 3, 2015 at 4:49 PM, Spencer Krum <notifications@github.com>
wrote:

> Sorry! Thats 'my' bot. We're gonna fix it to not spam but haven't done
> that yet. This is where we are tracking it: puppet-community/pcci#4
> <https://github.com/puppet-community/pcci/issues/4>
>


Pretty easy to stop the spamming by disabling it until your bot is fixed :)

nibalizer:
It's been disabled.

mhaskel:
Hi @danzilio , thanks for your work on this. This seems like it could be useful, but we don't want to include it in stdlib so I'd suggest publishing it as a separate module if you're still interested in this. Thanks again!

PR #443, opened 2015-04-21
Prep work for new specs

DavidS:
These are a few small commits to get the code base into shape for the new specs I'm working on.


DavidS:
Also contains a few fixes to functions that should go out with the next release, no matter what.

lib/puppet/parser/functions/pw_hash.rb
```diff
@@ -42,15 +42,13 @@
     if 'test'.crypt('$1$1') != '$1$1$Bp8CU9Oujr9SSEw53WV6G.'
       # JRuby < 1.7.17
       if RUBY_PLATFORM == 'java'
-        # override String#crypt for password variable
-        def password.crypt(salt)
-          # puppetserver bundles Apache Commons Codec
-          org.apache.commons.codec.digest.Crypt.crypt(self.to_java_bytes, salt)
-        end
+        # puppetserver bundles Apache Commons Codec
+        org.apache.commons.codec.digest.Crypt.crypt(password.to_java_bytes, salt)
```
elyscape: So this breaks the function completely.  More in 15 minutes.

elyscape: So, with this change, `salt` is no longer defined anywhere.  Why was this change made?  The commit message is "avoid ruby magic when running on java".  However, while the runtime may be Java, we're still running Ruby code, so "ruby magic" still works fine.

Here's the execution flow of this function before this change:

1. Check if `String#crypt` supports enhanced salts.  If it does not:
  1. If we're in Puppet Server, redefine the `crypt` function for the `password` variable to use the implementation provided by Apache Commons Codec.
  2. Otherwise, fail.

2. Call the `crypt` function on `password` with the argument `"$#{hash_type}$#{args[2]}"`.

Here's execution flow of this function after the change:

1. Check if `String#crypt` supports enhanced salts.  If it does not:
  1. If we're in Puppet Server, blow up by trying to use a variable that has not been defined (`salt`).
  2. Otherwise, fail.

2. If `String#crypt` supports enhanced salts, call the `crypt` function on `password` with the argument `"$#{hash_type}$#{args[2]}"`.

One way to fix this without using "ruby magic" would be to change the new line 46 to this:
```ruby
org.apache.commons.codec.digest.Crypt.crypt(password.to_java_bytes, "$#{hash_type}$#{args[2]}")
```
That being said, the reason I wrote the function the way I did is that it avoids having to duplicate specifying what the salt is.  If a change is made in the future (e.g. the salt is assembled ahead of time or something), the change would have to be made in two places.  This way, that was not necessary.

Possible fixes here are to either:

1. Revert the commit that made this change.
2. Change line 46 as mentioned above.
3. Assemble the salt before doing any checking and feed that into line 52.

lib/puppet/parser/functions/range.rb
```diff
@@ -41,29 +41,9 @@ module Puppet::Parser::Functions
...
       "arguments given (#{arguments.size} for 1)") if arguments.size < 1
 
-    if arguments.size > 1
-      start = arguments[0]
-      stop  = arguments[1]
-      step  = arguments[2].nil? ? 1 : arguments[2].to_i.abs
-
-      type = '..' # We select simplest type for Range available in Ruby ...
-
-    elsif arguments.size > 0
```
elyscape: Why was support for `range('2..3')` removed?

lib/puppet/parser/functions/range.rb
```diff
@@ -41,29 +41,9 @@ module Puppet::Parser::Functions
...
-      value = arguments[0]
-
-      if m = value.match(/^(\w+)(\.\.\.?|\-)(\w+)$/)
```
elyscape: Why was support for `range('2..3')` and `range('2...3')` and `range('2-3')` removed?

lib/puppet/parser/functions/range.rb
elyscape: Notice that this checks for `arguments.size` to be *greater than* 0, which at this point means it's 1.  This commit removed non-dead code.

PR #444, opened 2015-04-27
(MODULES-1982)Create a tcp_conn_validator resource

Spredzy:
Some services might take a while to start, even though the service
returned the boot process is finish, the process is not yet actually
listening on the port. This resource would allow to ensure a service is
really up and running before continuing with the application of the
catalog. Other project already rely on a similar feature. [1][2]

[1]
https://github.com/puppetlabs/puppetlabs-mongodb/blob/master/lib/puppet/type/mongodb_conn_validator.rb
[2]
https://github.com/puppetlabs/puppetlabs-puppetdb/blob/master/lib/puppet/type/puppetdb_conn_validator.rb

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -97,6 +97,26 @@ If you want to use a standardized set of run stages for Puppet, `include stdlib`
  * `name`: An arbitrary name used as the identity of the resource.
  * `path`: The file in which Puppet will ensure the line specified by the line parameter.
 
+* `tcp_conn_validator`: This resource ensures a service (remote or not) is actually up and running before moving onward with the catalog application. Puppet will block until the tcp connection can be made. If no connection are possible after a certain amount of time this resource will fail.
+
+  ```
+  tcp_conn_validator { 'mysql' :
+    server => '192.168.0.42',
+    port   => '3306',
+  }
+  ```
+
+To validate the same service running on differents nodes one can do :
+
+  ```
+  $array_of_service = ['ip1:3306', '[::1]:3306', 'ip2:3306']
+  tcp_conn_validator { $array_of_service : }
+  ```
+
+  * `server` : An IP or array of IP address of the server to check. Required.
+  * `port` : The port one want to ensure a process is listening on. Required.
+  * `timeout` : Number of second before timing out. Optional.
+
 ### Functions
 
 #### `abs`
--- a/lib/puppet/type/tcp_conn_validator.rb
+++ b/lib/puppet/type/tcp_conn_validator.rb
@@ -1,0 +1,44 @@
+Puppet::Type.newtype(:tcp_conn_validator) do
+
+  @doc = "Verify that a TCP connection can be successfully established between a node
+          and the expected  server.  Its primary use is as a precondition to
+          prevent configuration changes from being applied if the server cannot be
+          reached, but it could potentially be used for other purposes such as monitoring."
+
+  ensurable do
+    defaultvalues
+    defaultto :present
+  end
+
+  newparam(:name, :namevar => true) do
+    desc 'An arbitrary name used as the identity of the resource.'
+  end
+
+  newparam(:server) do
+    desc 'An array containing DNS names or IP addresses of the server where the expected service should be running.'
+    defaultto '127.0.0.1'
+    munge do |value|
+      Array(value).first
+    end
+  end
+
+  newparam(:port) do
+    desc 'The port that the server should be listening on.'
+    defaultto '22'
+  end
+
+  newparam(:timeout) do
+    desc 'The max number of seconds that the validator should wait before giving up and deciding that the service is not running; defaults to 60 seconds.'
+    defaultto 60
+
+    validate do |value|
+      # This will raise an error if the string is not convertible to an integer
+      Integer(value)
+    end
+
+    munge do |value|
+      Integer(value)
+    end
+  end
+
+end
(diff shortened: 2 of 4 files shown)


lib/puppet/util/tcp_validator.rb
```diff
@@ -0,0 +1,36 @@
+require 'socket'
+require 'timeout'
+require 'ipaddr'
+
+module Puppet
+  module Util
+    class TcpValidator
```
igalic: is there a reason you're putting this into the puppet::util namespace, rather than just creating our own?
it kinda feels… dirty…

Spredzy: No special reason. I suspected it was a pattern because both puppetlabs-mongodb[1] and puppetlabs-puppetdb[2] uses it.

[1] https://github.com/puppetlabs/puppetlabs-mongodb/blob/master/lib/puppet/util/mongodb_validator.rb
[2] https://github.com/puppetlabs/puppetlabs-puppetdb/blob/master/lib/puppet/util/puppetdb_validator.rb

README.markdown
```diff
@@ -97,6 +97,19 @@ If you want to use a standardized set of run stages for Puppet, `include stdlib`
...
+  ```
+  tcp_conn_validator { 'mysql' :
+    server => '192.168.0.42',
+    port   => '3306',
+  }
+  ```
+
+  * `server` : An IP or array of IP address of the server to check. Required.
+  * `port` : The port one want to ensure a process is listening on. Required.
+  * `timeout` : Number of second before timint out. Optional.
```
DavidS: 'timing'

lib/puppet/util/tcp_validator.rb
DavidS: It would be great to move that to lib/puppetx/puppetlabs, as per https://github.com/puppetlabs/puppet/blob/master/lib/puppet_x.rb

Spredzy: @DavidS so you want this function to be in PuppetX::Util::TcpValidator scope ?

DavidS: The puppet_x.rb suggests to use `puppetx/moduleowner` as namespace. `PuppetX::Puppetlabs::TcpValidator` would sound good to me.

DavidS: Sorry, that should have been `puppet_x/moduleowner`

Spredzy:
@DavidS PR updated to reflect the namespace change. Is that ok? Should the path to the TcpValidator class should be change or is it fine as it is?

DavidS:
Uhm, yeah, that would be great. File names should almost always match their contents.

Spredzy:
@DavidS done.

nibalizer:
We're thinking a new module for this, discussion: https://groups.google.com/forum/#!topic/puppet-dev/5hZcxDk6pjk

lib/puppet/type/tcp_conn_validator.rb
```diff
@@ -0,0 +1,44 @@
+Puppet::Type.newtype(:tcp_conn_validator) do
+
+  @doc = "Verify that a TCP connection can be successfully established between a node
+          and the expected  server.  Its primary use is as a precondition to
+          prevent configuration changes from being applied if the server cannot be
+          reached, but it could potentially be used for other purposes such as monitoring."
+
+  ensurable do
```
adrienthebo: Since this resource doesn't really have an actual "state" I don't think it should be ensurable - having something like a "reachable => true/false" seems like a better approach to me.

lib/puppet/type/tcp_conn_validator.rb
```diff
@@ -0,0 +1,44 @@
...
+    desc 'An array containing DNS names or IP addresses of the server where the expected service should be running.'
+    defaultto '127.0.0.1'
+    munge do |value|
+      Array(value).first
+    end
+  end
+
+  newparam(:port) do
+    desc 'The port that the server should be listening on.'
+    defaultto '22'
```
adrienthebo: Is it possible to have a meaningful default for the port, or should the user have to specify a port to connect to every time?

lib/puppet/type/tcp_conn_validator.rb
nibalizer: Would reachable => false, mean 'make sure this thing is down before proceeding?'

lib/puppet/type/tcp_conn_validator.rb
nibalizer: User should have to specify a default every time imho

lib/puppet/type/tcp_conn_validator.rb
adrienthebo: Well, either those are the semantics or remove this sort of "ensurable" or "reachable" parameter entirely and indicate that this check must always pass. As it stands, I think that this resource will fail regardless of if it's present or absent.

nibalizer:
Can you submit this to puppetcommunity/puppet-healthcheck?

Spredzy:
@nibalizer will do. Will catch up with the thread in the ML also.

Spredzy:
As asked by @nibalizer and announced in the thread on puppet-dev this PR has been made against puppet-community/puppet-healtcheck module https://github.com/puppet-community/puppet-healthcheck/pull/1

Closing it here.

igalic:
\o/

PR #445, opened 2015-04-27
(MODULES-1982)Create a http_conn_validator resource

Spredzy:
Some services might take a while to start, even though the service
returned the boot process is finish, the process is not yet actually
listening on the port. This resource would allow to ensure a service is
really up and running before continuing with the application of the
catalog. Other project already rely on a similar feature. [1][2]

[1]
https://github.com/puppetlabs/puppetlabs-mongodb/blob/master/lib/puppet/type/mongodb_conn_validator.rb
[2]
https://github.com/puppetlabs/puppetlabs-puppetdb/blob/master/lib/puppet/type/puppetdb_conn_validator.rb

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -97,6 +97,23 @@ If you want to use a standardized set of run stages for Puppet, `include stdlib`
  * `name`: An arbitrary name used as the identity of the resource.
  * `path`: The file in which Puppet will ensure the line specified by the line parameter.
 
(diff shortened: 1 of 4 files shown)


cyberious:
If we are talking about a service listening on a port, how about a more generic netstat with port to see if it is listening?

Spredzy:
@cyberious purpose here is to be able to check remote hosts also. netstat would work only locally unless there is an option I am not aware about. Final goal is to try - to some extent - to do inter node orchestration of some sort when nodes are provisioned in //

Spredzy:
As asked by @nibalizer and announced in the thread on puppet-dev this PR has been made against puppet-community/puppet-healtcheck module https://github.com/puppet-community/puppet-healthcheck/pull/2

Closing it here

PR #446, opened 2015-05-05
Fix pw_hash() on JRuby < 1.7.17

elyscape:
The previous change to this function broke it on JRuby before 1.7.17 by
attempting to use a variable that wasn't defined (`salt`). To fix this,
define `salt` ahead of time and use that instead of building the salt
later.

cf. https://github.com/puppetlabs/puppetlabs-stdlib/pull/443#discussion_r29718588


DavidS:
Thanks for catching that.

PR #447, opened 2015-05-05
Restore removed functionality to range()

elyscape:
This restores functionality removed from `range()` by a recent commit, specifically the ability to make calls such as `range('2..3')`, `range('2...3')`, and `range('2-3')`.  Such removal should be done only on a new major version due to it being a breaking change and should be documented correctly.  I suspect this was done due to misreading of [one of the removed checks](https://github.com/puppetlabs/puppetlabs-stdlib/pull/443#discussion_r29721193), so I have also cleaned up the function a bit to clarify it. 

cf. https://github.com/puppetlabs/puppetlabs-stdlib/pull/443#commitcomment-11055565

### DIFF
--- a/lib/puppet/parser/functions/range.rb
+++ b/lib/puppet/parser/functions/range.rb
@@ -37,18 +37,17 @@ Will return: [0,2,4,6,8]
     EOS
   ) do |arguments|
 
-    # We support more than one argument but at least one is mandatory ...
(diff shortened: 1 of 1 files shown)


DavidS:
:+1: for your psychic debugging skills and fixing the problem. Thanks!

PR #450, opened 2015-05-07
Added replace attribute

WhatsARanjit:
We have a use case where a Solaris /etc/logadm.conf file has entries like:

```
/var/adm/wtmpx -C 3 -P 'Sun Apr 26 07:10:04 2015' -g adm -m 644 -o adm -p 1m -t '/var/adm/oldlogs/wtmpx-$nodename-%Y%m%d' -z 0
```

Using the shell command to add entries by design puts a timestamp in the line.  A customer would like to manage lines of this file.  Using file_line and the match attribute causes the line to be continually updated with new timestamp.

```
$timestamp = chomp(generate('/bin/date', '+%a %b %d %T %Y'))
$wtmpx_regex = '\/var\/adm\/wtmpx \-C \d+ \-P \'[\w\s:]+\''

file_line { 'wtmpx':
  path    => '/tmp/logadm.conf',
  line    => "/var/adm/wtmpx -C 3 -P '${timestamp}' -g adm -m 644 -o adm -p 1m -t '/var/adm/oldlogs/wtmpx-\$nodename-%Y%m%d' -z 0A",
  match   => $wtmpx_regex,
}
```
The time stamp will be different for each system because it is the time the entry was added.

### DIFF
--- a/lib/puppet/provider/file_line/ruby.rb
+++ b/lib/puppet/provider/file_line/ruby.rb
@@ -1,7 +1,11 @@
 Puppet::Type.type(:file_line).provide(:ruby) do
   def exists?
-    lines.find do |line|
-      line.chomp == resource[:line].chomp
+    if resource[:replace] and match_count > 0
+      return true
+    else
+      lines.find do |line|
+        line.chomp == resource[:line].chomp
+      end
     end
   end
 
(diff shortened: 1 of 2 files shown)


WhatsARanjit:
JIRA ticket:
https://tickets.puppetlabs.com/browse/MODULES-2024

elyscape:
This PR and the issue need some work. I'll elaborate on my concerns when I get back to a keyboard.

WhatsARanjit:
My base was off.  I'll re-PR.

WhatsARanjit:
Resubmitted in https://github.com/puppetlabs/puppetlabs-stdlib/pull/450

PR #451, opened 2015-05-07
Added replace attribute

WhatsARanjit:
We have a use case where a Solaris /etc/logadm.conf file has entries like:

```
/var/adm/wtmpx -C 3 -P 'Sun Apr 26 07:10:04 2015' -g adm -m 644 -o adm -p 1m -t '/var/adm/oldlogs/wtmpx-$nodename-%Y%m%d' -z 0
```

Using the shell command to add entries by design puts a timestamp in the line. A customer would like to manage lines of this file. Using file_line and the match attribute causes the line to be continually updated with new timestamp.

```
$timestamp = chomp(generate('/bin/date', '+%a %b %d %T %Y'))
$wtmpx_regex = '\/var\/adm\/wtmpx \-C \d+ \-P \'[\w\s:]+\''

file_line { 'wtmpx':
  path    => '/tmp/logadm.conf',
  line    => "/var/adm/wtmpx -C 3 -P '${timestamp}' -g adm -m 644 -o adm -p 1m -t '/var/adm/oldlogs/wtmpx-\$nodename-%Y%m%d' -z 0A",
  match   => $wtmpx_regex,
}
```

The time stamp will be different for each system because it is the time the entry was added.

### DIFF
--- a/lib/puppet/provider/file_line/ruby.rb
+++ b/lib/puppet/provider/file_line/ruby.rb
@@ -1,7 +1,11 @@
 Puppet::Type.type(:file_line).provide(:ruby) do
   def exists?
-    lines.find do |line|
-      line.chomp == resource[:line].chomp
+    if resource[:replace] and count_matches(match_regex) > 0
+      true
+    else
+      lines.find do |line|
+        line.chomp == resource[:line].chomp
+      end
     end
   end
 
@@ -32,10 +36,13 @@ Puppet::Type.type(:file_line).provide(:ruby) do
     @lines ||= File.readlines(resource[:path])
   end
 
+  def match_regex
+    resource[:match] ? Regexp.new(resource[:match]) : nil
+  end
+
   def handle_create_with_match()
-    regex = resource[:match] ? Regexp.new(resource[:match]) : nil
     regex_after = resource[:after] ? Regexp.new(resource[:after]) : nil
-    match_count = count_matches(regex)
+    match_count = count_matches(match_regex)
 
     if match_count > 1 && resource[:multiple].to_s != 'true'
      raise Puppet::Error, "More than one line in file '#{resource[:path]}' matches pattern '#{resource[:match]}'"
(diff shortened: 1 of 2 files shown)


WhatsARanjit:
https://tickets.puppetlabs.com/browse/MODULES-2024

elyscape:
As I mentioned on the previous PR, this PR and the issue need some work. I'll elaborate on my concerns when I get back to a keyboard.

elyscape:
In the meantime, instead of using `file_line` to manage your `logadm.conf`, I'd recommend instead using something [like this custom `logadm` type](https://github.com/elatov/puppet/blob/master/logadm/lib/puppet/type/logadm.rb).

WhatsARanjit:
@elyscape Yeah saw that, it could be a type that works.  But in absence of a provider and spec testing I opted to add an attribute to `file_line`.  I'll look into it though.

elyscape:
Not all types use providers.  That custom `logadm` type is an example of such a type.

nibalizer:
It seems to me like the problem here is managing logadm.conf. Logadm, it seems, is one of those terrible daemons that writes back to its own config file. However, I think I have a solution. From the logadm docs:

```
By specifying -p now as a rotation condition, a log rotation is forced.
```
Setting -p now and removing -P <timestamp> then all files will be rotated whenever logadm is invoked by cron. This means that cron can handle scheduling and puppet can manage logadm.conf idempotently.

DavidS:
Hi @WhatsARanjit ,

thank you for your patch, I think it's a worth-while addition.

To be acceptable to stdlib, this needs at least a short description in the README and a unit test in `spec/unit//puppet/provider/file_line/ruby_spec.rb`.

Like the `file#replace` parameter, this should default to `true` to keep existing configurations running. Also, I have the impression, that the replace as implemented works exactly inverse to what the file type's version does. Please verify that too.

Cheers, D.

mhaskel:
@WhatsARanjit ping? any update on this?

bmjen:
Merged in #494

PR #452, opened 2015-05-08
(MODULES-2026) Support hashes in suffix function

mld:
The puppetlabs/stdlib module already supports hashes in the prefix function. This patch adds the same functionality to suffix. It's basically the prefix function with s/prefix/suffix/.

https://tickets.puppetlabs.com/browse/MODULES-2026

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -640,7 +640,7 @@ This function removes leading and trailing whitespace from a string or from ever
 
 #### `suffix`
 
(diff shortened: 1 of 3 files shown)


igalic:
@mld this will need a [rebase](http://git-scm.com/book/en/Git-Branching-Rebasing)

hunner:
This can now be done in Puppet 4's DSL, and in any way that a user would prefer rather than an inflexible function. Thanks for the work though!

PR #453, opened 2015-05-09
(MODULES-2031) Add 'url_encode' function to stdlib

adamcin:
We recently upgraded our stdlib module to 4.6 and discovered that one of
our modules depended on legacy behavior of the uriescape function that
had changed significantly along the way:

https://github.com/puppetlabs/puppetlabs-stdlib/pull/164
(commit: b2e23dc65bd9851dcb6ad60ffca1acbc70b617e1)

Before PR#164, the effective behavior of uriescape was to replace
characters that are illegal or significant within URIs with
percent-delimited, hex-encoded equivalents, as is necessary when
encoding a free-text UTF-8 string within a URL path segment ( / "text"
    / ) such that any forward-slashes, ampersands, question marks, and
octothorpes embedded within the text do not change the structure of
the URL from the perspective of a URL parser.

After PR#164, the effective behavior of uriescape changed, for the
most part, to minimally escape a complete URL (agnostic of its parts)
in such a way to make it safe to pass as a double-quoted argument to a
shell command like curl, to prevent command injection.

We still need the legacy behavior for building URIs, such as JDBC URLs
in config files that embed a password value from hiera containing
URI-significant characters like '&' or '?'.

Since the uriescape function has been adjusted to match the behavior
of the Ruby URI.escape function, a simple enhancement would be to
expose one of the recommended alternative Ruby functions
(ERB::Util.url_encode), as a distinct puppet function.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -682,7 +682,21 @@ Converts an object, array or hash of objects that respond to upcase to uppercase
 
 #### `uriescape`
 
-Urlencodes a string or array of strings. Requires either a single string or an array as an input. *Type*: rvalue
+Escapes whole-URL strings (single string or an array of strings) for use as double-quoted arguments to shell commands and HTML attributes by URL-encoding unsafe characters like space, double-quote, octothorpe, and curly braces.
+
+This function is ideal for sanitizing untrusted whole-URL string inputs to prevent command or HTML injection without destroying their validity. In other words, a safe URL passed to this function will be unchanged, in contrast to the `url_encode` function, which will turn a valid URL into a completely inert string containing only alphanums and percents.
+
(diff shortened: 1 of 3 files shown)


nibalizer:
This is in the class of 'bubble up ruby libraries to puppet' via functions in stdlib. We have quite a lot of these. Can't we get both tasks(uriencode and urlencode) of these done with one function? This feels bloaty. Is there a way forward that doesn't just 'throw more crap in stdlib'

elyscape:
A better option might be an optional parameter to feed into `URI.escape` to allow a customized set of characters to escape.  Though we should perhaps think about changing to some other implementation, as `URI.escape` is deprecated.  Unfortunately, [there doesn't seem to be a particularly good replacement that doesn't require a gem](http://stackoverflow.com/a/13059657).

hunner:
Due to inactivity and not coming to a consensus, I'm closing this PR. Feel free to reopen it with further comments.

PR #454, opened 2015-05-11
(FM-2130) Document new location of facts.d cache

elyscape:
b11311a moved the location of the facts.d cache but neglected to change the comment at the top of the file indicating its location. This rectifies that.


adrienthebo:
:+1:

PR #455, opened 2015-05-12
Remove all the pops stuff

hunner:
The rspec-puppet matchers don't allow to check the return types, but
this is a pretty rare thing to need to do anyway, so probably not worth
patching rspec-puppet


mhaskel:
@hunner merged, thanks!

PR #457, opened 2015-05-18
[MODULES-2048] Add a validate_supported_os() function

raphink:


nibalizer:
I'm thinking this isn't a good idea. There aren't docs so I'm not entirely sure what the usecase is. However, I don't see any advantage over putting a series of case statements with fail() line in params.pp or something. Determining os suitability is essentially walking a little tree down the paths you need to evaluate. Thus function has no recursive component, and I'm not sure it should. It also has the problem that it requires a module to be listed as supported, that means that nothing can work on new releases of distros until the module author gets around to updating metadata.json

-1

igalic:
the "documentation" (and further discussion) is in the [Jira](https://tickets.puppetlabs.com/browse/MODULES-2048).
****
okay, so the author has to update the metadata.json, but when the `fail()` is in params.pp it's okay?

what?

nibalizer:
@igalic if you have in params.pp:

```puppet
case $::osfamily {
  /Debian/: { stuff }
  /RedHat/: { stuff }
  default: { fail("wont work here")}
}
```
Then this will work on all fedoras and all centoses. Under this patch, any new centos that comes out will need to be explicitly added to manifest.json. Its a question of implicitly supporting versus implicitly unsupporting, imho

igalic:
@nibalizer from what i gather, from @raphink's patch, only if those are actually mentioned in metadata.json, will they be evaluated.

nibalizer:
How about building a function that loads metadata.json wholesale? Then any user can validate however they need. Also can infer module version and author.

Example

```puppet

$metadata = load_metadata_json($mod_name)

// do something with $metadata['something']['something']

```

igalic:
related to this discussion is the one going on in #442

hunner:
Due to inactivity and not coming to a consensus, I'm closing this PR. Feel free to reopen it with further comments.

nibalizer:
https://github.com/puppetlabs/puppetlabs-stdlib/pull/483

PR #459, opened 2015-05-21
Add parser function "array_contains"

eirslett:
This function returns true if an element is
found in the array, false otherwise. It delegates
work to Ruby's underlying array.include? function.

One of the use cases for this function is where
a list of master nodes is defined, and a manifest
looks at its $fqdn to determine whether it's part
of that master list, in which case it should a
process with slightly different startup parameters.


mhaskel:
Hi @eirslett, thanks for the contribution!  It looks like this functionality is already in puppet with the `in` [expression](https://docs.puppetlabs.com/puppet/latest/reference/lang_expressions.html#in), so I'm going to close this PR. Thanks!

PR #461, opened 2015-05-27
validate_integer, validate_numeric: explicitely reject hashes in arrays

DavidS:
Without this patch, Ruby 1.8's Hash#to_s behaviour causes [{1=>2}] to be
treated as "12" when validating values.

### DIFF
--- a/lib/puppet/parser/functions/validate_integer.rb
+++ b/lib/puppet/parser/functions/validate_integer.rb
@@ -109,6 +109,7 @@ module Puppet::Parser::Functions
       # check every element of the array
       input.each_with_index do |arg, pos|
         begin
+          raise TypeError if arg.is_a?(Hash)
           arg = Integer(arg.to_s)
           validator.call(arg)
         rescue TypeError, ArgumentError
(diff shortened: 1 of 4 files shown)


lib/puppet/parser/functions/validate_integer.rb
```diff
@@ -109,6 +109,7 @@ module Puppet::Parser::Functions
       # check every element of the array
       input.each_with_index do |arg, pos|
         begin
+          raise TypeError.new if arg.is_a?(Hash)
```
elyscape: The `.new` shouldn't be necessary.

lib/puppet/parser/functions/validate_numeric.rb
```diff
@@ -71,6 +71,7 @@ module Puppet::Parser::Functions
       # check every element of the array
       input.each_with_index do |arg, pos|
         begin
+          raise TypeError.new if arg.is_a?(Hash)
```
elyscape: The `.new` shouldn't be necessary.

DavidS:
Indeed, thanks for the hint. Also, rebased.

nibalizer:
This gave me a big belly laugh. Also +1.

PR #462, opened 2015-05-29
fqdn_rotate: Don't use the value itself as part of the random seed

elyscape:
Previously, the random number generator was seeded with the array or string to be rotated in addition to any values specifically provided for seeding. This behavior is potentially insecure in that it allows an attacker who can modify the source data to choose the post-shuffle order. While this is probably not a particularly big issue, it's still worth changing.

This also improves the documentation for `fqdn_rotate`.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -259,7 +259,20 @@ fqdn_rand_string(10, '', 'custom seed')
 
 #### `fqdn_rotate`
 
-Rotates an array a random number of times, based on a node's fqdn. *Type*: rvalue.
+Rotates an array or string a random number of times, combining the `$fqdn` fact and an optional seed for repeatable randomness.
+
+*Usage:*
+~~~
+fqdn_rotate(VALUE, [SEED])
+~~~
+*Examples:*
+~~~
+fqdn_rotate(['a', 'b', 'c', 'd'])
+fqdn_rotate('abcd')
+fqdn_rotate([1, 2, 3], 'custom seed')
+~~~
+
+*Type*: rvalue.
 
 #### `get_module_path`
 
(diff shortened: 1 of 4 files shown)


DavidS:
Hi @elyscape , I think I like that change, but I'm having a hard time to put into words why this is needed. Would you mind enhancing the commit message to explain that?

elyscape:
@DavidS: Sure. Will do.

elyscape:
@DavidS: Done.

elyscape:
Hold up there's a bug in the spec tests.

elyscape:
Or not.  By chance, the rotation in [this test](https://github.com/elyscape/puppetlabs-stdlib/blob/fix/fqdn_rotate_seeds_with_argument/spec/functions/fqdn_rotate_spec.rb#L34-L37) is the same before and after the change.

That being said, question: why was [this test](https://github.com/puppetlabs/puppetlabs-stdlib/blob/f3e79ddcd56a221c7799b35efde7e9803a5c7923/spec/functions/fqdn_rotate_spec.rb#L8-L11) added?
```ruby
it {
  pending("Current implementation ignores parameters after the first.")
  is_expected.to run.with_params("one", "two").and_raise_error(Puppet::ParseError)
}
```
This has never been the case.  It has always accepted a seed in the second parameter.  In light of that, I have removed it.

elyscape:
Ah, looks like there are no spec tests for the custom seed functionality.  I'll add some real quick.

elyscape:
Okay, improved spec tests are in.  Improved acceptance tests will come in another PR that will also improve the acceptance tests for `fqdn_rand_string()`.

DavidS:
Thank you for this work!

PR #463, opened 2015-05-29
 (MODULES-2071) Patch file_line provider to use multiple with after

rmaika:
Without this, `file_line` resource without the `match` parameter but with the `after` param will throw an error if there are multiple matches for the after expression. This patch creates the handling for the `multiple` parameter in handle_create_with_after. This allows you to add a line after the `after` expression if it appears at multiple points in a file without specifying the `match` param.

Updated reference to `file_line` in the README to reflect that the multiple parameter can be set when using `after` and/or `match` as the matching regex.


DavidS:
Hi @rmaika , thank you for your contribution!

PR #473, opened 2015-06-09
WIP "Add fact for is_systemd"

nibalizer:


nibalizer:
Note that I don't think this actually works. I keep getting ruby string => float errors which confuse me. Looking for feedback.

elyscape:
Maybe you could check to see if the system had a package named systemd installed?

nibalizer:
Ya, or check if the systemctl utility is available. I'm not sure what the correct approach /should/ be. I do know that the approach currently used is mostly what people are doing inside their puppet manifests now, so might be easier to sell them on using the new fact.

DavidS:
The string->float errors are probably coming from stringify_facts?

To the actual detection of systemd being active on a node, please see the `systemctl is-system-running` command: http://www.freedesktop.org/software/systemd/man/systemctl.html#is-system-running 

I would expect that to give an error in the case of systemd being installed but not the actual PID1. This should be testable on a jessie system (Debian 8) that was forced to sysvinit without removing the systemd/systemctl tooling.

igalic:
i'd love to hear from anyone who's systemd system is not `degraded`…

elyscape:
`systemctl is-system-running` was added to systemd in version 215.  EL7 currently uses systemd 208, and many other distros likely use older versions as well, so we should probably figure out an alternate way to detect systemd.

PR #476, opened 2015-06-18
Add ensure_latest function.

hashbrowncipher:
This provides a convenient way to idempotently ask for a package, but also
makes it effortless to always get the latest version.  This prevents the
"version sprawl" enabled by ensure_packages, where machines in a fleet that
are provisioned over time get different package versions.  At the same time,
we are maintaining the familiar interface provided by ensure_packages.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -206,6 +206,10 @@ Converts the case of a string or of all strings in an array to lowercase. *Type*
 
 Returns 'true' if the variable is empty. *Type*: rvalue.
 
+#### `ensure_latest`
+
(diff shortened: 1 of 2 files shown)


elyscape:
I think a better option is to specify a version, or latest, when calling `ensure_packages()`, like so:
```puppet
ensure_packages(['openssl', 'bash'], {'ensure' => 'latest' })
ensure_packages('wget', {'ensure' => '1.12-5.el6_6.1' })
```

hashbrowncipher:
My goal here is to provide syntactic sugar.  `ensure_packages` is itself sugar for a call to `ensure_resources`.  In the same way, `ensure_latest` is meant as syntactic sugar for `ensure_packages`, but avoids a pain point by providing a different default.

hunner:
I think for the sake of simplicity, we're going to leave `ensure_packages()` as is and have that be the use case. You're welcome to update the documentation to indicate how to use `ensure => latest` with ensure_packages. Thanks though!

PR #480, opened 2015-07-06
Add validate_slength's optional 3rd arg to README

alexjfisher:
I was thinking about creating a validate_non_empty_string function or a validate_minimum_slength function when I discovered validate_slength could already take a minimum string length as its 3rd argument.

### DIFF
--- a/lib/puppet/parser/functions/validate_slength.rb
+++ b/lib/puppet/parser/functions/validate_slength.rb
@@ -43,9 +43,7 @@ module Puppet::Parser::Functions
       min_length = 0
     end
 
-    if min_length > max_length
-      raise Puppet::ParseError, "validate_slength(): Expected second argument to be equal to or larger than third argument"
-    end
+    raise Puppet::ParseError, "validate_slength(): Expected second argument to be equal to or larger than third argument" unless max_length >= min_length
 
     validator = lambda do |str|
       unless str.length <= max_length and str.length >= min_length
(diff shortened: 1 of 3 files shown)


lib/puppet/parser/functions/validate_slength.rb
```diff
@@ -44,7 +44,7 @@ module Puppet::Parser::Functions
     end
 
     if min_length > max_length
-      raise Puppet::ParseError, "validate_slength(): Expected second argument to be larger than third argument"
+      raise Puppet::ParseError, "validate_slength(): Expected second argument to be equal to or larger than third argument"
```
DavidS: That does not match the `>` in L46

alexjfisher: The second argument is 'max_length' and the third is 'min_length'.  Or are you suggesting reversing the comparison in L46 for clarity?

DavidS: Yeah, that threw me off. Text and code should always align.

alexjfisher:
@DavidS I've added a commit that hopefully improves the clarity of the code.

There is already a spec test for the corner case of min_length == max_length.
https://github.com/puppetlabs/puppetlabs-stdlib/blob/master/spec/functions/validate_slength_spec.rb#L54

alexjfisher:
BTW, what do you think about adding a validate_non_empty_string function?
It would be a lot neater that validate_slength("string",999999,1)

DavidS:
closing as merged in #482

PR #486, opened 2015-07-14
[MODULES-2234] - Add checksum_file function to compute file checksum

amuraru:


elyscape:
Instead of providing `md5_file`, why not provide `hash_file` that takes a filename and a hash algorithm to use?  That would make it so that one function could provide hashing for MD5, SHA-1/256/384/512, and whatever other algorithms might be useful.

amuraru:
@elyscape Eli, good call. Made the algo configurable. See the new version of the patch

lib/puppet/parser/functions/checksum_file.rb
```diff
@@ -0,0 +1,40 @@
...
+        else
+          file = nil
+        end
+      end
+    end
+    if args.length == 2
+      algo = args[1]
+    else
+      algo = Puppet[:digest_algorithm]
+    end
```
DavidS: This `if` can be reduced to `algo = args[1] || Puppet[:digest_algorithm]`.

lib/puppet/parser/functions/checksum_file.rb
```diff
@@ -0,0 +1,40 @@
...
+    Puppet::Util::Checksums.method("#{algo}_file").call(file)
```
DavidS: Using `send` (http://ruby-doc.org/core-2.2.2/Object.html#method-i-send) here will save you another allocation and call.

lib/puppet/parser/functions/checksum_file.rb
```diff
@@ -0,0 +1,40 @@
...
+
+module Puppet::Parser::Functions
+
+  newfunction(:checksum_file, :type => :rvalue, :doc => <<-'ENDHEREDOC') do |args|
+    Computes the MD5 checksum for the given file.
+
+    For example:
+
+        $myhash = checksum_file('/etc/puppet/data/myhash.yaml')
+    ENDHEREDOC
```
DavidS: Please add a pointer or explanation here to which checksums are allowable.

lib/puppet/parser/functions/checksum_file.rb
```diff
@@ -0,0 +1,40 @@
...
+
+    unless args.length >= 1
+      raise Puppet::ParseError, ("checksum_file(): wrong number of arguments (#{args.length}; must be 1)")
```
DavidS: The error message doesn't match reality.

DavidS:
I like the basic premise, but there are some implementation issues and no unit tests. Please look at other examples in `spec/functions` on how to do unit testing.

daenney:
One thing I fail to understand, how does this work in a master/agent scenario? You'd be computing the checksum of a file on the master which is then useful in the agent how? The puppet file resource would already be doing all this for you.

lib/puppet/parser/functions/checksum_file.rb
```diff
@@ -0,0 +1,40 @@
+require 'puppet/parser/files'
+require 'puppet/util/checksums'
+
+module Puppet::Parser::Functions
+
+  newfunction(:checksum_file, :type => :rvalue, :doc => <<-'ENDHEREDOC') do |args|
+    Computes the MD5 checksum for the given file.
```
elyscape: This needs to be updated now that MD5 isn't the only available algorithm.

lib/puppet/parser/functions/checksum_file.rb
```diff
@@ -0,0 +1,40 @@
...
+
+    For example:
+
+        $myhash = checksum_file('/etc/puppet/data/myhash.yaml')
```
elyscape: This needs to be updated now that the checksum used can be specified.

DavidS:
@daenney maybe put that hash into a config, or something? No way to retrieve the hash from the file resource, too.

daenney:
Sure, but why would you care about the checksum of a file on the master? If you want to put in some checksum of a file in a config somewhere so that at startup a daemon can validate that the config file hasn't been tampered with (this I find a stretch already), you'd want to checksum the file on the agent, which is not where functions are executed.

mhaskel:
@amuraru thanks for the work on this! We haven't heard from you in a while so we're going to close this PR for now. We also had some questions about use cases (added to the MODULES ticket). If this is functionality you're still interested in having added to the module and you can address the questions in the ticket and comments in the PR please resubmit. Thanks again!

PR #487, opened 2015-07-20
Fix documentation error in upcase

liv3d:
The documentation example shows an incorrect response when using the
function, this PR corrects the example to agree with what the function
actually does.


DavidS:
good catch. thanks!

PR #489, opened 2015-07-21
adding support for hash in the size function

gcmalloc:


spec/functions/size_spec.rb
```diff
@@ -8,15 +8,18 @@
...
+  it { is_expected.to run.with_params('1').and_raise_error(Puppet::ParseError, /Requires either string, array or hash to work/) }
+  it { is_expected.to run.with_params('1.0').and_raise_error(Puppet::ParseError, /Requires either string, array or hash to work/) }
   it { is_expected.to run.with_params([]).and_return(0) }
   it { is_expected.to run.with_params(['a']).and_return(1) }
   it { is_expected.to run.with_params(['one', 'two', 'three']).and_return(3) }
   it { is_expected.to run.with_params(['one', 'two', 'three', 'four']).and_return(4) }
 
+  it { is_expected.to run.with_params({}).and_return{0} }
+  it { is_expected.to run.with_params({'1' => '2'}).and_return{1} }
+  it { is_expected.to run.with_params({'1' => '2', '4' => '4'}).and_return{2} }
```
DavidS: all of them should read `and_return(0)`

lib/puppet/parser/functions/size.rb
```diff
@@ -37,6 +37,8 @@ module Puppet::Parser::Functions
 
     elsif item.is_a?(Array)
       result = item.size
+    elsif item.is_a?(Hash)
```
DavidS: you can put this as `|| item.is_a?(Hash)` at the end of L38 and save two lines here.

DavidS:
Hi @gcmalloc,

I like this change and it only needs some small changes to get it ship shape. Please use `git commit --amend` this change so it stays a single commit and force push it to this PR.

Thanks for your time and work!

gcmalloc:
Thanks for the kindness !

PR #494, opened 2015-07-30
[#puppethack] Adding replace attribute to file_line

rmaika:
Adding tests and minor changes to make @WhatsARanjit's PR #451 work. 

The patch adds the optional `replace` attribute to file_line. It defaults to true, which maintains current behavior. Setting the attribute to false will allow users to ensure that a line matching the `match` attribute exists without overwriting it with the contents of `line` if the matching lines exists

### DIFF
--- a/lib/puppet/provider/file_line/ruby.rb
+++ b/lib/puppet/provider/file_line/ruby.rb
@@ -1,17 +1,23 @@
 Puppet::Type.type(:file_line).provide(:ruby) do
   def exists?
-    lines.find do |line|
-      line.chomp == resource[:line].chomp
+    if !resource[:replace] and count_matches(match_regex) > 0
+      true
+    else
+      lines.find do |line|
+        line.chomp == resource[:line].chomp
+      end
     end
   end
 
   def create
-    if resource[:match]
-      handle_create_with_match
-    elsif resource[:after]
-      handle_create_with_after
-    else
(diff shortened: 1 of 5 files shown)


lib/puppet/provider/file_line/ruby.rb
```diff
@@ -43,7 +52,7 @@ def handle_create_with_match()
 
     File.open(resource[:path], 'w') do |fh|
       lines.each do |l|
-        fh.puts(regex.match(l) ? resource[:line] : l)
+        fh.puts(match_regex.match(l) ? resource[:line] : l)
```
bmjen: If `match_regex` returns nil... this will result in a `NoMethodError`

bmjen: On second look, given this code path, you *should* be able to assume that `match_regex` is not `nil`.

rmaika: Yeah, the handle_create_with_match function will not be used to write the line unless resource[:match] is not nil. So the creation of the regex that match_regex returns should not fail in this block.

rmaika: it also isn't a new behavior with this patch. The line only changes because where the `Regexp` is created is moved and renamed from `regex` to `match_regex`

PR #496, opened 2015-07-30
Remove colorful language from module.

pizzaops:
Removed use of "crap" from documentation because I'm a buzzkill.


bodepd:
@pizzaops so corporate!

PR #499, opened 2015-08-06
[MODULES-2370] allow `match` parameter to influence `ensure => absent` behavior.

jearls:
Split the `destroy` method of the file_type::ruby provider into two
private methods:  `handle_destroy_line` which is the same as the previous
`destroy` method, and `handle_destroy_with_match` which will destroy any
line which matches the `match` parameter, raising an error if multiple
lines match and the `multiple` parameter is not `true`.  This new
behavior is only used if the new boolean parameter `match_for_absence`
is `true` (it defaults to `false`).


daenney:

PR #504, opened 2015-08-13
(MODULES-2410) Add new functions dos2unix and unix2dos

gibbsoft:
These two new functions are very useful when using a File resource with a cross-platform template.

~~~
file{$config_file:
  ensure  => file,
  content => unix2dos(template('my_module/settings.conf.erb')),
}

file{$config_file:
  ensure  => file,
  content => dos2unix(template('my_module/settings.conf.erb')),
}
~~~


DavidS:
Hi, thank you for your contribution. These functions look very useful. There are a few things to fix though to keep the module maintainable in the long run:

* rebase your fix on top of current master.
* add the :arity parameter to the functions to ensure that only a single argument can be passed.
* ensure that `args[0].is_a? String` and throw a nice error (like https://github.com/puppetlabs/puppetlabs-stdlib/blob/master/lib/puppet/parser/functions/base64.rb#L25) 
* add a few tests demoing the error behaviour (no args, too many args, not a string)
* please follow the style of the other tests (see https://github.com/puppetlabs/puppetlabs-stdlib/blob/master/spec/functions/empty_spec.rb for an example)

gibbsoft:
Thanks for your comments David, I've amended and rebased as you suggested.

DavidS:
Thanks for your quick response. The additions to the tests and functions look fabulous. The git history though, did take a turn for the worse. In the worst case, just extract the diff of your changes and create a completely new commit of it on top of a new checkout of the master.

PR #507, opened 2015-08-24
(MODULES-2456) Modify union to accept more than two arrays

Jetroid:
Add spec tests to test the new functionality:
 *Case for 3 arrays.
 *Case for 4 arrays.
Modify README to note new functionality.

This is for issue MODULE-2456, follow the precedent of MODULE-444.

This change allows union to be much more useful, unioning many arrays
in one line rather than in n lines. Additionally, as this is only added
functionality, and does not affect the 2 array case that all modules
currently using array are using, it should not affect any existing
modules utilizing union.

This is now useful, for example, for merging many arrays of resources
(eg: packages.) to generate just one list with no duplicates, to avoid
duplicate resource declarations.

### DIFF
--- a/lib/puppet/parser/functions/union.rb
+++ b/lib/puppet/parser/functions/union.rb
@@ -4,7 +4,7 @@
 
 module Puppet::Parser::Functions
   newfunction(:union, :type => :rvalue, :doc => <<-EOS
-This function returns a union of two arrays.
+This function returns a union of two or more arrays.
 
 *Examples:*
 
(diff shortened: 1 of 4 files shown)


lib/puppet/parser/functions/union.rb
```diff
@@ -14,18 +14,20 @@ module Puppet::Parser::Functions
     EOS
   ) do |arguments|
 
-    # Two arguments are required
-    raise(Puppet::ParseError, "union(): Wrong number of arguments " +
-      "given (#{arguments.size} for 2)") if arguments.size != 2
+    # Check that 2 or more arguments have been given ...
+    raise(Puppet::ParseError, "concat(): Wrong number of arguments " +
```
DavidS: you mean 'union', not 'concat' here?

lib/puppet/parser/functions/union.rb
```diff
@@ -14,18 +14,20 @@ module Puppet::Parser::Functions
...
+      raise(Puppet::ParseError, 'union(): Every parameter must be an array') unless argument.is_a?(Array)
     end
 
-    result = first | second
+    result = arguments[0]
+    arguments.shift
+
+    arguments.each do |argument|
+      result = result | argument
+    end
```
DavidS: L25-32 can be replaced by `arguments.reduce(:|)`.

DavidS:
Hi, this seems to be a worthwhile addition. There seems to be one typo in the error message and the actual function body could be simplified significantly. I've pointed that out in code comments.

Cheers!

Jetroid:
Thanks for the criticism, DavidS. 

Your comments are all valid. Have made the necessary changes and pushed.

DavidS:
Sweet. Would you mind squashing it into a single commit, just so the history is nice and clean?

http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

Jetroid:
No worries. Complete.

Thanks!

DavidS:

PR #508, opened 2015-08-24
MODULES-2457 - Remove newline from encoded string

mhaskel:


nibalizer:
lgtm

DavidS:
Would we be able to come up with an estimate how many people this change would trip over?

mhaskel:
@DavidS hm, not sure. I just saw the MODULES ticket came in and that seemed reasonable / bug-like.

DavidS:
Please consider #510 : adding the newlines is actually base64 conformant as the encoding specifies a maximum line length

mhaskel:
Merged #510 instead, if the newline needs to be stripped it should be done with `chomp()`

PR #509, opened 2015-08-24
Improve parseyaml function

dmitryilyin:
* Add default value support
* Update tests

### DIFF
--- a/lib/puppet/parser/functions/parseyaml.rb
+++ b/lib/puppet/parser/functions/parseyaml.rb
@@ -3,21 +3,20 @@
 #
 
 module Puppet::Parser::Functions
-  newfunction(:parseyaml, :type => :rvalue, :doc => <<-EOS
+  newfunction(:parseyaml, :type => :rvalue, :arity => -2, :doc => <<-EOS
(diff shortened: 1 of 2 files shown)


lib/puppet/parser/functions/parseyaml.rb
```diff
@@ -3,21 +3,20 @@
...
-        "given #{arguments.size} for 1")
-    end
-
     require 'yaml'
 
-    YAML::load(arguments[0])
-
+    data = YAML::load(arguments[0])
+    data = arguments[1] unless data
+    data
```
DavidS: L17-19 could just be `YAML::load(arguments[0]) || arguments[1]` .

DavidS:
Nice idea and good call on those additional specs :-) I've added a small comment on the style of the actual function body. The changes from the doc string need to be repeated in the README.

The main question though is why parseyaml needs a separate default value implementation when there is pick and pick_default ?

lib/puppet/parser/functions/parseyaml.rb
holser: +1

PR #515, opened 2015-08-27
Ticket/MODULES-2478 Make root_home fact work on AIX using native lsuser command

jfautley:
The root_home fact returns nil on AIX due to the operating system not providing getent. The recommended replacement for this is the 'lsuser' command.

This PR adds a new fact, confined to kernel=>aix to use libuser to return root's configured home directory.

### DIFF
--- a/lib/facter/root_home.rb
+++ b/lib/facter/root_home.rb
@@ -30,3 +30,16 @@ Facter.add(:root_home) do
     hash['dir'].strip
   end
 end
+
+Facter.add(:root_home) do
+  confine :kernel => :aix
+  root_home = nil
+  setcode do
+    str = Facter::Util::Resolution.exec("lsuser -C -a home root")
+    str && str.split("\n").each do |line|
+      next if line =~ /^#/
+      root_home = line.split(/:/)[1]
+    end
+    root_home
+  end
+end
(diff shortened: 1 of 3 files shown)


lib/facter/root_home.rb
```diff
@@ -30,3 +30,16 @@ def get_root_home
...
   end
 end
+
+Facter.add(:root_home) do
+  confine :kernel => :aix
+  root_home = nil
+  setcode do
+    str = Facter::Util::Resolution.exec("lsuser -C -a home root")
+    str && str.split("\n").each do |line|
+      next if line =~ /#/
```
DavidS: Given your example, this should probably be matching `/^#/` ? Not that I expect any sane person to have a hash in its root directory ...

DavidS:
Hi Jon, thanks for your contribution! If you could squash that into a single commit, it would be grand to keep the git history clean.

Cheers, David

See https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History#Squashing-Commits

lib/facter/root_home.rb
jfautley: Good spot, done, thanks.

jfautley:
Commits squashed into one, and comment-line match fixed.

DavidS:

PR #516, opened 2015-08-28
Adds an is_a() function

binford2k:
The data type system is very hard to understand. Many people don't understand why

    type_of([1,2,3]) == Array

will fail, but

    type_of([1,2,3]) <= Array

passes. This does a simpler validation that doesn't rely on explicit data types. Instead, use

    $foo = [1,2,3]
    if $foo.is_a(Array) {
      notify { 'This is an array': }
    }

### DIFF
--- a/lib/puppet/functions/is_a.rb
+++ b/lib/puppet/functions/is_a.rb
@@ -1,0 +1,26 @@
+# Boolean check to determine whether a variable is of a given data type.
+#
+# @example how to check a data type
+#   # check a data type
+#       foo = 3
+#       $bar = [1,2,3]
+#       $baz = 'A string!'
+#
+#       if $foo.is_a(Integer) {
+#         notify  { 'foo!': }
+#       }
+#       if $bar.is_a(Array) {
+#         notify { 'bar!': }
+#       }
+#       if $baz.is_a(String) {
+#         notify { 'baz!': }
+#       }
+#
+# See the documentation for "The Puppet Type System" for more information about types.
+# See the `assert_type()` function for flexible ways to assert the type of a value.
+#
+Puppet::Functions.create_function(:is_a) do
+  def is_a(type, value)
+    Puppet::Pops::Types::TypeCalculator.instance?(value,type)
+  end
+end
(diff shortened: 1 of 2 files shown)


hlindberg:
The intent is to use the `=~` operator, not `==`, and there is no need to use `type_of` unless doing more advanced type calculations.  Here is the difference between using `=~` directly on the value, and on the type. There is no need to call a function to get the "matches type" already built into the language.

```
[1,2,3] =~ Array
type_of([1,2,3]) =~ Type[Array]
```

If logic needs to do different things depending on type - I recommend using a case expression - say that different paths should be taken based on if the value is an array or a hash - then do like this:

```
case $value {
  Array : {  }
  Hash : {  }
  default : { }
}
```
The behavior of a case option match is essentially the same as the `=~` operator.

Also note that there is an `assert_type` function in Puppet itself that can be used as an `is_a` - standard behavior is to fail if value is not an instance, but this can be modified to do other things.

If we in documentation recommend anything else than using `=~` as the primary choice (promoting the idea that type checking is the same as pattern matching - analogous to string matching regular expressions) we should address that in documentation. While it is true that comparisons (<, >, ==, etc.) can be made directly on types - that is something I expect to see very rarely in Puppet code.

With all of this in mind, I really don't think an `is_a` function adds any value.

DavidS:
TIL. Thanks, @hlindberg !

binford2k:
With all due respect, while you may not expect to see direct equality comparisons in Puppet code, I expect to see it a ton from people trying to learn Puppet and failing. Using a "regex" comparison rather than equality makes sense to you, but it does not make sense to a large proportion of Puppet users who are in the subset of humans who are not Henrik ;-)

The value this adds is a mnemonic that makes sense to readers of the code and to people learning how the language and data type system works.

binford2k:
Docs ticket filed at https://tickets.puppetlabs.com/browse/DOCUMENT-406

DavidS:
thanks!

ripienaar:
+1 I think the intent of reading code written with ```is_a``` is much clearer than the alternatives.

hlindberg:
In any case (if this makes it in or not), the parameters are in the wrong order - the instance value should be the first parameter so you can write `$x.is_a(String)`. The function should also type check its arguments i.e. (Any, Type).

DavidS:
I've reserved a little bit of time next week to work on this: https://tickets.puppetlabs.com/browse/MODULES-2516

PR #518, opened 2015-09-03
accept any case of boolean strings

logicminds:
* previously the str2bool function did not accept 'TRUE' as a bool
    type. This causes the function to now accept TRUE | FALSE strings
    as a boolean type in order to be converted to a proper boolean.
  * This would also cause Y, N, YES, NO to be accepted as boolean types
    as well.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -623,7 +623,7 @@ Returns a new string where runs of the same character that occur in this set are
 
 #### `str2bool`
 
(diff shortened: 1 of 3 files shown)


DavidS:
Hi Corey, this looks really useful!

Please add a short note to the description of the function in the README.

Cheers, David

logicminds:
Just updated the README and function description.

PR #520, opened 2015-09-10
Add function destringify

msutter:
Loads a fact or other variable, and in case of a stringified fact
containing an
array, string, or hash, it returns the data in the corresponding
native data type.

Puppet 3.x will sometimes convert all fact values to strings
(e.g. "false" instead of false), depending on the stringify_facts
setting
and the installed Facter version.

If you’re writing code that might be used with pre-4.0 versions of
Puppet,
you’ll need to take extra care when dealing with structured facts.

For example:

if the '::processors' fact is stringified as follow:
```ruby
  "{\"count\"=>8, \"speed\"=>\"2.7 GHz\"}"

  $my_destringified_processors_fact = destringify(::processors)
    => {"count"=>8, "speed"=>"2.7 GHz"}
```

This could also be done with following puppet/stdlib functions:
```puppet
  $_fact = $::my_structured_fact
  if is_string($_fact) {
    $_yaml = regsubst($_fact,'=>',': ', 'G')
    $_destringified_fact = parseyaml($_yaml)
  } else {
    $_destringified_fact = $_fact
  }
```

But using this function it will be reduced to:
```puppet
$_destringified_fact = destringify($::my_structured_fact)
```

### DIFF
--- a/lib/puppet/parser/functions/destringify.rb
+++ b/lib/puppet/parser/functions/destringify.rb
@@ -1,0 +1,63 @@
+#
+# destringify.rb
+#
+# Loads a fact or other variable, and in case of a stringified fact containing an
+# array, string, or hash, it returns the data in the corresponding native data type.
+#
+# Puppet 3.x will sometimes convert all fact values to strings
+# (e.g. "false" instead of false), depending on the stringify_facts setting
+# and the installed Facter version.
+#
+# If you’re writing code that might be used with pre-4.0 versions of Puppet,
+# you’ll need to take extra care when dealing with structured facts.
+#
+# For example:
+#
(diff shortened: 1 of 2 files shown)


DavidS:
@msutter, afaict, `Facter.value()` always returns a structure fact. The stringification is done by puppet. So to solve your stated use-case, it'd be much nicer to have a `facter_lookup` function that returns the un-stringified value, instead of making unspeakable parsing things.

DavidS:
Hi Marc, thanks for this pull request. However due to inactivity and a better alternative this pull request is being closed. If you want to re-submit this PR wrapping Facter.value(), we will gladly look at the new pull request.

PR #521, opened 2015-09-10
Adds cache_data function for caching values on the master

ehelms:
Allows retrieving data from a cache file, or creates it with
supplied data if the file doesn't exist. This is useful for having
data that's randomly generated once on the master side (e.g. a password),
but then stays the same on subsequent runs.

### DIFF
--- a/lib/puppet/parser/functions/cache_data.rb
+++ b/lib/puppet/parser/functions/cache_data.rb
@@ -1,0 +1,44 @@
+require 'fileutils'
+require 'yaml'
+require 'etc'
+
+# Retrieves data from a cache file, or creates it with supplied data if the file doesn't exist
+#
+# Useful for having data that's randomly generated once on the master side (e.g. a password), but
+# then stays the same on subsequent runs.
+#
+# Usage: cache_data(name, initial_data)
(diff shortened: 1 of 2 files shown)


ehelms:
Another design option I contemplated was specifying the namespace in the name declaration and parsing it, e.g. `cache_data('mysql/mysql_password', 'my_password')`

nibalizer:
:-1:

Hiera already provides the ability to push constant data into the puppet catalog. A hiera backend could be written or extended to generate passwords on the first lookup. Adding caching to software is not something that should be lightly done. This also fails when multiple masters are being used in a load balancer. If we want to add a generate-and-cache thing to puppet (which I totally see and agree with the utility of) it's a much much bigger discussion that should happen on puppet-dev.

nibalizer:
As usual, anything that is too 'out there' for the stdlib is welcome in https://github.com/puppet-community/puppet-extlib

bmjen:
:+1: @nibalizer 

I agree that this particular function isn't really within the scope of what we want to accomplish with stdlib and might fit extlib more.

ehelms:
Thanks for the feedback, I have moved this to extlib -- https://github.com/puppet-community/puppet-extlib/pull/17

PR #522, opened 2015-09-11
Add "echo" function

dmitryilyin:
* Outputs the content and type of a variable
  into the debug log.


bmjen:
Thanks for the contribution @dmitryilyin. Unfortunately, I don't think this particular function is within the scope of what we'd like to support in the stdlib library. However, I do recommend submitting a PR for this function in https://github.com/puppet-community/puppet-extlib.

PR #525, opened 2015-09-17
(#MODULES-2584) add new function 'parents'

gibbsoft:
I would like to submit a new function to stdlib called 'parents'.

#### `parents`

Returns an array containing the parent directories of a given absolute pathname. Very useful when using the file resource to avoid the Puppet error `Cannot create /xxx/yyy; parent directory /xxx does not exist`. *Type*: rvalue.

Sample usage:
~~~

file {parents($install_dir):
  ensure => directory,
} ->

file {$install_dir:
  ensure => directory,
}

~~~

Condensed form:
~~~

file {[ parents($install_dir), $install_dir ]:
  ensure => directory,
}

~~~

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -541,6 +541,33 @@ Returns the lowest value of all arguments. Requires at least one argument. *Type
 
 Converts a number or a string representation of a number into a true boolean. Zero or anything non-numeric becomes 'false'. Numbers greater than 0 become 'true'. *Type*: rvalue.
 
+#### `parents`
+
+Returns an array containing the parent directories of a given absolute pathname. Very useful when using the file resource to avoid the Puppet error `Cannot create /xxx/yyy; parent directory /xxx does not exist`. *Type*: rvalue.
+
+Sample usage:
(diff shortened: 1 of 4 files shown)


hunner:
Thanks for your contribution. These kind of "mkdir -p" functions show up from time to time, but there are a few issues with functions like this; how does it avoid duplicate resources when two parents() try to manage /etc? How does it account for when someone sets an owner other than root?

I would recommend vendoring the function in one of your own modules that fits your needs as it is hard to generalize them. Also, this can now be done with the Puppet 4 parser and doesn't need a function for it, if you want.

gibbsoft:
As the function merely returns an array or strings, it is not responsible for creating any directories or managing resource conflicts, such implementation is left to the discretion of the user.  The function is quite flexible and can be useful in several ways, for example parents('/this/that/other')[-1] can be used to return the immediate parent of 'other'.

I've not seen any documentation of Puppet 4 parser being able to do this, can you give me an example please?

PR #529, opened 2015-09-22
MODULES-2614 Check for numeric values as empty fails on those

roman-mueller:


HelenCampbell:
Closed due to reasons mentioned in the following ticket: https://tickets.puppetlabs.com/browse/MODULES-2614

PR #538, opened 2015-10-15
add functionality to bool2str function

mmckinst:
The bool2str function already exists and converts a boolean true/false to a string true/false. It is not documented in the README.

There are cases when something expects a boolean value to be something other than true/false, such as yes/no, t/f, Yes/No, on/off, etc.

This pull request add an optional two element array that bool2str will accept and convert the boolean true/false to the corresponding value in the array. If no array is given, it converts boolean true/false to a string true/false (which also preserves backwards compatibility).

Examples:

```
bool2str(true)                   => 'true'
bool2str(true, ['yes', 'no'])    => 'yes'
bool2str(false, ['t', 'f'])      => 'f' 
```

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -170,6 +170,22 @@ Converts a boolean to a number. Converts values:
   * 'true', 't', '1', 'y', and 'yes' to 1.
   Requires a single boolean or string as an input. *Type*: rvalue.
 
+#### `bool2str`
+
+Converts a boolean to a string using optionally supplied arguments. The optional
+second and third arguments represent what true and false will be converted to
+respectively. If only one argument is given, it will be converted from a boolean
+to a string containing 'true' or 'false'.
+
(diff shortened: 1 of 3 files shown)


DavidS:
bool2str is already of questionable value, given that it basically implements `$val ? { true => 'true', false => 'false', default => fail("need a bool for val, not '$val'") }`.

That said, I would still merge this as it has good docs and tests, if you change the array to two separate parameters. That's another two keystrokes saved.

mmckinst:
@DavidS I concur on the usefulness of the old bool2str. Hopefully this will make it more useful because there's been plenty of times where I've done something like ```$val ? 'On' : 'Off'``` in a template. But now I can convert values right in a manifest and use augeas or file_line to modify files too.

I've updated and squashed my code so it doesn't use an array any more.

PR #541, opened 2015-10-27
Fix capitalize docs

mattflaschen:
Capitalize lower-cases the remaining characters (due to the Ruby function having this behavior); document this, and make minor wording tweaks.


DavidS:
thanks!

PR #542, opened 2015-11-03
use properly encoded characters

greg0ire:
This is more severe than it sounds. These characters make puppet fail with the following message : 

> Error 400 on SERVER: "\xC3" on US-ASCII

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -349,7 +349,7 @@ This is a supported release
 
  * Add an ensure\_packages function. (8a8c09e)
 
-##### 2012-11-23 - Erik DalÃƒÂ©n <dalen@spotify.com> - 3.2.0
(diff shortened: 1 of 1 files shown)


HelenCampbell:
Thank you :)

igalic:
really?

if i put my own name (Galić) in .pp files it fails with the same reason.
particularly in apply mode :(

but then again, é is in US-ASCII, ć is not…

greg0ire:
Yeah, I don't know why US-ASCII is being use here rather than utf-8…

igalic:
bizarrely, it's the ruby default, starting 1.9, unless you tell it to use unicode…

greg0ire:
> you tell it to use unicode…

Guess this is what actually needs to be fixed…

PR #545, opened 2015-11-18
Add clamp function

mpolenchuk:
Clamp keeps value within the range.
Employ of soft() makes the whole thing is independant of order.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -202,6 +202,14 @@ Removes the record separator from the end of a string or an array of strings; fo
 
 Returns a new string with the last character removed. If the string ends with '\r\n', both characters are removed. Applying `chop` to an empty string returns an empty string. If you want to merely remove record separators, then you should use the `chomp` function. Requires a string or an array of strings as input. *Type*: rvalue.
 
+#### `clamp`
+
+Keeps value within the range [Min, X, Max] by sort based on integer value (order of params doesn't matter). Takes strings, arrays or numerics. Strings are converted and compared numerically. Arrays of values are flattened into a list for further handling. For example:
+  * `clamp('24', [575, 187])` returns 187.
+  * `clamp(16, 88, 661)` returns 88.
+  * `clamp([4, 3, '99'])` returns 4.
+  *Type*: rvalue.
+
 #### `concat`
 
 Appends the contents of multiple arrays onto the first array given. For example:
--- a/lib/puppet/parser/functions/clamp.rb
+++ b/lib/puppet/parser/functions/clamp.rb
@@ -1,0 +1,30 @@
+#
+# clamp.rb
+#
+
+module Puppet::Parser::Functions
+  newfunction(:clamp, :type => :rvalue, :arity => -2, :doc => <<-EOS
+    Clamps value to a range.
+    EOS
+  ) do |args|
+
+    args.flatten!
+
+    raise(Puppet::ParseError, 'clamp(): Wrong number of arguments, ' +
+          'need three to clamp') if args.size != 3
+
+    # check values out
+    args.each do |value|
+      case [value.class]
+        when [String]
+          raise(Puppet::ParseError, "clamp(): Required explicit numeric (#{value}:String)") unless value =~ /^\d+$/
+        when [Hash]
+          raise(Puppet::ParseError, "clamp(): The Hash type is not allowed (#{value})")
+      end
+    end
+
+    # convert to numeric each element
+    # then sort them and get a middle value
+    args.map{ |n| n.to_i }.sort[1]
+  end
+end
--- a/spec/functions/clamp_spec.rb
+++ b/spec/functions/clamp_spec.rb
@@ -1,0 +1,16 @@
+require 'spec_helper'
+
+describe 'clamp' do
+  it { is_expected.not_to eq(nil) }
+  it { is_expected.to run.with_params().and_raise_error(ArgumentError) }
+  it { is_expected.to run.with_params([]).and_raise_error(Puppet::ParseError) }
+  it { is_expected.to run.with_params(12, 88, 71, 190).and_raise_error(Puppet::ParseError, /Wrong number of arguments, need three to clamp/) }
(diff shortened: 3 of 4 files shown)


tphoney:
This would also need a readme update. Thanks for the work.

mpolenchuk:
Done

tphoney:
Thanks for the quick response :+1: 

So reading your implementations of clamp i would prefer the parameter ordering of: value, min, max
rather than min, value, max.

I think the function should be limited to 3 values, and not allowing arrays.

mpolenchuk:
Actually the order of params doesn't matter, it could be [max, min, value] or [max, value, min] or ... whatever.
Function technically limited to 3, but it provides flexible input: get arrays, strings as numerics and numerics itself. This was done intentionally.

mpolenchuk:
In other words you might have $range = [minLimit, maxLimit] and valueX.
Just call clamp($range, valueX) and get value within the range.

mpolenchuk:
*Bump*

lib/puppet/parser/functions/clamp.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    result = []
+
+    args.each do |x|
+      if x.is_a?(Array)
+        result += x
+      else
+        result << x
+      end
+    end
```
DavidS: couldn't this just be `args = args.flatten` ?

lib/puppet/parser/functions/clamp.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    raise(Puppet::ParseError, "clamp(): Wrong number/type of arguments " +
+          "need three to clamp") if result.size != 3
+
+    result.map{ |n| n.to_i }.sort[1]
```
DavidS: Please use a min/max formulation here to keep this understandable. Even as a seasoned dev, this took me some time to figure out.

lib/puppet/parser/functions/clamp.rb
```diff
@@ -0,0 +1,31 @@
...
+  end
+end
+
+# vim: set ts=2 sw=2 et :
```
DavidS: please remove this.

spec/functions/clamp_spec.rb
```diff
@@ -0,0 +1,13 @@
...
+describe 'clamp' do
+  it { is_expected.not_to eq(nil) }
+  it { is_expected.to run.with_params().and_raise_error(Puppet::ParseError) }
+  it { is_expected.to run.with_params("string1").and_raise_error(Puppet::ParseError) }
+  it { is_expected.to run.with_params([]).and_raise_error(Puppet::ParseError) }
+  it { is_expected.to run.with_params([4,3,99]).and_return(4) }
+  it { is_expected.to run.with_params(444,3,99).and_return(99) }
+  it { is_expected.to run.with_params([4,873],73).and_return(73) }
+  it { is_expected.to run.with_params([4],8,99).and_return(8) }
+  it { is_expected.to run.with_params([4],[31], 9911).and_return(31) }
```
DavidS: These need to test all allowed inputs and input combinations. Specifically at least all the tests with (mixed) strings, error handling in the case of non-numeric strings, and error handling with more than three elements.

lib/puppet/parser/functions/clamp.rb
```diff
@@ -0,0 +1,31 @@
...
+#
+
+module Puppet::Parser::Functions
+  newfunction(:clamp, :type => :rvalue, :doc => <<-EOS
+    Clamps value to a range.
+    EOS
+  ) do |args|
+
+    raise(Puppet::ParseError, "clamp(): Wrong number of arguments " +
+          "need at least one") if args.size < 1
```
DavidS: More than three elements don't make sense, and should be forbidden.

README.markdown
```diff
@@ -202,6 +202,10 @@ Removes the record separator from the end of a string or an array of strings; fo
 
 Returns a new string with the last character removed. If the string ends with '\r\n', both characters are removed. Applying `chop` to an empty string returns an empty string. If you want to merely remove record separators, then you should use the `chomp` function. Requires a string or an array of strings as input. *Type*: rvalue.
 
+#### `clamp`
+
+Keeps value within the range [Min, X, Max] by sort based on integer value. Takes strings, arrays or numerics.  *Type*: rvalue.
```
DavidS: Please explain how strings and arrays are handled. Specifically whether strings are compared numerically or lexically (which locale?) and how arrays figure into it.

DavidS:
Hi Michael, I added a few comments on issues with the code. I think the overall idea of the clamp function is great, but the code does need a bit work. Especially with the move of puppet 4 to more conscious and strict typing, please reconsider your plan to allow arbitrary arrays and strings as arguments. I would much prefer a more strict signature like tp suggested, that does just take three ints.

lib/puppet/parser/functions/clamp.rb
mpolenchuk: yes, thanx

lib/puppet/parser/functions/clamp.rb
mpolenchuk: Using of "sort" was a main goal of that function otherwise we could use mix/max now.
I believe sort() is so intuitive/native as far as it might be (get middle value by sort[1]).

lib/puppet/parser/functions/clamp.rb
mpolenchuk: done

spec/functions/clamp_spec.rb
mpolenchuk: done

lib/puppet/parser/functions/clamp.rb
mpolenchuk: done

README.markdown
mpolenchuk: done

mpolenchuk:
Hi David, thank you for the feedback. I've fixed issues w/ the code.

DavidS:
Thanks for your work, @mpolenchuk !

PR #549, opened 2015-11-20
Fix package_provider for puppet enterprise

snandam:
Facter/puppet run throws the below error
```sh
Could not retrieve fact='package_provider', resolution='<anonymous>': Malformed version number string 3.8.0 (Puppet Enterprise 3.8.0)
```

To reproduce the error:
```sh
facter -p
```

Version:
```sh
puppetversion => 3.8.0 (Puppet Enterprise 3.8.0)
facterversion => 2.4.3
```


lib/facter/package_provider.rb
```diff
@@ -12,10 +12,20 @@
 
 Facter.add(:package_provider) do
   setcode do
-    if Gem::Version.new(Facter.value(:puppetversion)) >= Gem::Version.new('3.6')
-      Puppet::Type.type(:package).newpackage(:name => 'dummy', :allow_virtual => 'true')[:provider].to_s
+    # if puppet enterprise is install use pe_version instead of puppetversion
+    if Facter.value(:pe_version) != nil
```
elyscape: This could all be simplified to:
```ruby
puppetversion = Facter.value(:pe_version) || Facter.value(:puppetversion)
if Gem::Version.new(puppetversion) > Gem::Version.new('3.6')
  Puppet::Type.type(:package).newpackage(:name => 'dummy', :allow_virtual => 'true')[:provider].to_s
else
  Puppet::Type.type(:package).newpackage(:name => 'dummy')[:provider].to_s
end
```

elyscape:
Also, please add a spec test for this.

snandam:
Thank you so much for recommendations. Will do the same.

snandam:
I couldn't find an existing [test](https://github.com/puppetlabs/puppetlabs-stdlib/blob/master/spec/unit/facter/package_provider_spec.rb) to particular variable usage. 

Sorry don't have experience writing rspec tests but learning to do the same :) Excuse me for the delay. If anyone has any examples I can look at will be greatly appreciated.

snandam:
Will create another pull request with tests.

PR #551, opened 2015-11-23
Fix reference to validate_bool in IP4 function

mattbostock:
The documentation in `validate_ipv4_address` references `validate_bool`,
but I believe this should read `validate_ipv4_address` instead, which
makes more sense.


DavidS:
Thanks!

PR #553, opened 2015-12-08
adds new parser called is_absolute_path

logicminds:
I really like the validate_absolute_path function and needed similar functionality to perform conditional logic on installer locations so I created the is_absolute_path function.

This uses the same code as the validate function but only accepts one argument and also returns a value.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -474,6 +474,10 @@ Boolean check to determine whether a variable is of a given data type. This is e
 See the [the Puppet type system](https://docs.puppetlabs.com/references/latest/type.html#about-resource-types) for more information about types.
 See the [`assert_type()`](https://docs.puppetlabs.com/references/latest/function.html#asserttype) function for flexible ways to assert the type of a value.
 
+#### `is_absolute_path`
+
+Returns 'true' if the given path is absolute. *Type*: rvalue.
+
 #### `is_array`
 
 Returns 'true' if the variable passed to this function is an array. *Type*: rvalue.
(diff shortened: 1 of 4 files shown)


igalic:
@logicminds rather than duplicating the code, i'd now switch validate_absolute_path to calling this function.

logicminds:
@igalic I gave that some serious thought but I just assumed the performance would be better without having to call a puppet function from within a puppet function multiple times.  This might be negligible if using once but it folks are using this 50 times across their entire catalog, a 100 ms delay adds up pretty quick.  I have no benchmarks to back up my claim though.

igalic:
so, we have a [loop](https://github.com/puppetlabs/puppetlabs-stdlib/blob/master/lib/puppet/parser/functions/validate_absolute_path.rb#L44-L67)

```ruby
candidates.each do |path|
  function_is_absolute_path([path]) or raise Puppet::ParseError, ("#{path.inspect} is not an absolute path.")
end
```

now, a compiler that cannot figure out that this is a hot-spot, and needs optimization probably isn't worth its money, even if it's free.

imo, it's probably cheaper to keep the code in one parser function, than wonder why down the road the two of them are behaving differently.

however: i'm not an stdlib maintainer, so lets wait what @puppetlabs has to say!

logicminds:
Well that is definitely much cleaner.

logicminds:
This PR is only adding a function while most of this conversation is about altering an existing function to rely on a yet to be included function. We need to have a new PR for altering the validation_absolute_path and keep discussing there instead of here.

bmjen:
@logicminds I agree with @igalic. For maintainability purposes it would be best to use DRY principles.

Also, I'd be a proponent of coupling the refactoring of `validate_absolute_path` in this PR if you have the bandwidth to do it. This way the conversation can be cohesive in the why and the when.

logicminds:
ok I'll refactor soon and rebase.

logicminds:
@bmjen  this is ready to go.

bmjen:
@logicminds Thanks!

PR #554, opened 2015-12-08
(#2886) seeded_rand: new function

kjetilho:
seeded_rand is needed for repeatable randomness across nodes in a cluster


bmjen:
Thanks for the contribution @kjetilho !

PR #562, opened 2015-12-18
Add a simple fact to return PE roles

binford2k:
This just returns a list of the toplevel PE profiles that the node is
classified with. Clearly, this will only take effect on the second run,
but I don't think there is a cleaner/better way.

### DIFF
--- a/lib/facter/pe_roles.rb
+++ b/lib/facter/pe_roles.rb
@@ -1,0 +1,13 @@
+# Simply return an array of all PE profiles this node is classified with.
+# Clearly subject to a chicken and egg, but there's probably not a better way.
+Facter.add(:pe_roles) do
+  setcode do
+    begin
+      classes = File.readlines(Puppet.settings[:classfile])
+      classes = classes.grep(/puppet_enterprise::profile::([^:]*)$/) { |cl| $1.chomp }
+      classes.uniq
+    rescue
+      nil
+    end
+  end
+end


DavidS:
If you have the possibility to influence the shape of the configuration, I'd turn the roles into an array that is configured, instead of including classes. Then this wouldn't be needed at all.

Isn't there some place in PE where we can shove this?

Please squash the fixup commit.

jonnytpuppet:
@bmjen I'd appreciate your opinion on this. Do we really want a feature in stdlib that only works on the second time round? Is there prior-art for this?

Thanks

binford2k:
@ericwilliamson has some input as to plans for the puppet_enterprise module that might obsolete this PR.

bmjen:
@jonnytpuppet I don't recall any function/feature we've merged that required two executions to yield the correct response. Based on recent comments, this PR might not be needed any longer? @DavidS 's feedback needs to be addressed as well, about finding a better place to do this than in stdlib.

bmjen:
@binford2k is there any news on the future of this PR? Whether the work will be migrated to the puppet_enterprise module?

ericwilliamson:
@binford2k @bmjen Any work on this in the puppet_enterprise module probably won't happen for a couple releases.

The  suggested approach is the best I can think of for now until we get around to using trusted facts.

bmjen:
Hi @binford2k and @ericwilliamson, Is possible for you to find an alternative way to handle this? We discussed this in modules triage and don't feel that this function should belong in stdlib, especially in it's current form requiring 2 runs to get the expected output. Is there an urgency or other requirement that prevents this from being implemented on the enterprise/integration side?

For now I'm going to close this PR. But if there are any objections, please contact myself or @hunner to discuss possible solutions.

PR #563, opened 2015-12-29
Lint

ghoneycutt:


ghoneycutt:
I realize it probably needs a jira ticket and the commit msg reformatted. Wanted to see if you were interested in this first.

DavidS:
Great stuff. I don't need a ticket, but I would prefer to have it added to https://github.com/puppetlabs/puppetlabs_spec_helper and hooked up to https://github.com/puppetlabs/puppetlabs_spec_helper/blob/master/lib/puppetlabs_spec_helper/rake_tasks.rb#L233 so that everyone profits.

ghoneycutt:
@DavidS Why would this be part of puppetlabs_spec_helper? This is to provide extra lint checks, not spec_tests.

DavidS:
puppetlabs_spec_helper has become a catch-all bucket for test/qa related stuff. somewhere on my bucket list is extracting the nice parts of PSH into a more aptly named utility gem and bury the rest in depreciation oblivion.

ghoneycutt:
@DavidS How is the Rakefile from that project used in a module? I'm not sure how changing code there is used/accessed by other modules.

That link to the Rakefile enforces some opinions on which puppet-lint checks to disable that I would not be able to use or recommend. Could you override things like the puppet-lint checks?

DavidS:
https://github.com/puppetlabs/puppetlabs-stdlib/blob/master/Rakefile#L3 is the usage, and you can also see there how to change the config afterwards. The goal is to have a common set of rules that is available everywhere without having to template/c&p it across everywhere.

I'd love to hear your detailed input on the defaults over at PSH.

HelenCampbell:
Hey @ghoneycutt , Thanks for the work in looking in to this, however as David specifies we use puppetlabs_spec_helper for things like this. We would certainly be happy to review any changes like this in a PR against the spec_helper. For now I'll go ahead and close this. Thanks!

PR #567, opened 2016-01-16
Add test for basename on path with scheme

alechenninger:
I didn't see explicit documentation on whether or not this would work, so I added a test to confirm. Test passes, so I assume this is intended behavior. A test makes this behavior explicit.


DavidS:
hah! good catch :-)

PR #568, opened 2016-01-19
Fix reference to validate_bool in function

mattbostock:
The documentation in `validate_ip_address` references `validate_bool`,
but I believe this should read `validate_ip_address` instead, which
makes more sense.

Looks like this was copied from `validate_ipv4_address`, which I fixed
in 7b068781.


jonnytpuppet:
Great catch!

PR #570, opened 2016-01-22
Add is_ipv4_address and is_ipv6_address functions

gfidente:
These are useful when making decisions based on the type of IP
address received.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -518,6 +518,14 @@ Returns 'true' if the variable returned to this string is an integer. *Type*: rv
 
 Returns 'true' if the string passed to this function is a valid IP address. *Type*: rvalue.
 
+#### `is_ipv6_address`
+
+Returns 'true' if the string passed to this function is a valid IPv6 address. *Type*: rvalue.
+
+#### `is_ipv4_address`
+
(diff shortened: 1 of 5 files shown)


EmilienM:
Why not re-using validate_ipv6_address and add an option to not fail the catalog but just return False?
That would reduce the code impact and could be backward compatible if you set the failure to True by default.

HelenCampbell:
Hey @gfidente , thank you for your work! This would be a welcome addition to stdlib. Would you be willing to make the changes suggested by @EmilienM ? I can merge in as-is, however the changes suggested would make for a good improvement.

hunner:
The `validate_*` messages traditionally fail the compile if they are unmatched so creators don't have to hand-write the logic and failure messages over and over, and `is_*` return true/false for actionable items.

This would be improved by updating the validate ipv4/6 functions to call the is_* methods and raise a failure on false return, but not required.

hunner:
Perhaps a `is_ip_address` would be useful too...

gfidente:
hi all, thanks for checking this!

maybe we can add a parameter for IPv6 to is_ip_address and call the same code from all is_ipv4 is_ipv4 and validate_* ?

PR #571, opened 2016-01-27
Add globbing for loadyaml

sspreitzer:
Can someone please review and guide/help me with the rspec?

### DIFF
--- a/lib/puppet/parser/functions/loadyaml.rb
+++ b/lib/puppet/parser/functions/loadyaml.rb
@@ -1,24 +1,29 @@
 module Puppet::Parser::Functions
 
   newfunction(:loadyaml, :type => :rvalue, :doc => <<-'ENDHEREDOC') do |args|
-    Load a YAML file containing an array, string, or hash, and return the data
-    in the corresponding native data type.
+    Load a list of YAML files containing an array, string, or hash, and 
+    return the data in the corresponding native data type.
 
     For example:
 
-        $myhash = loadyaml('/etc/puppet/data/myhash.yaml')
+        $myhash = loadyaml('/etc/puppet/data/*.yaml', '/etc/myconf/*.yml')
     ENDHEREDOC
 
-    unless args.length == 1
-      raise Puppet::ParseError, ("loadyaml(): wrong number of arguments (#{args.length}; must be 1)")
+    unless args.length >= 1
+      raise Puppet::ParseError, ("loadyaml(): wrong number of arguments (#{args.length}; must be >= 1)")
     end
 
-    if File.exists?(args[0]) then
-      YAML.load_file(args[0])
-    else
-      warning("Can't load " + args[0] + ". File does not exist!")
-      nil
+    ret = {}
+    args.each do |arg|
+
+      Dir.glob(arg) do |entry|
+
+        ret = ret.merge( YAML.load_file(entry) )
+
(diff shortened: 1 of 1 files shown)


DavidS:
Hi Sasha,

thanks for your contribution. The tests complain about the function being changed from returning the data in the file to returning the filename(s) in an array, which matches your code changes, which in turn is not what I assume you're intending to.

To address this, first you need to decide how multiple yaml files should be merged, and then add that to the implementation. The existing tests need to continue to pass as is to avoid breaking existing code.

sspreitzer:
Hi DavidS

Thank you for the hint, will follow up on that later!

sspreitzer:
Hi David

I think some of the errors should be fixed now. For the others a change to rspec might be needed as they function design also changes.
* It is now possible to provide a list of paths to be globbed
* The globbing mechanism can not check whether a file is available or not as it will return all available files. If they are not available, they will not be parsed

DavidS:
Looking at https://github.com/puppetlabs/puppetlabs-stdlib/blob/master/spec/functions/loadyaml_spec.rb , you'll need to modify the tests at L10 and L19 to match the implementation, e.g. to `Dir.expects(:glob). ...`

The test expecting multiple arguments to throw an error can go away, since the new function can cope with multiple arguments now.

Since you're expanding the functionality, you should also add test cases to cover this:

* What happens if you pass in a glob that matches multiple files?
* What happens if one of the files contains non-Hash data?
* What happens if there is a directory matching the glob?

sspreitzer:
Hi @DavidS 

Thank you for your help! :-)

> What happens if one of the files contains non-Hash data?

This is an interesting question. Is there a design pattern that tells whether Puppet should Error or can it silently proceed with skipping the wrong file in such case?
Both ways are implementable.

DavidS:
Currently the loadyaml function supports arbitrary data in the file,
including arrays, hashes, strings and numbers. Dropping some data might be
very surprising. What would you think about e.g. putting the result of each
file into an array cell? with the new puppet 4 language features, people
can munge the data as they need themselves.
Thinking about that I had another question: how can you avoid globbing, if
you do not want this behavior, e.g. in security-critical scenarios, or when
your file name just has a asterisk or a question mark in it? Together with
the question above, it might be much easier to implement a separate glob()
function, that just expands the filenames and allow people to use that as
an input to the unmodified loadyaml function. Using puppet 4 language would
make that quite possible and glob() can be used with every function that
requires a filename (like generate() or template()).

sspreitzer:
PS.: I am still working on this..

HelenCampbell:
Hey @sspreitzer , thank you for your contribution! Just following up on this as it's been a while now since this PR has been active. Is there anything we can help you with further? Our usual policy is to close inactive PRs after 30 days or so, but I don't want to close this incase you are still actively working on it.

sspreitzer:
hi helen
absolutely, i will follow up on this this weekend. sry for the delay. got million other things running

sspreitzer:
okay, forget about this PR, I will create a seperate glob function and later on patch loadyaml to support its inpute. thanks a lot.

HelenCampbell:
@sspreitzer Thanks for the update on this, feel free to give @DavidS or I a ping whenever you need any help or review on your new PR. Again, thank you for all the effort :)

PR #572, opened 2016-02-06
Add clientcert_rand

hashbrowncipher:
The fqdn is not guaranteed to stay stable across Puppet runs. By contrast, the client certificate is likely to stay the same. This is useful for when Puppet is responsible for determining hostnames.

### DIFF
--- a/lib/puppet/parser/functions/clientcert_rand.rb
+++ b/lib/puppet/parser/functions/clientcert_rand.rb
@@ -1,0 +1,19 @@
+require 'digest/md5'
+
+Puppet::Parser::Functions::newfunction(:clientcert_rand, :arity => -2, :type => :rvalue, :doc =>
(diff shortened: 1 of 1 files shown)


bmjen:
Thanks @hashbrowncipher! Can you write up a new section in the docs for the new function? Also do you think you can add some tests?

HelenCampbell:
Hey @hashbrowncipher , thanks for your work so far on this :) though is there any way you would be able to do the updates that @bmjen has requested? We aim to keep the documentation and tests as updated as possible, especially when it comes to adding new functionality.

hashbrowncipher:
Yep, gotcha. It's on my to-do list, but below a number of other things.

tphoney:
Hi @hashbrowncipher thanks for the work you have put in so far. If you want to resubmit this PR with the suggested changes from @bmjen. We will gladly review it. I am closing this PR.

PR #573, opened 2016-02-09
Add dig function

keymone:
Makes deep lookup in nested hashes(and/or arrays) easier.

Usage:

    dig($hash, ["key1", "key2", ...])
    dig($hash, ["key1", "key2", ...], "default")
    dig(
      [{"a" => {"b" => "c"}}],
      [0, "a", "b"]) # => "c"

Returns `nil` or optional `default` arg if at any point in the chain key couldn't be looked up.

### DIFF
--- a/lib/puppet/parser/functions/dig.rb
+++ b/lib/puppet/parser/functions/dig.rb
@@ -1,0 +1,54 @@
+#
+# dig.rb
+#
+
+module Puppet::Parser::Functions
+  newfunction(:dig, :type => :rvalue, :doc => <<-EOS
+Looks up into a complex structure of arrays and hashes and returns nil
+or the default value if nothing was found.
+
+Path is an array of keys to be looked up in data argument. The function
+will go down the structure and try to extract the required value.
+
+$data = {
+  'a' => {
+    'b' => [
+      'b1',
+      'b2',
+      'b3' ]}}
+
+$value = dig($data, ['a', 'b', '2'], 'not_found')
+=> $value = 'b3'
+
+a -> first hash key
+b -> second hash key
+2 -> array index starting with 0
+
+not_found -> (optional) will be returned if there is no value or the path
+did not match. Defaults to nil.
+
+In addition to the required "path" argument, "dig" accepts default
+argument. It will be returned if no value was found or a path component is
+missing. And the fourth argument can set a variable path separator.
+    EOS
+             ) do |arguments|
+    # Two arguments are required
+    raise(Puppet::ParseError, "dig(): Wrong number of arguments " +
+                              "given (#{arguments.size} for at least 2)") if arguments.size < 2
+
+    data, path, default = *arguments
+
+    if !(data.is_a?(Hash) || data.is_a?(Array))
+      raise(Puppet::ParseError, "dig(): first argument must be a hash or an array, " <<
+                                "given #{data.class.name}")
+    end
+
+    unless path.is_a? Array
+      raise(Puppet::ParseError, "dig(): second argument must be an array, " <<
+                                "given #{path.class.name}")
+    end
+
+    value = path.reduce(data) { |h, k| (h.is_a?(Hash) || h.is_a?(Array)) ? h[k] : break }
+    value.nil? ? default : value
+  end
+end
--- a/README.markdown
+++ b/README.markdown
@@ -254,6 +254,42 @@ Deletes all instances of the undef value from an array or hash. For example, `$h
 
 Returns the difference between two arrays. The returned array is a copy of the original array, removing any items that also appear in the second array. For example, `difference(["a","b","c"],["b","c","d"])` returns ["a"]. *Type*: rvalue.
 
+#### `dig`
+
+*Type*: rvalue.
+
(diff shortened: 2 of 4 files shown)


elyscape:
This already exists as [`try_get_value`](https://github.com/puppetlabs/puppetlabs-stdlib#try_get_value).

elyscape:
That being said, it might make sense to extend `try_get_value` to support taking an array as its second argument so that no string splitting needs to be performed.

keymone:
with `#dig` making it's way into standard ruby (http://ruby-doc.org/core-2.3.0/Hash.html#method-i-dig) i think it makes sense to preserve the name. i also like implementation and api more, but i'm subjective here obviously.

keymone:
how would you feel about me reimplementing `#try_get_value` in terms of `#dig` function, and marking it deprecated?

elyscape:
Might be a good idea. Pinging @DavidS and @hunner for input.

keymone:
i've dropped some test cases from try_get_value but it remains soft-backwards-compatible, meaning it will produce an error instead of unexpected value in some cases. i don't think it's good idea to tolerate lookup functions being called on values that are not indexable.

lib/puppet/parser/functions/dig.rb
```diff
@@ -0,0 +1,56 @@
...
+will go down the structure and try to extract the required value.
+
+$data = {
+  'a' => {
+    'b' => [
+      'b1',
+      'b2',
+      'b3' ]}}
+
+$value = try_get_value($data, ['a', 'b', '2'], 'not_found')
```
bmjen: should be `dig`

lib/puppet/parser/functions/dig.rb
```diff
@@ -0,0 +1,56 @@
...
+      raise(Puppet::ParseError, "dig(): second argument must be an array, " <<
+                                "given #{path.class.name}")
+    end
+
+    value = path.reduce(data) { |h, k| (h.is_a?(Hash) || h.is_a?(Array)) ? h[k] : break }
+    value.nil? ? default : value
+  end
+end
+
+# vim: set ts=2 sw=2 et :
```
tphoney: I dont think this should be here

spec/functions/try_get_value_spec.rb
```diff
@@ -27,22 +27,6 @@
     it 'should exist' do
       is_expected.not_to eq(nil)
     end
-
-    it 'should be able to return a single value' do
-      is_expected.to run.with_params('test').and_return('test')
-    end
```
hunner: This test should actually stay, to check for backwards compatibility.

spec/functions/try_get_value_spec.rb
```diff
@@ -27,22 +27,6 @@
...
-
-    it 'should use the default value if data is a single value and path is present' do
-      is_expected.to run.with_params('test', 'path', 'default').and_return('default')
-    end
-
-    it 'should return default if there is no data' do
-      is_expected.to run.with_params(nil, nil, 'default').and_return('default')
-    end
```
hunner: This one too?

tphoney:
Thanks for the work. Could you squash the commits as well.

spec/functions/try_get_value_spec.rb
keymone: the purpose of this function is to do a lookup on nested data structure. passing string where nested data structure is expected is just wrong and dig() will fail explicitly in this case.

spec/functions/try_get_value_spec.rb
keymone: same thing: https://github.com/puppetlabs/puppetlabs-stdlib/pull/573#discussion_r52651183

keymone:
@tphoney do you mean squash and push --force into same branch?

HelenCampbell:
@keymone Yep! I usually do something like: http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html , then push to the same branch to update this PR with your changes. Also, would you be able to add some info in the readme for your new function? Thanks!

keymone:
@HelenCampbell @tphoney done

tphoney:
The test cases that @hunner marked to be kept are important. The try_get_value function should still be able to function with a single parameter, otherwise it might affect how current users are using try_get_value.

HelenCampbell:
OR another possible solution to that would be to keep the try_get_value function alone, and add dig as an additional function. The worry is that the implementation in this PR may break some existing setups.

keymone:
@tphoney try_get_value promotes bad coding practices by being tolerant to bad input data. if somebody's codebase depends on bad coding practices, trying to be backward compatible with that will only make everything worse.

@HelenCampbell i could revert my changes to try_get_value, but breeding duplicate functions with subtly different functionality is a recipe for disaster. i'd love to see stdlib as clean and strict as possible.

keymone:
@hunner i've reverted spec changes and preserved original `try_get_value` functionality

cc/ @tphoney @HelenCampbell

hunner:
@keymone Thanks! I agree that bad coding practices should not be promoted, but hopefully the deprecation will get people to migrate to dig and correct their practices.

PR #579, opened 2016-03-17
Add check if Gem is defined

sulaweyo:
On e.g. Ubuntu 12.04 LTS Gem is not there by default so i added a check to not fail in that fact if this is the case.

### DIFF
--- a/lib/facter/package_provider.rb
+++ b/lib/facter/package_provider.rb
@@ -12,7 +12,7 @@ require 'puppet/type/package'
 
 Facter.add(:package_provider) do
   setcode do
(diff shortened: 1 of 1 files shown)


bmjen:
@sulaweyo What version of rubygems and ruby are you running? It sounds weird that any installation of puppet (far enough to run facter) on any OS would not have Gem installed and defined.

sulaweyo:
It popped up in our CI. It was some Ubuntu 12.04 LTS which did not have all current updates in (for a bunch of reasons). But still i'd say to check if Gem ist there is never wrong to do.

bmjen:
@sulaweyo Sounds good. I agree that it's not a significant change to add a check. Thanks!

PR #581, opened 2016-03-22
improve suffix function to support the same feature set as prefix

vicinus:
At the moment the function prefix allows a hash as input, but the function suffix only allows arrays. This patch adds the same feature set to suffix as the function prefix already has.


bmjen:
Awesome work @vicinus ! Thanks!

PR #582, opened 2016-03-24
MODULES-3201 - Fixed typo 'absense' to 'absence'

derekmceachern:
Believe that the 'match_for_absense' should actually be 'match_for_absence'


igalic:
it appears that the latest release of puppet 4 in strict mode issues slightly different errors than the previous release… this has nothing to do with your changes.

PR #585, opened 2016-03-28
[MODULES-2370] file_line.rb: Fix `line` attribute validation

jearls:
`file_line` type: During validation, do not require `line` attribute if:
* `ensure` is `absent`,
* `match` is not empty,
* and `match_for_absence` is `true`.

Also update `spec` tests to reflect this.

### DIFF
--- a/lib/puppet/type/file_line.rb
+++ b/lib/puppet/type/file_line.rb
@@ -111,8 +111,13 @@ Puppet::Type.newtype(:file_line) do
   end
 
   validate do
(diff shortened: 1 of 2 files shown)


aleksey-novikov:
@jearls Looks good! :+1:

jimr6007:
even with the "line" param specified and using stdlib version 4.11, should this functionality be working? Meaning, the ability to remove a line from a file if the "match" param matches any line in the file?

PR #586, opened 2016-03-29
catch StandardError rather than the gratuitous Exception

ffrank:
Catching `Exception` is gratuitous and makes Ruby behave weirdly in case of badly timed SIGINTs, SIGTERMs etc.

All proper errors should descend from `StandardError` rather than `Exception` directly (`Puppet::Error` does this). Catching those is much less problematic.

### DIFF
--- a/lib/facter/facter_dot_d.rb
+++ b/lib/facter/facter_dot_d.rb
@@ -48,7 +48,7 @@ class Facter::Util::DotD
         end
       end
     end
-  rescue Exception => e
+  rescue StandardError => e
     Facter.warn("Failed to handle #{file} as text facts: #{e.class}: #{e}")
   end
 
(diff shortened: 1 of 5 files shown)


ripienaar:
:+1:

bmjen:
Good catch @ffrank !

dmitryilyin:
https://github.com/puppetlabs/puppetlabs-stdlib/pull/588

DavidS:
The casualty of this change was `parseyaml` , because on ruby 1.9.3, `Psych::SyntaxError` is a `RuntimeError`  :angry: 

See #589 for a fix.

ffrank:
Oh wow, so it raises as though someone tried to load invalid Ruby code at runtime. Not sure that's a sound design. Cool workaround though!

PR #591, opened 2016-04-12
Add support for regular expressions to delete

jyaworski:

PR #592, opened 2016-04-12
Use reject instead of delete_if

jyaworski:
Ping @hunner.

@smoeding alerted me to this problem.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -263,7 +263,11 @@ Takes a resource reference and an optional hash of attributes. Returns 'true' if
 
 #### `delete`
 
-Deletes all instances of a given element from an array, substring from a string, or key from a hash. For example, `delete(['a','b','c','b'], 'b')` returns ['a','c']; `delete('abracadabra', 'bra')` returns 'acada'. `delete({'a' => 1,'b' => 2,'c' => 3},['b','c'])` returns {'a'=> 1}. *Type*: rvalue.
+Deletes all instances of a given element from an array, substring from a string, or key from a hash. Arrays and hashes may also match on regular expressions by providing a full regular expression.
+
+For example, `delete(['a','b','c','b'], 'b')` returns ['a','c']; `delete('abracadabra', 'bra')` returns 'acada'. `delete({'a' => 1,'b' => 2,'c' => 3},['b','c'])` returns {'a'=> 1}, `delete(['abf', 'ab', 'ac'], '^ab.*')` returns ['ac']. `delete(['ab', 'b'], 'b')` returns ['ab'].
+
+*Type*: rvalue.
 
 #### `delete_at`
 
--- a/lib/puppet/parser/functions/delete.rb
+++ b/lib/puppet/parser/functions/delete.rb
@@ -2,8 +2,6 @@
 # delete.rb
 #
 
-# TODO(Krzysztof Wilczynski): We need to add support for regular expression ...
-
 module Puppet::Parser::Functions
   newfunction(:delete, :type => :rvalue, :doc => <<-EOS
 Deletes all instances of a given element from an array, substring from a
@@ -22,19 +20,23 @@ string, or key from a hash.
 
     delete('abracadabra', 'bra')
     Would return: 'acada'
+
+    delete(['abracadabra'], '^.*bra.*$')
+    Would return: []
+
+    delete(['abracadabra'], '^.*jimbob.*$')
+    Would return: ['abracadabra']
   EOS
   ) do |arguments|
 
-    if (arguments.size != 2) then
-      raise(Puppet::ParseError, "delete(): Wrong number of arguments "+
-                                  "given #{arguments.size} for 2.")
-    end
+    raise(Puppet::ParseError, "delete(): Wrong number of arguments "+
+                              "given #{arguments.size} for 2") unless arguments.size == 2
 
     collection = arguments[0].dup
     Array(arguments[1]).each do |item|
       case collection
         when Array, Hash
-          collection.delete item
+          collection.reject! { |coll_item| (coll_item =~ %r{\b#{item}\b}) }
         when String
           collection.gsub! item, ''
         else
--- a/spec/functions/delete_spec.rb
+++ b/spec/functions/delete_spec.rb
@@ -4,6 +4,7 @@ describe 'delete' do
   it { is_expected.not_to eq(nil) }
   it { is_expected.to run.with_params().and_raise_error(Puppet::ParseError) }
   it { is_expected.to run.with_params([]).and_raise_error(Puppet::ParseError) }
+  it { is_expected.to run.with_params([], 'two') }
   it { is_expected.to run.with_params([], 'two', 'three').and_raise_error(Puppet::ParseError) }
   it { is_expected.to run.with_params(1, 'two').and_raise_error(TypeError) }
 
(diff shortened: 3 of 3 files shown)


smoeding:
I think this still includes a change that would break existing functionality.

stdlib 4.11:
```
delete([ 'abracadabra' ], 'bra') => [ 'abracadabra' ]
delete([ 'abracadabra' ], 'abr') => [ 'abracadabra' ]
```

With this proposed change:
```
delete([ 'abracadabra' ], 'bra') => [ 'abracadabra' ]
delete([ 'abracadabra' ], 'abr') => [ ]
```

Maybe a new optional parameter would be helpful, so the user could specify if the second parameter should be used literally or as a regular expression. Otherwise all special regex characters would lead to a semantic change anyway. What about this for example:

```
delete([ 'a', 'b', 'c', '.' ], '.')
```

jyaworski:
I'm OK with adding a flag to turn on the regex matching, but I'm really confused why your second case is matching. Do we need to explicitly do something like ^expr$?

Any ideas on what that flag might look like? An entirely different function even?

smoeding:
In your previous commit you added `(coll_item =~ %r{#{item}}) == 0` as condition. The result of the match operator is the position of the match. Matching `abr` in `abracadabra` matches from position 0. So in this case every element that starts with the item is removed.

Using anchors would somehow solve this problem. But it would also force the user to provide a full matching regex. To remove all entries that contain the letter 'a' somewhere would have to be written as `.*a.*` while everywhere else (grep, awk, sed, perl, ruby, ...) a simple `a` would suffice. I think this would be confusing.

So one solution could be an additional method parameter (e.g. `is_regex = false`) that the user has to set to `true` to switch to regex matching. Without this parameter the simple string comparison just as before would be used.

jyaworski:
Ah, OK. I couldn't find docs on how regex matching return codes worked. We want to do whole-word matching. \b, possibly, from this link. http://stackoverflow.com/questions/4879395/ruby-regex-rejecting-whole-words

jyaworski:
OK @smoeding and @hunner: I think I've got this.

I did end up having to accept a third, optional argument. There's no other way (that I saw) to do this reasonably and preserve existing functionality without making a lot of assumptions (like whole word regex matching by default).

spec/functions/delete_spec.rb
```diff
@@ -4,19 +4,22 @@
   it { is_expected.not_to eq(nil) }
   it { is_expected.to run.with_params().and_raise_error(Puppet::ParseError) }
   it { is_expected.to run.with_params([]).and_raise_error(Puppet::ParseError) }
-  it { is_expected.to run.with_params([], 'two', 'three').and_raise_error(Puppet::ParseError) }
+  it { is_expected.to run.with_params([], 'two', 'three') }
```
DavidS: this first block of tests test that the wrong number of arguments raises an error. as you're adding an optional third param, the correct fix here is 

it { is_expected.to run.with_params([], 'two', 'three', 'four').and_raise_error(Puppet::ParseError) }

README.markdown
```diff
@@ -265,7 +265,7 @@ Takes a resource reference and an optional hash of attributes. Returns 'true' if
 
 #### `delete`
 
-Deletes all instances of a given element from an array, substring from a string, or key from a hash. Arrays and hashes may also match on regular expressions. For example, `delete(['a','b','c','b'], 'b')` returns ['a','c']; `delete('abracadabra', 'bra')` returns 'acada'. `delete({'a' => 1,'b' => 2,'c' => 3},['b','c'])` returns {'a'=> 1}, `delete(['abf', 'ab', 'ac'], '^ab.*')` returns ['ac']. *Type*: rvalue.
+Deletes all instances of a given element from an array, substring from a string, or key from a hash. Arrays and hashes may also match on regular expressions by setting an optional third parameter to a true value. For example, `delete(['a','b','c','b'], 'b')` returns ['a','c']; `delete('abracadabra', 'bra')` returns 'acada'. `delete({'a' => 1,'b' => 2,'c' => 3},['b','c'])` returns {'a'=> 1}, `delete(['abf', 'ab', 'ac'], '^ab.*', true)` returns ['ac']. *Type*: rvalue.
```
DavidS: the array example could use the same example data as the additional test, using `['ab', 'b'], 'b'` -> `['ab']` to show the significance of not matching regexes.

DavidS:
Hi @jyaworski , a flag seems to be indeed the only way forward here. Good work overall, only a few minor nits around docs and tests.

jyaworski:
@DavidS fixed. Thanks.

README.markdown
```diff
@@ -265,7 +265,7 @@ Takes a resource reference and an optional hash of attributes. Returns 'true' if
 
 #### `delete`
 
-Deletes all instances of a given element from an array, substring from a string, or key from a hash. Arrays and hashes may also match on regular expressions. For example, `delete(['a','b','c','b'], 'b')` returns ['a','c']; `delete('abracadabra', 'bra')` returns 'acada'. `delete({'a' => 1,'b' => 2,'c' => 3},['b','c'])` returns {'a'=> 1}, `delete(['abf', 'ab', 'ac'], '^ab.*')` returns ['ac']. *Type*: rvalue.
+Deletes all instances of a given element from an array, substring from a string, or key from a hash. Arrays and hashes may also match on regular expressions by setting an optional third parameter to a true value. For example, `delete(['a','b','c','b'], 'b')` returns ['a','c']; `delete('abracadabra', 'bra')` returns 'acada'. `delete({'a' => 1,'b' => 2,'c' => 3},['b','c'])` returns {'a'=> 1}, `delete(['abf', 'ab', 'ac'], '^ab.*', true)` returns ['ac']. `delete(['ab', 'b'], 'b')` returns ['ab']. *Type*: rvalue.
```
nibalizer: This line is really quite long, can it be broken up?

lib/puppet/parser/functions/delete.rb
```diff
@@ -20,19 +20,28 @@ module Puppet::Parser::Functions
 
     delete('abracadabra', 'bra')
     Would return: 'acada'
+
+    delete(['abracadabra'], '^.*bra.*$', true)
```
nibalizer: I don't like passing in a magic 'true' at the end of this function. That's bad ux.

README.markdown
jyaworski: Done.

lib/puppet/parser/functions/delete.rb
jyaworski: What do you recommend?

lib/puppet/parser/functions/delete.rb
```diff
@@ -20,19 +20,28 @@ module Puppet::Parser::Functions
...
     collection = arguments[0].dup
+    use_regex = arguments.size == 3 ? arguments[2] : false
     Array(arguments[1]).each do |item|
       case collection
         when Array, Hash
-          collection.delete_if { |coll_item| coll_item =~ %r{#{item}} }
+          if use_regex
+            collection.reject! { |coll_item| (coll_item =~ %r{#{item}}) } # make no assumptions about regex format
+          else
+            collection.reject! { |coll_item| coll_item == item }
```
bmjen: Hey @jyaworski, we can't merge this change, because right here it changes old behavior and is backwards incompatible. Given the old code if you had `delete("foobarfoo", "bar")` then the result is `foofoo`. In this version, this same code would result in `foobarfoo` since it is doing absolute compares.

DavidS:
We've been discussing this in Triage, and it turned out that this is a backwards incompatible:

    it { is_expected.to run.with_params(['one', 'two', 'twentytwo', 'another two'], 'two').and_return(['one']) }

passes on the current implementation, but fails on yours, so that's not good, either.

jyaworski:
@DavidS @Hunner so it looks like we need another function entirely then. Please revert #591 and I'll change this PR to add a delete_regex function.

DavidS:
Ohhh! so the "other" behaviour is not yet released? That then, of course, doesn't make this a breaking change.  More investigation required.

bmjen:
I don't think you need a delete_regex function, because the regex matching will work fine with this 1-line change: `collection.reject! { |coll_item| (coll_item =~ %r{#{item}}) }` here https://github.com/puppetlabs/puppetlabs-stdlib/pull/592/files#diff-85b753300e047711d62ec6a8a075a7deL35. 

It looks like what you want is a delete function that will match on absolute string matching... based on the diff in functionality. This can be accomplished via a new function.. or the user providing a regex that of `^string$`

lib/puppet/parser/functions/delete.rb
jyaworski: This is only for arrays and hashes. Your example is a string.

spec/functions/delete_spec.rb
jyaworski: Done.

lib/puppet/parser/functions/delete.rb
bmjen: it's the same idea for Arrays and Hashes, I believe. The old behavior will delete the key/value pair based on a fuzzy match of the substring. The updated code now requires an absolute match.

jyaworski: I don't believe it did, but I'm unable to verify at the moment. If you want the fuzzy match functionality I can wrap it in ^.*escapedvar.*$
 
Is that amenable?

jyaworski:
To be clear, this is the functionality I'm trying to do:

```ruby
delete(['lo', 'bond0', 'bond1', 'eth0'], '^bond.*$') => ['lo', 'eth0']
```

Delete all keys/entries in a hash/array matching a regular expression.

jyaworski:
Okay @bmjen and @hunner. This is no longer backwards-incompatible.

bmjen:
This looks good @jyaworski! Thanks for making the updates. Can you rebase this and we can push it through. 👍

jyaworski:
@bmjen done. Thanks.

PR #597, opened 2016-04-18
(MODULES-3271) Ensure that is_email_address works on unsupported rubies

DavidS:


DavidS:
Merged as #598 into the release branch. will get to master from there

PR #600, opened 2016-04-26
Add the default value to the "loadyaml" function

dmitryilyin:
This value will be returned if the is no file to load or a file could not be parsed.
It's similar to the "parseyaml" function's default value.

Add the "loadjson" function too.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -660,10 +660,40 @@ Returns the keys of a hash as an array. *Type*: rvalue.
 
 #### `loadyaml`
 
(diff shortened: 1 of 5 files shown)


dmitryilyin:
ping

jonnytpuppet:
@bmjen This PR looks good to me. Do you think we can merge this or would this be a breaking change?

bmjen:
This looks fine. It does not appear backwards incompatible, as any incorrect call to `loadyaml` without a default will still yield the same result.

Thanks for the contribution @dmitryilyin !

PR #601, opened 2016-04-27
(MODULES-1439) Adds any2bool function

petems:
* Basically a combination of `string2bool` and `num2bool`

So there's a request for this to be added to stdlib (https://tickets.puppetlabs.com/browse/MODULES-1439).

Questions:

* Do we want to muddle the waters by the string casting to numeric part to the check? So should '-1' be true because it's a string, or false because when num2bool its false?
* There's already a `any2bool` as part of [puppi](https://github.com/example42/puppi/blob/master/lib/puppet/parser/functions/any2bool.rb), should we give this a different name?

### DIFF
--- a/spec/functions/any2bool_spec.rb
+++ b/spec/functions/any2bool_spec.rb
@@ -1,0 +1,42 @@
+require 'spec_helper'
+
+describe 'any2bool' do
+  it { is_expected.not_to eq(nil) }
+  it { is_expected.to run.with_params().and_raise_error(Puppet::ParseError, /wrong number of arguments/i) }
+
+  it { is_expected.to run.with_params(true).and_return(true) }
+  it { is_expected.to run.with_params(false).and_return(false) }
+
+  it { is_expected.to run.with_params('1.5').and_return(true) }
+
+  describe 'when testing stringy values that mean "true"' do
+    [ 'TRUE','1', 't', 'y', 'true', 'yes'].each do |value|
+      it { is_expected.to run.with_params(value).and_return(true) }
+    end
+  end
+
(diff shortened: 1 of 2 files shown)


hunner:
> Do we want to muddle the waters by the string casting to numeric part to the check? So should '-1' be true because it's a string, or false because when num2bool its false?
Sure, that makes it work on puppet 3 and 4 the same way.

> There's already a any2bool as part of puppi, should we give this a different name?
I think that puppi's purpose is different from stdlib's purpose, so example42 can do namespacing as needed similar to when functions migrate from other modules to stdlib. Thanks for catching that!

nibalizer:
+1

bmjen:
Thanks @petems !

nibalizer:
Note that we need to call out the puppi incompatibility in the changelog for the next release.

petems:
👍

spec/functions/any2bool_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+  it { is_expected.to run.with_params('1.5').and_return(true) }
+
+  describe 'when testing stringy values that mean "true"' do
+    [ 'TRUE','1', 't', 'y', 'true', 'yes'].each do |value|
+      it { is_expected.to run.with_params(value).and_return(true) }
+    end
+  end
+
+  describe 'when testing stringy values that mean "false"' do
+    [ 'FALSE','', '0', 'f', 'n', 'false', 'no', 'undef', 'undefined', nil, :undef ].each do |value|
```
igalic: 'off' would also be good :D

PR #605, opened 2016-05-11
Add delete_regex

jyaworski:
To maintain backwards compatibility, add a third parameter to the delete
function to specify what type the second should be interpreted as. This
is needed for cases where the regex matcher is overzealous.

This was uncovered while testing this against @smoeding's sendmail module as seen here:

https://github.com/smoeding/puppet-sendmail/issues/8

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -263,15 +263,23 @@ Takes a resource reference and an optional hash of attributes. Returns 'true' if
 
 #### `delete`
 
-Deletes all instances of a given element from an array, substring from a string, or key from a hash. Arrays and hashes may also match on regular expressions by providing a full regular expression.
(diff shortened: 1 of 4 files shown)


smoeding:
Either that or introduce a new function `delete_regex` to clearly indicate a new functionality.

Trying to add the regex functionality to the `delete` function without giving the user control where the parameter should be used as literal or as regex is a disaster waiting to happen. Just think of all the characters (like '.', '*' or '?') that have a special meaning when used in a regex.

hunner:
Yeah, I think `delete_regex` would make more sense from a UX perspective :)

jyaworski:
@hunner done.

PR #607, opened 2016-05-26
Remove symlinks, because that breaks things

hunner:


igalic:
my guess is that the failures are unrelated

hunner:
I think they actually are related, I just haven't untangled why the symlinks are needed.

tphoney:
@hunner have you had any further luck with this ?

DavidS:
This has been addressed in #616

PR #608, opened 2016-06-20
Add clientcert_rand (reopens #572)

hashbrowncipher:
This reopens PR #572, which was closed due to inactivity on my end.


hunner:
Thanks for your submission @hashbrowncipher ; instead of introducing yet another *_rand function to stdlib, would you be interested in using https://github.com/puppetlabs/puppetlabs-stdlib/blob/master/lib/puppet/parser/functions/seeded_rand.rb and seeding it with `$clientcert` ? eg `clientcert_rand(max) = seeded_rand($::clientcert, $max)`

PR #609, opened 2016-06-22
[REVIEW ONLY] (FM-5384) Prototype first batch of validate_* type aliases

DavidS:
This puppet 4 work is designed to provide a easy way off the validate_* functions. The README in types/compat explains more how this is intended to work.


types/compat/absolute_path.pp
```diff
@@ -0,0 +1,7 @@
+# Emulate the is_absolute_path and validate_absolute_path functions
+#
+# The first pattern is originally from is_absolute_path, which had it from 2.7.x's lib/puppet/util.rb Puppet::Util.absolute_path?
+# slash = '[\\\\/]'
+# name = '[^\\\\/]+'
+# %r!^(([A-Z]:#{slash})|(#{slash}#{slash}#{name}#{slash}#{name})|(#{slash}#{slash}\?#{slash}#{name}))!i,
+type Stdlib::Compat::Absolute_path = Variant[Pattern[/^(([a-zA-Z]:[\\\/])|([\\\/][\\\/][^\\\/]+[\\\/][^\\\/]+)|([\\\/][\\\/]\?[\\\/][^\\\/]+))/], Pattern[/^\//]]
```
igalic: stuff like this makes me wish a type could be a function…

hlindberg: The types Path, AbsolutePath, and RelativePath together with WindowsPath would be good additions to Puppet's type system btw.

types/compat/numeric.pp
```diff
@@ -0,0 +1,23 @@
...
+#   assert_type(Integer[0, 10], $value) |$expected, $actual| {
+#     warning("The 'value' parameter for the 'ntp' class has type ${actual}, but should be ${expected}.")
+#   }
+# }
+# ```
+#
+# > Note that you need to use Variant[Integer[0, 10], Float[0, 10]] if you want to match both integers and floating point numbers.
+#
+# This allows you to find all places where a consumers of your code call it with unexpected values.
+type Stdlib::Compat::Numeric = Variant[Numeric, Pattern[/^-?(?:(?:[1-9]\d*)|0)(?:\.\d+)?(?:[eE]-?\d+)?$/], Array[Variant[Numeric, Pattern[/^-?(?:(?:[1-9]\d*)|0)(?:\.\d+)?(?:[eE]-?\d+)?$/]]]]
```
hlindberg: You are allowed to use self referencing types - something like this should work:
`type Stdlib::Compat::Numeric = Variant[Numeric, Pattern[/^-?(?:(?:[1-9]\d*)|0)(?:\.\d+)?(?:[eE]-?\d+)?$/], Array[Stdlib::Compat::Numeric]]`
Although that would allow nested arrays of numeric, which you may not want

types/compat/re.pp
```diff
@@ -0,0 +1,3 @@
+# Emulate the validate_re function
+# validate_re(value, re) translates to Pattern[re], which is not valid puppet and can't be aliased.
```
hlindberg: Actually `Pattern[re]` is valid Puppet, but puppet does not (yet) have the ability to create parameterized aliases. Pattern accepts regexp in string form, regexp and (IIRC) a Regexp type.

types/compat/string.pp
```diff
@@ -0,0 +1,2 @@
+# Emulate the is_string and validate_string functions
+type Stdlib::Compat::String = String
```
hlindberg: Why would you use this compat type instead of just String ?

types/compat/numeric.pp
hlindberg: Another suggestion is to break out the part that is "Numeric in String format" as a separate Compat type, and then reference that in Compat::Numeric (i.e. the Pattern part).

types/compat/bool.pp
```diff
@@ -0,0 +1,2 @@
+# Emulate the is_bool and validate_bool functions
+type Stdlib::Compat::Bool = Boolean
```
hlindberg: Why would you use `Stdlib::Compat::Bool` instead of `Boolean` type directly ?

types/compat/README.md
```diff
@@ -0,0 +1,45 @@
...
+```
+
+
+After:
+```
+class ntp (
+  Stdlib::Compat::Bool              $broadcastclient,
+  Stdlib::Compat::Absolute_path     $config,
+  Stdlib::Compat::Array             $keys,
+  Optional[Stdlib::Compat::Numeric] $minpoll,
```
hlindberg: You probably want `Optional[Stdlib::Compat::Numeric] $minpoll = undef` here as a user would otherwise have to pass a value, or have data binding that defines the default.
Remember that `Optional` only means "also accepts `Undef`".

types/compat/absolute_path.pp
dalen: Creating a UnixPath type and WindowsPath type and having this as `Variant[UnixPath, WindowsPath]` would indeed be simpler (and in most cases you don't want to allow windows paths anyway :) )

ripienaar:
Why name these `Stdlib::Compat::Absolute_path` rather than just `Stdlib::Absolute_path`? It won't clash with anything without the `Compat` today but it seems to me in the near future when Puppet 4 is all thats supported this naming will be particularly odd?

dalen:
Wouldn't it be nicer to create 4.x versions of these functions that can emit a warning if they for example match `validate_numeric` but wouldn't match `Numeric`?

Those 4.x versions would only be loaded if the future parser or 4.x is used, so it would be pretty seamless for the user and then they can migrate straight to the new types instead of moving to compatibility types first.

hlindberg:
Agree with @dalen, 4.x functions would be better. They could be under control of the `--strict` option that allows users to specify if strictness warnings are disabled, emitted, or if they want an error. Stdlib could also possibly be integrated with the migration support API that is used by catalog preview. I can imagine adding general hooks in that API - although, just using the 4.x issue system would help a lot as warnings would then have issue-codes that allow better presentation (aggregation, reference to documentation how to deal with the issue etc.).

Without the functions there is little to help users migrating away from the Compat types which they then may be stuck with for ever, and it will be harder to deprecate the use of data types (we have no mechanism for that now).

types/compat/re.pp
DavidS: Yup, that is a very unfortunate choice of words. I'll fix that to ".. which is not directly mappable as a type alias, but can be specified as Pattern[re]."

types/compat/README.md
DavidS: You are right. While this is intended as a simple example, and not full-fledged code, having a `undef` default here makes sense.

types/compat/absolute_path.pp
DavidS: https://github.com/voxpupuli/puppet-tea/tree/master/types has this already, and better done that the validate_absolute_path function.

types/compat/numeric.pp
DavidS: I had that separate in a prior version, but chose to forego this for a more compact solution, to avoid any temptation to re-use this outside the very narrow focus of this change.

types/compat/absolute_path.pp
DavidS: @igalic can't tell whether you're facetious.

DavidS:
Thank you all for taking the time to have a look. Here's a summary of my answers. It mostly revolves around 

> If you do not need to remain backwards compatible to support existing users of your code, please do use puppet's core types directly, or have a look at voxpupuli's collection of extended types at https://github.com/voxpupuli/puppet-tea.

which I have added to the README.md to make the intent of these types clearer.

1. @ripienaar Why `::Compat::`? Because these types have a very specific use-case, and that is to mimic the `validate_*` functions. Given that the API of those functions is neither uniform, nor strict, I do not really want anyone to use these aliases for an extended time anyways. This is a throw-away namespace.

2. @hlindberg Why use the 1:1 mappings for existing core types? "Don't make me think." In the context of translating the functions calls, it makes it clear that `validate_*` to `Stdlib::Compat::*` is always safe, without needing to lookup special cases. The extra handwaving around parameterized types is already bad enough.

3. @dalen/@hlindberg Why not a function? Because having types in the signature of a class/define has very real benefits over just calling a function: location and content of error message, readability, documentation extraction. The goal is to get rid of the functions, not improve them.



I'll upload a version with a few editorial changes from the code comments in a few minutes. I'd love to hear from you if I could convince you of the (limited, but distinct) value of those types in the context of smooth upgrades.

types/compat/string.pp
DavidS: Please see longer form answer below

types/compat/bool.pp
DavidS: Please see longer form answer below

hlindberg:
I am thinking more like dalén; if you have to change, you can just as well change to the real thing and get help in doing so. While there is value in moving from function calls for validation, to types, the adding of a new set of functions that tell users which type to use and at the same time deprecate the use of the validate_ function seems like a helpful thing that requires 0 effort from the user. The Compat::T types should still be added for smoother transition from the functions - it is not an either or here.

DavidS:
So you're saying that - in addition to the proposed Compat types - the validate_ functions should gain warnings around "surprising" values being accepted?

DavidS:
@hlindberg  I've added a sketch of what I think your proposal was about as a separate commit (for easier reviewing). If that's what you meant, I also need to adapt the README of course.

hlindberg:
@DavidS yes, that is what I meant.

DavidS:
:100:

hlindberg:
@DavidS Something like what you implemented would be helpful, bearing in mind not to swamp the user with output.
You may want to re-implement the functions in question as 4.x functions so that they override the 3.x existing ones when future parser or 4.x is in effect (otherwise the method should probably continue to work the same way without any warnings). You can possibly detect this by checking `Puppet[:strict]`, it is one of `:off, :warning, :error` in later 4.x versions and `nil` (not implemented) in 3.x and early 4.x. When encountering a case that should be reported, there are methods that can be called that only warns once for a given key (of your choosing) (search the puppet code base for logic that is using Puppet[:strict] and you will find some examples of usage.

@dalen any more thoughts regarding this?

DavidS:
@hlindberg I tried re-implementing those functions in puppet4 API a year ago, when I joined the modules team, and it is a mess. The less this code needs to change before being deleted, the better.

Thanks for the pointers on warn once logic. That will be very handy!

DavidS:
Thank you all for your feedback. The work needed to get to a satisfactory result is bigger than expected. I've split this up into multiple tickets starting at [MODULES-3529](https://tickets.puppetlabs.com/browse/MODULES-3529). I'll link the other tickets there as I create them.

PR #611, opened 2016-06-29
(MODULES-3407) Clarify that 'after' in file_line accepts regex.

ntpttr:
Update the README to specify that the 'after' parameter in
the file_line resource accepts a regular expression.


DavidS:
Thanks!

PR #613, opened 2016-06-29
Add puppet_server fact to return agent's server

reidmv:
It is frequently useful to configure an agent to retrieve a resource
from it's configured master, or make further configuration adjustments
to itself based on what server it's using. Similar to the rationale for
stdlib providing a puppet_vardir fact, this commit adds a puppet_server
fact.

Note that the closest equivalent available today is $settings::server,
which returns only the MASTER's configured server, not the AGENT's. This
makes $settings::server unreliable, and not useful in a multi-master
deployment or a deployment involving a load balancer.

### DIFF
--- a/lib/facter/puppet_settings.rb
+++ b/lib/facter/puppet_settings.rb
@@ -33,3 +33,11 @@ Facter.add(:puppet_environmentpath) do
     end
   end
 end
+
+Facter.add(:puppet_server) do
+  setcode do
+    Facter::Util::PuppetSettings.with_puppet do
+      Puppet[:server]
+    end
+  end
+end


wilson208:
@reidmv I know this is an old PR, but if you are still interested in getting it merged, could you add some spec tests to accompany this change? If not, close this PR.

Thanks :)

reidmv:
@wilson208 Ah, yeah, I remember this.

Does your comment indicate that if this submission had tests, it would be merged?

Let's please be explicit on a viable/non-viable decision first regarding the utility/merge-ability of a fact called "puppet_server". After that's been made clear, and if it merits it, we can buff up the PR code quality to standard.

There's no point writing more code until a decision has been made that the fact merits inclusion in stdlib. (do you agree?)

reidmv:
@wilson208 I took a look at this and refactored to match the current codebase. As you can see, this fact is functionally identical to two existing facts (as is evidenced by the fact it belongs in the same file as them now), and those two facts do not have spec tests.

It would be nice to add tests but realistically this PR should not be gated on that.

PR #614, opened 2016-06-29
(MODULES-3543) Fix define_with_params to handle undef properly

DavidS:
As described in PUP-6422,

	ensure_resources('File[/tmp/a]', { owner => undef })

would not actually create the file. This fixes it, and adds tests to prove
it.


DavidS:
Build is green except for the known dig() issues.

PR #615, opened 2016-06-30
(MODULES-3543) Fixup defined_with_params to work on all puppet versions

DavidS:


DavidS:
Everything green except the dig failure (PUP-6460)

PR #617, opened 2016-07-05
(MODULES-3529) add deprecation function

tphoney:


README.markdown
```diff
@@ -289,6 +289,10 @@ Deletes all instances of a given value from a hash. For example, `delete_values(
 
 Deletes all instances of the undef value from an array or hash. For example, `$hash = delete_undef_values({a=>'A', b=>'', c=>undef, d => false})` returns {a => 'A', b => '', d => false}. *Type*: rvalue.
 
+#### `deprecation`
+
+Function to print deprecation warnings, it will display a provided message as well as the caller. The deprecation message will only be displayed once per caller. It is affected by the puppet setting 'strict', which can be set to :error, :off and :warning (default) *Type*: String.
```
DavidS: Doesn't tell reader how `strict` affects the function.

lib/puppet/functions/deprecation.rb
```diff
@@ -0,0 +1,28 @@
...
+#
+require "puppet_x/puppetlabs/stdlib/deprecation_state"
+
+Puppet::Functions.create_function(:deprecation) do
+  dispatch :deprecation do
+    param 'String', :message
+  end
+
+  def deprecation(message)
+    called_by = Kernel.caller.first
```
DavidS: We talked about that we want to have the puppet call site (file/line), not the ruby caller (which will always be the same). Maybe @hlindberg knows the magic incantation?

spec/functions/deprecation_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+    end
+  end
+end
+
+if Puppet.version.to_f >= 4.0
+  describe 'deprecation' do
+    before(:each) {
+      # this is to reset the deprecation warning track, and strict variable to default
+      PuppetX::Puppetlabs::Stdlib::DeprecationState.class_variable_set :@@deprecation_warning_said, []
+      Puppet.settings[:strict] = :warning
```
DavidS: Affecting global state should be reversed in `after` block, to avoid poisoning the environment for later tests.

lib/puppet_x/puppetlabs/stdlib/deprecation_state.rb
```diff
@@ -0,0 +1,16 @@
+module PuppetX
+  module Puppetlabs
+    module Stdlib
+      class DeprecationState
+        # This is so we display one message per caller
+        @@deprecation_warning_said = []
```
hlindberg: This is not good - no @@vars please - they hold on to the data forever.

lib/puppet/functions/deprecation.rb
hlindberg: There is a way to get calling puppet scope. No easy way to get file line as you need to find the calling instruction to do that. There is a ticket to improve so that calling file/line would be available to functions.
A 4.x function that wants access to calling scope must be an internal / system function. See the `defined` 4.x function in Puppet's code base for an example how to do an internal function, and how to get the calling scope parameter in the dispatcher.

lib/puppet/functions/deprecation.rb
```diff
@@ -0,0 +1,28 @@
...
+      # depending on configuration setting of strict
+      case Puppet.settings[:strict]
+      when :off
+        # do nothing 
+      when :error
+        Puppet.err("#{message}, called by #{called_by}")
+      else
+        Puppet.warning("#{message}, called by #{called_by}")
+      end
+      PuppetX::Puppetlabs::Stdlib::DeprecationState.add_deprecation_warning_said(called_by)
```
hlindberg: Puppet already has call-once deprecation feature. No need to write another. Also, the deprecation per call site may not be the best - there is an api where a key can be provided that serves as a uniqueness key for the deprecation being issued.
A call may look like this:
```
            Puppet.warn_once(:unresolved_module_dependencies,
                             "unresolved_dependencies_for_module_#{from_module_data.name}",
                             msg)
```
A key, that can be used to denote this particular type of deprecation (for instance also adding it to disabled_warnings). The second is the uniqueness key - in the example, the deprecation can appear once per module name. The `msg` is the message text including any positional information that is formatted by the user/caller of the method.

lib/puppet_x/puppetlabs/stdlib/deprecation_state.rb
```diff
@@ -0,0 +1,16 @@
...
+        def self.deprecation_warning_said(calling_method)
+          @@deprecation_warning_said.include?(calling_method)
+        end
+        def self.add_deprecation_warning_said(calling_method)
+          @@deprecation_warning_said.push(calling_method) unless @@deprecation_warning_said.include?(calling_method)
+        end
+      end
+    end
+  end
+end
```
hlindberg: This construct should be replaced by called to `Puppet.warn_once`

spec/functions/deprecation_spec.rb
tphoney: resolved

tphoney:
@hlindberg @DavidS I think that your last recommendation about using catalog_preview may be outside of our requirements. 

I have a question: if Puppet.settings[:strict] = :error, what logging / error method would you recommend i use. http://www.rubydoc.info/gems/puppet/4.4.1/Puppet/Util/Logging  
We still only want to display the message once

tphoney:
@hlindberg @DavidS i have made the suggested changes, what do you think of the README, i pulled in your comments @hlindberg .

hlindberg:
there is an :err (or is it :error - I mix them up all the time) log level. This level does not cause puppet to fail. The spirit of {{\--strict=error}} is that you want the evaluation to actually stop as you would use that level to be pedantic and enforce that all deprecations have been dealt with. I suggest failing in that case rather than just logging an error.

README.markdown
```diff
@@ -289,6 +289,10 @@ Deletes all instances of a given value from a hash. For example, `delete_values(
 
 Deletes all instances of the undef value from an array or hash. For example, `$hash = delete_undef_values({a=>'A', b=>'', c=>undef, d => false})` returns {a => 'A', b => '', d => false}. *Type*: rvalue.
 
+#### `deprecation`
+
+Function to print deprecation warnings, Logs a (non deprecation) warning once for a given key. A key, that can be used to denote this particular type of deprecation (for instance also adding it to disabled_warnings). The second is the uniqueness key - in the example, the deprecation can appear once per module name. The msg is the message text including any positional information that is formatted by the user/caller of the method  It is affected by the puppet setting 'strict', which can be set to :error (outputs as an error message), :off (no message / error is displayed) and :warning (default, outputs a warning)  *Type*: String, String, String.
+
```
hlindberg: adding to disabled_warnings requires an update to puppet as it lists all keys that can be disabled - so that part will not work without a change to that setting in puppet.

DavidS: That is very unfortunate, as we hoped to expose this functionality to the user. Specifically, allowing them(us) to specify finegrained deprecations, at least at the module level.

DavidS:
Given the vague descriptions of the core infrastructure we're using here, a few integration tests exercising the whole stack would be great.

Also, please put the ticket number in the Commit and PR title for automatic linking to JIRA.

README.markdown
hlindberg: Actually what I said is misleading, the first parameter `kind` to warn once, is the key to `disabled_warnings`. So, `kind` can be set to `'deprecation'` which means it can be disabled. Here is an example from puppet source: ` Puppet.warn_once('deprecation', 'symbol_comparison', 'Comparing Symbols to non-Symbol values is deprecated')`. If a new kind is wanted and that it is also wanted to be able to disable that kind then puppet must be updated. (It would not be unreasonable to support any kind in the disable_warnings (with the only downside that typos will not be caught)).

tphoney:
@hlindberg @DavidS changes made.

tphoney:
@DavidS i didnt know you were pointing that comment at me. What you mean by full stack testing ?

DavidS:
I was thinking about something like this: https://github.com/tphoney/puppetlabs-stdlib/compare/add_deprecate_function...DavidS:MODULES-3529-acceptance-tests?expand=1

tphoney:
@DavidS i changed the check for puppet4 in the acceptance test to 

> if: get_puppet_version =~ /^4/ do 
following other acceptance tests.

DavidS:
@tphoney please apply shellquoting and then fix the fialing tests:

```
diff --git a/spec/acceptance/deprecation_spec.rb b/spec/acceptance/deprecation_spec.rb
index 54cdfdb..4215e6c 100644
--- a/spec/acceptance/deprecation_spec.rb
+++ b/spec/acceptance/deprecation_spec.rb
@@ -1,5 +1,6 @@
 #! /usr/bin/env ruby -S rspec
 require 'spec_helper_acceptance'
+require 'shellwords'
 
 describe 'deprecation function' do
   before :each do
@@ -12,7 +13,7 @@ describe 'deprecation function' do
       deprecation('key', 'message')
       file { '/tmp/deprecation': ensure => present }
       EOS
-      @result = on(default, puppet('apply', '--strict=error', '-e', pp), acceptable_exit_codes: (0...256))
+      @result = on(default, puppet('apply', '--strict=error', '-e', Shellwords.shellescape(pp)), acceptable_exit_codes: (0...256))
     end
 
     it "should return an error" do
@@ -34,7 +35,7 @@ describe 'deprecation function' do
       deprecation('key', 'message')
       file { '/tmp/deprecation': ensure => present }
       EOS
-      @result = on(default, puppet('apply', '--strict=warning', '-e', pp), acceptable_exit_codes: (0...256))
+      @result = on(default, puppet('apply', '--strict=warning', '-e', Shellwords.shellescape(pp)), acceptable_exit_codes: (0...256))
     end
 
     it "should not return an error" do
@@ -56,7 +57,7 @@ describe 'deprecation function' do
       deprecation('key', 'message')
       file { '/tmp/deprecation': ensure => present }
       EOS
-      @result = on(default, puppet('apply', '--strict=off', '-e', pp), acceptable_exit_codes: (0...256))
+      @result = on(default, puppet('apply', '--strict=off', '-e', Shellwords.shellescape(pp)), acceptable_exit_codes: (0...256))
     end
 
     it "should not return an error" do
```

tphoney:
@DavidS boom !

DavidS:
retriggered jenkins to pickup the new beaker version

tphoney:
@DavidS tests are now passing

DavidS:
Wohoo!! :tada:

PR #618, opened 2016-07-08
(MODULES-3568) Move dig to dig44 and deprecate dig

ntpttr:
A new version of dig was introduced in Puppet 4.5.0 that
isn't compatible with the stdlib version of dig. To maintain
backwards compatibility and ensure that tests for stdlib aren't
broken, this patch renames dig to dig44 and adds a deprecation
warning to the stdlib dig function.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -295,6 +295,8 @@ Returns the difference between two arrays. The returned array is a copy of the o
 
 #### `dig`
 
+DEPRECATED: This function has been replaced in Puppet 4.5.0, use dig44() for backwards compatibility or use the new version.
+
 *Type*: rvalue.
 
 Retrieves a value within multiple layers of hashes and arrays via an array of keys containing a path. The function goes through the structure by each path component and tries to return the value at the end of the path.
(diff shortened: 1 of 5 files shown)


ntpttr:
Triggered a recheck for failed gem installs in the CI

README.markdown
```diff
@@ -293,7 +293,7 @@ Deletes all instances of the undef value from an array or hash. For example, `$h
 
 Returns the difference between two arrays. The returned array is a copy of the original array, removing any items that also appear in the second array. For example, `difference(["a","b","c"],["b","c","d"])` returns ["a"]. *Type*: rvalue.
 
-#### `dig`
+#### `dig44`
```
DavidS: Please keep the entry for the old dig() around and replace its description with the deprecation warning to reduce possible confusion.

spec/functions/dig_spec.rb
```diff
@@ -1,13 +1,50 @@
...
+  end
+
+  it "should give a deprecation warning when called" do
+    scope.expects(:warning).with("dig() DEPRECATED: This function has been replaced in Puppet 4.5.0, please use dig44() for backwards compatibility or use the new version.")
+    scope.function_dig([{}, []])
+  end
+
+  it "should raise a ParseError if there are less than 2 arguments" do
+    expect { scope.function_dig([]) }.to raise_error(Puppet::ParseError)
+  end
```
DavidS: Please only test that the warning is triggered and dig44 is called. All the rest of the tests do not test the dig() function and can go away.

DavidS:
Thanks for your contribution, @ntpttr! The travis check is failing on an unrelated issue, you can ignore that for now.

tphoney:
great work !

README.markdown
ntpttr: Okay, added that back in.

spec/functions/dig_spec.rb
ntpttr: Done.

ntpttr:
It looks like sone other PRs pass the ruby 1.9.3 check for some reason so rechecking again to see if it works

DavidS:
Thanks!

PR #620, opened 2016-07-15
add hash2ini function

mmckinst:


lib/puppet/parser/functions/hash2ini.rb
```diff
@@ -0,0 +1,47 @@
...
+    end
+
+
+
+    ini = []
+    ini << settings['header'] << nil
+    h.keys.each do |section|
+      ini << "#{settings['section_prefix']}#{section}#{settings['section_suffix']}"
+      h[section].each do |k, v|
+        ini << "#{k}#{settings['key_val_separator']}#{settings['quote_char']}#{v}#{settings['quote_char']}"
```
hunner: This will behave strangely if `v` is not a string.

hunner:
I don't know if I feel comfortable adding this to stdlib. Perhaps needing this function indicates that the inifile module needs a way to manage the entire file instead of just by parts? This is possible with custom types that implement self.instances for purging, but not with ini_setting directly yet. It could be done though!

lib/puppet/parser/functions/hash2ini.rb
mmckinst: This is true. I can document that a string is expected for the value, all merging of of arrays of values must be done beforehand. There's no real standard on arrays of values for ini files so I'd just leave it up to the programmer to do what is necessary rather than put the logic in hash2ini().

hunner:
So, thinking about this it would be great if the inifile module got a new `ini_file` type that enabled purging via evalgenerate similar to how firewallchain does it. Here's an example interface:
```puppet
$hash = {
  'main' => {
    'logging' => 'INFO',
    'limit'   => 314,
    'awesome' => true,
  },
  'dev' => {
    'logging'      => 'DEBUG',
    'log_location' => '/var/log/dev.log',
  }
}
$defaults = {
  path              => '/tmp/foo.ini',
  key_val_separator => ': ',
}
create_ini_settings($hash,$defaults)

ini_file { '/tmp/foo.ini':
  purge    => true,
}
```

mmckinst:
feature request filed at https://tickets.puppetlabs.com/browse/MODULES-3644 . TBD how it will be done.

mmckinst:
@hunner what are your thoughts on a hash2yaml and hash2json that work the same as the above. use cases are also config files written in YAML and JSON that you want to manage 100% from puppet. 

FWIW, a function similar to hash2yaml was proposed in PR #259

PR #629, opened 2016-08-03
(modules-3533) deprecation for 3.x number function

tphoney:


tphoney:
@DavidS RFC please, i think what i am missing here is:
- README updates
- finalised deprecation warning

lib/puppet/parser/functions/is_float.rb
```diff
@@ -8,6 +8,8 @@ module Puppet::Parser::Functions
     EOS
   ) do |arguments|
 
+    function_deprecation([:is_float, 'This method is deprecated, please use the stdlib validate_legacy function, with Stdlib::Compat::Numeric. There is further documentation     for validate_legacy function the README.'])
```
DavidS: This (and the messages below) have copy&paste errors
* type name
* spacing

types/compat/numeric.pp
```diff
@@ -0,0 +1,23 @@
...
+#   assert_type(Integer[0, 10], $value) |$expected, $actual| {
+#     warning("The 'value' parameter for the 'ntp' class has type ${actual}, but should be ${expected}.")
+#   }
+# }
+# ```
+#
+# > Note that you need to use Variant[Integer[0, 10], Float[0, 10]] if you want to match both integers and floating point numbers.
+#
+# This allows you to find all places where a consumers of your code call it with unexpected values.
+type Stdlib::Compat::Numeric = Variant[Numeric, Pattern[/^-?(?:(?:[1-9]\d*)|0)(?:\.\d+)?(?:[eE]-?\d+)?$/], Array[Variant[Numeric, Pattern[/^-?(?:(?:[1-9]\d*)|0)(?:\.\d+)?(?:[eE]-?\d+)?$/]]]]
```
DavidS: There needs to be one `Stdlib::Compat::X` type for each `validate_X` function, that matches exactly what the function would match. Please add the missing ones for integer and float.

lib/puppet/parser/functions/validate_integer.rb
```diff
@@ -53,6 +53,8 @@ module Puppet::Parser::Functions
 
     ENDHEREDOC
 
+    function_deprecation([:validate_integer, 'This method is deprecated, please use the stdlib validate_legacy function, with Stdlib::Compat::Numeric. There is further documentation for validate_legacy function the README.'])
```
DavidS: 1. should all of these deprecation use the same key?
2. all of these deprecations should point to the right type (not ::Numeric, when it should be ::Integer)
3. "There is further documentation for validate_legacy function **>>in<<** the README"

PR #631, opened 2016-08-05
Fix markdown indentation

smortex:
The old indentation is weird.


tphoney:
Thanks for the fix ! @smortex

PR #630, opened 2016-08-04
(MODULES-3540) Addition of validate legacy function

HelenCampbell:

PR #633, opened 2016-08-08
(modules-3532) deprecate string type checks

tphoney:


seangriff:
Where is function_deprecation() defined?

tphoney:
@seangriff function_deprecation change is incoming.

tphoney:
@DavidS This is ready for review.

DavidS:
@seangriff @tphoney function_deprecation is from https://github.com/puppetlabs/puppetlabs-stdlib/blob/master/lib/puppet/parser/functions/deprecation.rb

tphoney:
@ekohl it is in this pr https://github.com/puppetlabs/puppetlabs-stdlib/pull/630

ekohl:
@tphoney to me it's odd to deprecate something when there's no replacement yet. Some of us still want to support Puppet 3 and now I'm flooded with deprecation warnings I can't solve. For every test run in rspec this generates about 60 lines of deprecations. Given you often 10s or 100s of tests in rspec, this change generates thousands of lines making the output of rspec unreadable.

DavidS:
Hi @ekohl , thanks for your feedback! I've already created follow-up tickets for addressing those issues: https://tickets.puppetlabs.com/browse/MODULES-3714 and https://tickets.puppetlabs.com/browse/MODULES-3717 . Those are scheduled for the next two weeks. As TP pointed out, the validate_legacy function is still in flight in #630, and should land within the next few days.

There are quite a few pieces in flight to figure out the best way to get people off puppet 3, and make use of the new puppet 4 features. The master branch will continue to be a bit rough until we can land everything.

ekohl:
@DavidS thanks for considering this. An upgrade guide for module writers (be it in README, a blog post or some other document) would be highly appreciated.

DavidS:
@ekohl [Status: Open](https://tickets.puppetlabs.com/browse/MODULES-3628] ; but first, we have to do all the codezz :-)

ekohl:
@DavidS it appears that's only available for logged in users, but I trust it'll be done.

DavidS:
anyone can register on the site, but I do understand, if you don't want to.

roidelapluie:
I don't have access even logged in

DavidS:
@ekohl @roidelapluie apologies. Marketing shite. Anyways, docs && blog post are in the making.

bcambl:
@DavidS 
I noticed a ton of deprecation warning today in my CI logs and some quick research brought me here.
Your statement of getting everyone off Puppet 3 is somewhat concerning because Red Hat Satellite 6.2 just went GA on 2016-07-27 with Puppet 3.8.6 (6.0 -> 6.1.9 = 3.6.2).
FWIW.

12ww1160:
Since this merge went into place, puppet parser fails to validate the files under types. In particular, the Syntax error at '=' at /puppetlabs-stdlib-master/types/compat/absolute_path.pp:7
This is true for any pp in the path containing the '=' under Puppet 3.  CI running 'puppet parser validate' fails now completely all catalogs with stdlib included. 

About that idea to **__'get people off puppet 3'__**: There are many companies (customers), which **__will not or cannot upgrade__** to puppet 4 for long time to come for various reasons. Backward compatibility should be maintained as long as puppet 3 is supported.

ripienaar:
@12ww1160 disagree. Backwards compatibility is provided through the convenience of installing a compatible version. 

We have to move forward and adopt new features as they come. Those who refuse to move forward do so at their own cost - by managing dependencies and installing a version that works for them and not getting new features. 

Why does everyone have to suffer to cater to those?

12ww1160:
@ripienaar  the README clearly states that this [version supports puppet 3](https://github.com/puppetlabs/puppetlabs-stdlib/tree/master#version-compatibility), which is now not a given as basic tests are failing. I'd call that bug .

ripienaar:
@12ww1160 I don't think this has been released? So it's not a "version" at all. Master is not a version. There is lots of ongoing work that appears to be towards a future version with puppet 4 support.

12ww1160:
my bad then, I was following the master. :-)

ripienaar:
@12ww1160 best to use releases via the forge. They do what they say. Master is in flux.

ekohl:
@ripienaar IMHO it's still a bug: either the module should support puppet 3 or the README should say things are in flux/puppet 3 is being dropped.

ripienaar:
@ekohl presumably once the next release arrived it will be clear? Do you expect all work between releases in a single commit and a master branch to always be correct and reflect the reality of a next release? Seems unreasonable.

ekohl:
@ripienaar I do think that should be a goal. So a commit that introduces flux should also update the readme. Thanks can and will slip through, but I don't think you should give up in advance.

DavidS:
@12ww1160 please exclude the `types` directory from syntax checking, if your puppet version doesn't yet support puppet types. See https://github.com/puppetlabs/modulesync_configs/blob/master/moduleroot/Rakefile#L10-L14 for an example how to do it.

@bcambl @ekohl : upcoming releases of stdlib will still support puppet3. Not all new features will be available, or make sense though. There will be a way to easily disable the deprecation warnings, if you're not yet ready to update to puppet 4.

mmckinst:
It looks like the validate_legacy functions is now in master (PR #630) as of a couple hours ago, so people can start using it.

Checking out stdlib from the master branch on git is used by puppetlabs in a bunch of their modules ([ntp](https://github.com/puppetlabs/puppetlabs-ntp/blob/master/.fixtures.yml), [apache](https://github.com/puppetlabs/puppetlabs-apache/blob/master/.fixtures.yml), [mysql](https://github.com/puppetlabs/puppetlabs-mysql/blob/master/.fixtures.yml)),  if someone was using one of those modules as a guide when creating their own module (as many people do) you end up with a lot of people having the same problem.

I'll probably change my code to not check out the master branch but I would at least expect the master branch to not be a WIP. If I'm getting deprecation warnings, telling me to use validate_legacy, I expect the validate_legacy function will exist so I can update my code.

ekohl:
I think https://github.com/puppetlabs/puppetlabs-stdlib/pull/636 will reduce the logging pain this change caused.

PR #634, opened 2016-08-12
[PUP-6611] Unexpected result from getparam if param is false

xepa:
Return if resource[param] is found (instead of if the value of resource[param] evaluates to true)


DavidS:
Thank you for finding and analysing this! Closing in favor of #663 which also addresses the test failures.

PR #636, opened 2016-08-15
(maint) Switch 3.x deprecation() to use Puppet warning logger

domcleal:
The deprecation function was calling the `Kernel#warn` function which
prints to stderr, rather than the Puppet logger. This causes problems
for Puppet module tests on Travis CI, which has a cap on the amount of
stdout/err permitted in its logs and also prevents users from finding
the deprecation warnings when running under a Puppet master.

### DIFF
--- a/lib/puppet/parser/functions/deprecation.rb
+++ b/lib/puppet/parser/functions/deprecation.rb
@@ -10,6 +10,6 @@ EOS
     key = arguments[0]
     message = arguments[1]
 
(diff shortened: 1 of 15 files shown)


DavidS:
Thanks for catching that. Do you still have time to address/fix  the other failing tests? Probably just needs more s/warn/warning.

domcleal:
Of course, PR updated with hopefully all instances changed correctly.

DavidS:
Thanks a lot!

PR #637, opened 2016-08-16
(MODULES-3534) Deprecation of ip functions

HelenCampbell:


DavidS:
This fails travis; and it should also do the s/warn/warning/ as demoed in https://github.com/puppetlabs/puppetlabs-stdlib/commit/39148468abbd9b8af74b776eb49f0a8388fc8541 from yesterday.

HelenCampbell:
@DavidS Yeah, it's still a work in progress. And thanks, changing that.

PR #638, opened 2016-08-16
Change all matched lines

scornelissen85:
When I add this:

```
file_line{ 'test':
  path => '/tmp/test.txt',
  line => 'new line',
  match => '^new(.*)',
  multiple => true,
  replace => true,
}
```

And I have this file (/tmp/test.txt):

```
new line
new line 2
```

Then there is nothing changed, because the line actually exists. I thought that file_line always ensured that all matched lines are changed, but it isn't. I'm not sure if this is a fix or more of a hack (I think the latter one). Also, is there an reason for not changing all matches?

### DIFF
--- a/lib/puppet/provider/file_line/ruby.rb
+++ b/lib/puppet/provider/file_line/ruby.rb
@@ -2,6 +2,8 @@ Puppet::Type.type(:file_line).provide(:ruby) do
   def exists?
     if resource[:replace].to_s != 'true' and count_matches(match_regex) > 0
       true
+    elsif resource[:replace].to_s != 'false' and count_matches(match_regex) > 1 and resource[:multiple].to_s != 'false'
+      false
     else
       lines.find do |line|
         line.chomp == resource[:line].chomp


tphoney:
@scornelissen85 thanks for the fix, could you have a look at the failing CI tests ?

wilson208:
@scornelissen85 I am closing this PR because it has not seen any activity in a number of months. Please feel free to reopen if you want to work on this again. Thanks!

lib/puppet/provider/file_line/ruby.rb
```diff
@@ -2,6 +2,8 @@
   def exists?
     if resource[:replace].to_s != 'true' and count_matches(match_regex) > 0
       true
+    elsif resource[:replace].to_s != 'false' and count_matches(match_regex) > 1 and resource[:multiple].to_s != 'false'
```
eputnam: one problem i see with this is that :replace defaults to true so any time its not explicitly set, this statement catches it. then, if match is not also set, it fails because `match_regex` would be nil.

PR #641, opened 2016-08-23
Deprecation function to be mutable in all cases

HelenCampbell:


lib/puppet/parser/functions/deprecation.rb
```diff
@@ -4,12 +4,15 @@ module Puppet::Parser::Functions
 EOS
   ) do |arguments|
 
+
```
tphoney: extra line not needed

spec/functions/is_array_spec.rb
```diff
@@ -21,4 +16,17 @@
   it { is_expected.to run.with_params('one').and_return(false) }
   it { is_expected.to run.with_params(1).and_return(false) }
   it { is_expected.to run.with_params({}).and_return(false) }
+  context 'Checking for deprecation warning' do
```
tphoney: should this be at the bottom ? it being at the top alerts people that this function is about to be deprecated, even in the tests

spec/functions/deprecation_spec.rb
```diff
@@ -42,6 +42,7 @@
   end
 else
   describe 'deprecation' do
+    ENV['STDLIB_LOG_DEPRECATIONS'] = "true"
```
DavidS: This still needs to go in a `before` [hook](https://www.relishapp.com/rspec/rspec-core/docs/hooks/before-and-after-hooks), and there should be tests for each possible value(-set).

HelenCampbell: After discussion David and I decided the code change to add before hooks and contexts is too much for the simplicity of the change. Instead I will add after hooks to each of the test for cleanup.

PR #643, opened 2016-08-26
MODULES-3699 fixes deprecation msg test

eputnam:
The changed test originally created a file called /tmp/deprecation, which was not referenced anywhere else in the module and wasn't actually being used by the test except to run the catalog. Instead of going through changing the file location to better accommodate solaris and windows, I changed to a notify resource instead. It does the same thing as the file resource did for the test but is far less breakable.

### DIFF
--- a/spec/acceptance/deprecation_spec.rb
+++ b/spec/acceptance/deprecation_spec.rb
@@ -3,15 +3,12 @@ require 'spec_helper_acceptance'
 require 'shellwords'
 
 describe 'deprecation function' do
-  before :each do
-    FileUtils.rm_rf '/tmp/deprecation'
-  end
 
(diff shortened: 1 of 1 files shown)


bmjen:
@eputnam, can you add a description to the PR explaining why replacing the actual file resource with a notify is the fix? The notify itself doesn't really contribute anything to the test

DavidS:
1) to add the fix description to the commit, you'll need to "git commit --amend" to record the message in the actual git structure, and not only on github

2) the tests did use the file to ascertain that the resources was/was not created. Just removing that part weakens the test.

eputnam:
cancelled in favor of https://github.com/puppetlabs/puppetlabs-stdlib/pull/646

PR #644, opened 2016-08-30
Refactor dig44 function

dmitryilyin:
The current implementation of the dig44 function has the
following problems:
* Doesn't recognise Puppet4's :undef value as an empty value
  and doesn't return the default value.
* Doesn't make a different between false and nil value
  and returns the default value for a non-empty false value


dmitryilyin:
hello?

PR #646, opened 2016-09-01
MODULES-3699 Deprecation spec fix 2

eputnam:
1. The manifest containing the deprecation function needed to be modified (surrounded in double quotes) to work as a puppet apply command on windows. 

2. "exist" was not working on Solaris 10 because the underlying command doesn't actually exist on Solaris 10, so I changed it to "be_file" which works just fine for this. 

3. Added conditional logic to accommodate Windows paths.

### DIFF
--- a/spec/acceptance/deprecation_spec.rb
+++ b/spec/acceptance/deprecation_spec.rb
@@ -1,19 +1,29 @@
 #! /usr/bin/env ruby -S rspec
 require 'spec_helper_acceptance'
-require 'shellwords'
 
 describe 'deprecation function' do
-  before :each do
-    FileUtils.rm_rf '/tmp/deprecation'
+
+  if fact('operatingsystem') == 'windows'
+    test_file = 'C:/deprecation'
+    else
+    test_file = "/tmp/deprecation"
+  end
+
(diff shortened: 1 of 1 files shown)


spec/acceptance/deprecation_spec.rb
```diff
@@ -1,19 +1,28 @@
...
-  before :each do
-    FileUtils.rm_rf '/tmp/deprecation'
+
+  if fact('operatingsystem') == 'windows'
+    test_file = 'C:/deprecation'
+    else
+    test_file = "/tmp/deprecation"
+  end
+
+  add_file_manifest = "\"deprecation('key', 'message') file { '#{test_file}': ensure => present, content => 'test', }\""
```
bmjen: can you make this look like this instead:

```
add_file_manifest = <<-EOS
  deprecation('key', 'message')
  file { '#{test_file}': 
    ensure => present,
    content => 'test', 
  }
EOS
```

It's just easier to read and maintain.

Same with `remove_file_manifest`.

eputnam: It'd be totally easier to read and maintain that way, but since it's run on the command line (puppet apply -e style) instead of with apply_manifest, Windows seems to need it formatted this way.

PR #660, opened 2016-10-01
Make the IPv4|6 regex much more strict

daenney:
This will correctly match IPv4 and IPv6 addresses and discard invalid ones.


daenney:
The two sets in the matrix that fail seem to fail on Hiera syntax checks? And only on Ruby 2.1.5?

daenney:
Though much more strict it still passes all the tests that we had defined so this Won't Break Anything™️

daenney:
Should I extend this to support CIDR matching for IPv4 and 6 too? Both the old and the new regexes don't support notations like `::1/128` for example.

daenney:
Poke @hunner and @HelenCampbell

daenney:
@davids Do you know why the tests explode on Ruby 2.1.5?

DavidS:
@daenney it seems that a change somewhere causes rspec-core 2.99 to be installed. That seems to cause all sorts of issues.

DavidS:
@daenney I flushed travis' gem caches for stdlib, and rekicked your cells. This seems to have fixed the issue. I do not completely understand why, tho.

igalic:
has anyone considered asking @thrnio for merging the module into stdlib? https://github.com/thrnio/puppet-ip
(or maybe just depend on it?)

hunner:
I don't know that persisting bugs is a good option. Okay, so it'll be a slight migration hurdle for people with broken manifests; it's still a bug and should be fixed. No one should be taking advantage of this anyway and so shouldn't actually run into it.

DavidS:
@hunner the whole point of the ::compat:: types is bug-for-bug compatibility. There is no reason to NOT add properly designed IP types as Stdlib::IPv4Address, Stdlib::IPv6Address and Stdlib::IPAddress, other than nobody's currently doing the work for them.

hunner:
Oh I see I see, it's not just fixing them, but specifically making the change somewhere other than `Stdlib::Compat::*`

DavidS:
exactly.

PR #661, opened 2016-10-02
Empower sort to (recursively) sort a hash

daenney:
This only allows for sorting based on the key and will throw an error if the keys are of different types. It's mostly meant to help out with cases where we're dealing with JSON as input or output and want to guarantee that the order of things don't change between Puppet runs, potentially breaking idempotency.

### DIFF
--- a/spec/functions/sort_spec.rb
+++ b/spec/functions/sort_spec.rb
@@ -21,4 +21,9 @@ describe 'sort' do
     it { is_expected.to run.with_params('a').and_return('a') }
     it { is_expected.to run.with_params('cbda').and_return('abcd') }
   end
+  context 'when called with a hash' do
+    it { is_expected.to run.with_params({ 'c' => 1, 'a' => 2 }).and_return({ 'a' => 2, 'c' => 1}) }
+    it { is_expected.to run.with_params({ 'c' => 1, 5 => 2 }).and_raise_error(Puppet::ParseError, /mixed key types/) }
+    it { is_expected.to run.with_params({ 'c' => 1, 'a' => 2, 'b' => { 'i' => 'am', 'because' => 'we are'}  }).and_return({ 'a' => 2, 'b' => { 'because' => 'we are', 'i' => 'am' }, 'c' => 1}) }
+  end
 end
--- a/README.markdown
+++ b/README.markdown
@@ -964,7 +964,7 @@ Returns the number of elements in a string, an array or a hash. *Type*: rvalue.
 
 #### `sort`
 
-Sorts strings and arrays lexically. *Type*: rvalue.
+Sorts strings, arrays and hashes (recursively) lexically. It cannot sort hashes with keys of different types since they are not comparable. *Type*: rvalue.
 
 #### `squeeze`
 
--- a/lib/puppet/parser/functions/sort.rb
+++ b/lib/puppet/parser/functions/sort.rb
@@ -1,10 +1,19 @@
 #
 # sort.rb
 #
+def recursive_hash_sort(value)
+  value.keys.sort.reduce({}) do |seed, key|
+    seed[key] = value[key]
+    if seed[key].is_a?(Hash)
+      seed[key] = recursive_hash_sort(seed[key])
+    end
+    seed
+  end
+end
 
 module Puppet::Parser::Functions
   newfunction(:sort, :type => :rvalue, :doc => <<-EOS
-Sorts strings and arrays lexically.
+Sorts strings, arrays and hashes (recursively) lexically.
     EOS
   ) do |arguments|
 
@@ -19,8 +28,13 @@ Sorts strings and arrays lexically.
       value.sort
     elsif value.is_a?(String) then
       value.split("").sort.join("")
+    elsif value.is_a?(Hash) then
+      begin
+        recursive_hash_sort(value)
+      rescue ArgumentError
+        raise(Puppet::ParseError, "sort(): hash contains mixed key types")
+      end
     end
-
   end
 end
 


daenney:
There's something wrong with rspec and Ruby 2.1.5:

```
/home/travis/build/puppetlabs/puppetlabs-stdlib/vendor/bundle/ruby/2.1.0/gems/rspec-core-2.99.2/lib/rspec/core/hooks.rb:512:in `extract_scope_from': You must explicitly give a scope (:each, :all, or :suite) when using symbols as metadata for a hook. (ArgumentError)
```

daenney:
There also seems to be a new thing going on where in Ruby 2+ `:a` (symbol) and `'a'` (string) are actually comparable, which didn't used to be the case in Ruby 1.9.x as witnessed by the tests passing on 1.9 but failing on 2+.

What do we want to do with this case? I can change the test to simply try with a completely different key type entirely which will make the tests pass but perhaps lead to slightly surprising behaviour.

daenney:
Poke @hunner

daenney:
@davids Do you know why the tests explode on Ruby 2.1.5?

domcleal:
@daenney rebase, I think 8a8ebc4 fixed the test error.

daenney:
Thanks @domcleal! Lets see if this fixes it.

daenney:
Alright, after talking it through with @domcleal I've updated the mixed key test to use very different key types, int vs string to make the test pass.

The reasoning behind it is that when you're using this from Puppet code you can't actually create symbol keys, so it won't be an issue that they happen to be comparable on newer versions of Ruby. It can become a problem if you're calling this from other Ruby code but in that case it should be the responsibility of the caller to deal with it if it's an issue for them.

domcleal:
> There also seems to be a new thing going on where in Ruby 2+ :a (symbol) and 'a' (string) are actually comparable, which didn't used to be the case in Ruby 1.9.x as witnessed by the tests passing on 1.9 but failing on 2+.
> 
> What do we want to do with this case?

As mentioned on #voxpupuli, I wouldn't worry about this because it's not possible (AFAIK) to create a symbolised hash in the Puppet DSL. It could be done from another parser function (calling function_sort), but that's such an edge case that you might as well raise the error or even let ArgumentError bubble up to the user.

Regarding the general change itself, I don't know that it makes sense in Ruby because hashes are not ordered. Since 1.9 hashes do have a consistent iteration order, so a hash will always iterate using .each (etc) in the same order, but the guarantee may not hold between different instances of a hash, or if the contents change slightly.

I think if you intend to actually sort the hash then you need to return an _array_ like Hash#sort itself does:

```ruby
> {'a' => 2, 'b' => 1}.sort
 => [["a", 2], ["b", 1]] 
```

This allows the user to iterate over the result in a very well defined order, because the array's ordered.

It's _possible_ that recreating the hash as you're doing here _might_ cause the iteration order to equal the original insertion order, but I don't think that Ruby guarantees it (particularly when Puppet runs on both MRI and JRuby). I also don't know that Puppet will guarantee it, because the function will be relying on both Ruby and Puppet hash behaviour. I wouldn't rely on this!

spec/functions/sort_spec.rb
```diff
@@ -21,4 +21,9 @@
     it { is_expected.to run.with_params('a').and_return('a') }
     it { is_expected.to run.with_params('cbda').and_return('abcd') }
   end
+  context 'when called with a hash' do
+    it { is_expected.to run.with_params({ 'c' => 1, 'a' => 2 }).and_return({ 'a' => 2, 'c' => 1}) }
+    it { is_expected.to run.with_params({ 'c' => 1, 5 => 2 }).and_raise_error(Puppet::ParseError, /mixed key types/) }
+    it { is_expected.to run.with_params({ 'c' => 1, 'a' => 2, 'b' => { 'i' => 'am', 'because' => 'we are'}  }).and_return({ 'a' => 2, 'b' => { 'because' => 'we are', 'i' => 'am' }, 'c' => 1}) }
```
domcleal: This test passes if you change the order of the hashes - hashes are equal no matter the order of the keys when declared. (Kind of related to my long comment about guarantees.)

DavidS: Agreed: Since all of this hinges on ruby's weird thing of enumerating keys in insertion order, that should also be the thing being tested here.

daenney: Then I guess we could `.to_s` it in the test and check we get what's expected? I get the concern but it would also be helpful to get some indication on how to fix it 😄.

daenney:
I don't understand what you're getting at at all with that comment @davids.

DavidS:
It doesn't feel right to me to clutter a simple and functional method like `sort` with a narrow special case that is resting on an oddity of one of the underlying runtimes.

I'd much rather see a solution that provides the functionality you need in a way that is robust and contained. One such way is just using puppet4's iteration features, like noted above: `$hash.keys.sort.each |k| { $v = $hash[$k] ... } ` where `$k` and `$v` would be the keys and values, walked in key-sorted order.

daenney:
The whole idea for adding it to sort is b/c people actually want to sort hashes. There's a myriad of community modules that have just one piece of Ruby code included, to do exactly this. So instead of cargo-culling this piece of code around from module to module it makes sense to me to add it to sodlib.

DavidS:
I'm not opposed to adding the functionality to sodlib. I'm opposed to
overloading functions with too many edge-cases. I'd rather see a separate,
well-defined, documented and tested sort_hash(). The stdlib functions are
already minefields, as they are.

daenney:
So instead of doing what happens in just about every language, a singe `sort()` function that can handle the different or most common types we're going to add a specific one for Hashes but keep String and Array in one b/c..?

hunner:
I agree that it's silly for every module dev to have to implement `$hash.keys.sort.each |k| { $v = $hash[$k] ... } ` when we could add it to stdlib. If you say that hashes cannot be sorted in memory but only on consumption, then what we really need is a puppet 4 function lambda (http://docs.puppetlabs.com/puppet/latest/reference/lang_lambdas.html) that sorts the hash on consumption:
```puppet
$hash.sorted_each |k,v| { ... }
```

gdubicki:
I just spent the whole day solving this, a lot of it because I was more and more demotivated because I could not believe that so obvious and simple problem is not solved anywhere in the Puppet ecosystem yet.

I even run into https://gist.github.com/hlindberg/6ecaab3756d812b890d7a7bd73e57def by @hlindberg but because of [this](https://gist.github.com/hlindberg/6ecaab3756d812b890d7a7bd73e57def?permalink_comment_id=3092323#gistcomment-3092323) it was not useful to me.

As I am pragmatic I would be happy with any solution that is reasonably easy to find. Would a PR that would add `sort_hash()` function be accepted, @DavidS @hunner ?

ekohl:
I think everyone involved in this PR has moved on from the Puppet ecosystem. I don't have a particular opinion about this myself right now (though the concern noted in https://github.com/puppetlabs/puppetlabs-stdlib/pull/661#issuecomment-251348731 about relying on hash not guaranteeing an order still feels correct), I just want to let you know that you shouldn't expect a reply from those people.

PR #659, opened 2016-09-29
Ensure validate functions use Puppet 4 deprecation

HelenCampbell:

PR #662, opened 2016-10-04
Permit undef passed as `nil` to validate_string

domcleal:
When validate_string is called via the Puppet 4 deprecation wrappers
from deprecation_gen (introduced in 970852d), `undef` is passed as `nil`
where it was previously passed as `''` from the Puppet 3-style function
API.

This change explicitly permits a `nil` value in validate_string, and
adds a test case to `is_string` which also accepts the same.

Fixes test failures in apt, concat etc:

    Error while evaluating a Function Call, nil is not a string.  It looks to be a NilClass at apt/manifests/source.pp:23:3 [..]
    # ./spec/fixtures/modules/stdlib/lib/puppet/parser/functions/validate_string.rb:34:in `block (2 levels) in <module:Functions>'
    # ./spec/fixtures/modules/stdlib/lib/puppet/parser/functions/validate_string.rb:32:in `each'
    # ./spec/fixtures/modules/stdlib/lib/puppet/parser/functions/validate_string.rb:32:in `block in <module:Functions>'
    # puppet-4.7.0/lib/puppet/parser/functions.rb:174:in `block (2 levels) in newfunction'
    # puppet-4.7.0/lib/puppet/util/profiler/around_profiler.rb:58:in `profile'
    # puppet-4.7.0/lib/puppet/util/profiler.rb:51:in `profile'
    # puppet-4.7.0/lib/puppet/parser/functions.rb:167:in `block in newfunction'
    # ./spec/fixtures/modules/stdlib/lib/puppet_x/puppetlabs/stdlib/deprecation_gen.rb:13:in `block (2 levels) in deprecation_gen'

(Introduced by #659)

### DIFF
--- a/lib/puppet/parser/functions/validate_string.rb
+++ b/lib/puppet/parser/functions/validate_string.rb
@@ -30,7 +30,7 @@ module Puppet::Parser::Functions
     end
 
     args.each do |arg|
-      unless arg.is_a?(String)
+      unless arg.is_a?(String) || arg.nil?
         raise Puppet::ParseError, ("#{arg.inspect} is not a string.  It looks to be a #{arg.class}")
       end
     end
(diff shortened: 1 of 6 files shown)


DavidS:
Thanks!

PR #664, opened 2016-10-06
Type updates

HelenCampbell:
Includes the addition of several new types inc absolute path, unix path, windows path, https url and http url. 
Also includes a fix to some of the current tests for the compat types tests. The puppet version number in the tests was incorrect, and has been changed from > 4.0 to > 4.5.


types/absolutepath.pp
```diff
@@ -0,0 +1,2 @@
+# A strict absolutepath type
+type Stdlib::Absolutepath = Variant[::Stdlib::Windowspath, ::Stdlib::Unixpath]
```
ripienaar: these `::Stdlib` - the `::` - are not needed in Puppet 4

README.markdown
```diff
@@ -75,6 +75,13 @@ The `stdlib::stages` class declares various run stages for deploying infrastruct
 
 ### Types
 
+#### `absolutepath`
+
+A strict absolute path type. Uses a Variant of Unixpath and Windowspath types.
+
+Acceptable input examples:    /usr2/username/bin:/usr/local/bin:/usr/bin:.
+                              C:\\WINDOWS\\System32
+
```
DavidS: This should become "Resource Types" and "Data Types" to avoid mixing those two.

README.markdown
```diff
@@ -75,6 +75,13 @@ The `stdlib::stages` class declares various run stages for deploying infrastruct
 
 ### Types
 
+#### `absolutepath`
+
+A strict absolute path type. Uses a Variant of Unixpath and Windowspath types.
+
+Acceptable input examples:    /usr2/username/bin:/usr/local/bin:/usr/bin:.
```
DavidS: This example is not listed in the tests, but should be, to avoid breaking the README ;-)

spec/aliases/httpsurl_spec.rb
```diff
@@ -0,0 +1,38 @@
+require 'spec_helper'
+
+if Puppet.version.to_f >= 4.5
+  describe 'test::httpsurl', type: :class do
+    describe 'valid paths handling' do
```
DavidS: describe string copy&paste error

spec/aliases/unixpath_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+    describe 'invalid path handling' do
+      context 'garbage inputs' do
+        [
+          nil,
+          [ nil ],
+          [ nil, nil ],
+          { 'foo' => 'bar' },
+          { },
+          '',
+          "C:/whatever"
```
DavidS: These tests should contain at least one testcase with wrong slashing, and one with a NUL character.

spec/aliases/windowspath_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+      context 'garbage inputs' do
+        [
+          nil,
+          [ nil ],
+          [ nil, nil ],
+          { 'foo' => 'bar' },
+          { },
+          '',
+          "httds://notquiteright.org",
+          "/usr2/username/bin:/usr/local/bin:/usr/bin:."
```
DavidS: Please add more examples of (slightly) invalid windows paths here.

types/absolutepath.pp
HelenCampbell: Good catch, thanks!

README.markdown
HelenCampbell: Example is added to the absolutepath_spec.rb

PR #665, opened 2016-10-06
Ignore :undefined_variable "reason" in getvar

keymone:
`catch` returns value of second argument to `throw`, which until https://github.com/puppetlabs/puppet/commit/860a2761f334c964068038b3ef6853f08beb1df5#diff-bb62983f077c2b43d065be52de182209R514 was `nil`, but now is non-falsey reason for error. short-circuit using return and eval to nil if `throw` was caught.

### DIFF
--- a/lib/puppet/parser/functions/getvar.rb
+++ b/lib/puppet/parser/functions/getvar.rb
@@ -21,8 +21,10 @@ module Puppet::Parser::Functions
 
     begin
       catch(:undefined_variable) do
-        self.lookupvar("#{args[0]}")
+        return self.lookupvar("#{args[0]}")
       end
+      
+      nil # throw was caught
     rescue Puppet::ParseError # Eat the exception if strict_variables = true is set
     end
 


DavidS:
I could reproduce this on the quick with puppet 4.7.0 only by modifying the puppet source code (see below). Still, the code is more robust with this change so in it goes!

```
diff --git a/lib/puppet/parser/scope.rb b/lib/puppet/parser/scope.rb
index 57294c1..a1fb5f0 100644
--- a/lib/puppet/parser/scope.rb
+++ b/lib/puppet/parser/scope.rb
@@ -506,7 +506,7 @@ class Puppet::Parser::Scope
       if next_scope
         next_scope.lookupvar(name, options)
       else
-        variable_not_found(name)
+        variable_not_found(name, "error")
       end
     end
   end
```

keymone:
@DavidS what tripped me was a mis-used variable name: `getvar('$::some_var')` which raised exception "class $ not found" which then leaked into result of `getvar`. thanks for merge.

DavidS:
@keymone See https://github.com/DavidS/puppetlabs-stdlib/commit/ba675774186aafde5a8fa97f08f0163235ef4d8a . thanks for coming back!

PR #666, opened 2016-10-06
(MODULES-3590) Fix match_for_absence parameter

HAIL9000:
Prior to this commit, due to a bug in the exists? method in the
file_line provider, match_for_absence didn't work as described (or
at all really). Update the exists? logic so that match_for_absence
works as described.

Additionally add a unit test to prevent regressions and update the
documentation for the parameter to reflect the fact that it is
ignored when `ensure => present`.


DavidS:
Congrats to fixing this annoyance. And congrats to #666 :-D

PR #672, opened 2016-10-11
(MODULES-3961) emit more deprecation warnings

DavidS:
This now emits one deprecation warning for each function used (but not for each call-site). Prior to this, only a single deprecation warning would have been triggered, potentially misleading users.

Additionally this adds v4 deprecation stubs for the functions that were missed.

### DIFF
--- a/lib/puppet/functions/is_absolute_path.rb
+++ b/lib/puppet/functions/is_absolute_path.rb
@@ -1,0 +1,10 @@
+Puppet::Functions.create_function(:is_absolute_path, Puppet::Functions::InternalFunction) do
+  dispatch :deprecation_gen do
+    scope_param
+    optional_repeated_param 'Any', :args
+  end
+  def deprecation_gen(scope, *args)
+    call_function('deprecation', 'is_absolute_path', "This method is deprecated, please use match expressions with Stdlib::Compat::Absolute_Path instead. They are described at https://docs.puppet.com/puppet/latest/reference/lang_data_type.html#match-expressions.")
+    scope.send("function_is_absolute_path", args)
+  end
+end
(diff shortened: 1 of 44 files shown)


DavidS:
When reviewing, please do have a good look that I didn't do a bad copy and paste job somewhere in here please!

hunner:
👍 for merge, but https://github.com/puppetlabs/puppetlabs-stdlib/blob/release/spec/acceptance/is_string_spec.rb#L106 is failing because undef is not considered a string any more. Not changed by this PR, but perhaps it should be changed?

DavidS:
Ah, yeah, that already happened and was fixed in validate_string. Need to add the same fix here too...

lib/puppet/functions/is_absolute_path.rb
```diff
@@ -0,0 +1,10 @@
+Puppet::Functions.create_function(:is_absolute_path, Puppet::Functions::InternalFunction) do
+  dispatch :deprecation_gen do
+    scope_param
+    optional_repeated_param 'Any', :args
```
alexjfisher: @DavidS @HelenCampbell 
I've got failing builds again...

`optional_repeated_param` not available in puppet < 3.8.1 ??
https://github.com/puppetlabs/puppet/commit/c7419c10b409d860334ab0302c5b8da374e039d2

HelenCampbell: @alexjfisher @DavidS has found the solution this morning and is working on a workaround. And yep it's due to the optional_repeated_param not being available before puppet 3.8.1.

DavidS: See https://tickets.puppetlabs.com/browse/MODULES-3962 for details.

PR #673, opened 2016-10-12
(MODULES-3962) Rework v4 function shims to work on puppet 3.7 and 4.0.0

DavidS:
This is a workaround for PUP-4438 (fixed in https://github.com/puppetlabs/puppet/commit/e01c4dc924cd963ff6630008a5200fc6a2023b08, 4.1.0, 3.8.1).

It works by manually passing through the scope, instead of relying on the
InternalFunction class.


alexjfisher:
Just tested this PR and this fix works for me.

PR #676, opened 2016-10-17
(MODULES-3980) Fix ipv4 regex validator

ntpttr:
The current pattern for stdlib::compat::ipv4 is incorrect, and
will return true for any four numbers separated with periods.
This commit improves the regex to validate that the IP octets
are between 1 and 255.


ntpttr:
Looks like jobs with ruby < 2.0 are failing because of a dependency with parallel_tests

hunner:
Also see #660

hunner:
We are working on the parallel_tests fix via modulesync_config, so please don't edit the Gemfile directly (see https://github.com/puppetlabs/puppetlabs-stdlib/blob/master/Gemfile#L1 )

hunner:
Oh, I see you already did the modulesync :D

PR #678, opened 2016-10-20
(MODULES-3983) Update parallel_tests for ruby 2.0.0

pmcmaw:
Update to the config_defaults.yml to support parallel_tests for versions earlier than ruby 2.0.0.


DavidS:
This has been superseded by #682

PR #679, opened 2016-10-21
Fix/master/modules 3980

hunner:
Resubmit of #676 

>The current pattern for stdlib::compat::ipv4 is incorrect, and
> will return true for any four numbers separated with periods.
> This commit improves the regex to validate that the IP octets
> are between 1 and 255.

Also, ruby's `ipaddr` does not like 0-padded octets like `010`


DavidS:
Superseded by #680

PR #680, opened 2016-10-24
(MODULES-3980) Fix ipv4 regex validator

DavidS:
This also updates all ipv4 tests to use the same test data for better
comparability. Closes #676, #679

Fix-Originally-By: Nate Potter <nathaniel.potter@intel.com>


DavidS:
Like everything else, this is currently blocked on #678

PR #681, opened 2016-10-24
Remove leading spaces

cacack:
This corrects puppet-linting error.


cacack:
Looks like blocked on #678.

PR #682, opened 2016-10-25
 This is to pin ruby version to parallel_tests

pmcmaw:
This updates the Gemfile from the modulesync to pin Ruby version to parallel_tests. This also allows for deprecations in the .sync.yml.


DavidS:
The cells not yet green have already started up beaker. Since this change is only affecting gem installation in bundler, I'll merge this now, to unblock the other PRs waiting on this.

PR #685, opened 2016-11-07
Call site output for deprecation warnings

HelenCampbell:


tphoney:
msg should be named something more obvious, maybe message_and_stracktrace

tphoney:
or output_message

lib/puppet/functions/deprecation.rb
```diff
@@ -8,15 +8,17 @@
   end
 
   def deprecation(key, message)
+    stacktrace = Puppet::Pops::PuppetStack.stacktrace()
+    output_message = message + ' - File and Line: ' + stacktrace[0].inspect
```
DavidS: re-use `message` here, saves you editing below.

Also, please format the message as `"${message} at ${stacktrace[0][0]}:${stacktrace[0]:[1]}"` to match common conventions on error reporting.

lib/puppet/functions/deprecation.rb
```diff
@@ -8,15 +8,19 @@
   end
 
   def deprecation(key, message)
+    stacktrace = Puppet::Pops::PuppetStack.stacktrace()
```
alexjfisher: @DavidS I'm sorry I keep doing this to you, but ...

`Evaluation Error: Error while evaluating a Function Call, uninitialized constant Puppet::Pops::PuppetStack`
https://travis-ci.org/voxpupuli/puppet-gluster/jobs/177770318#L917

That's puppet 3 with future parser enabled.

bmjen:
@DavidS @HelenCampbell @alexjfisher I'm going to revert this PR for now.

bmjen:
The `Puppet::Pops::PuppetStack` module has only been available since Puppet 4.6.0.
https://github.com/puppetlabs/puppet/commit/2211c646921f85a31611f018cd8bfaab76692c98

lib/puppet/functions/deprecation.rb
DavidS: ugh, thanks for catching that. I know how that happened: https://github.com/puppetlabs/puppetlabs-stdlib/blob/master/spec/functions/deprecation_spec.rb#L3 is a little too optimistic...

:-(

We'll retry tomorrow.

PR #687, opened 2016-11-08
(MODULES-3829) Use .dup to duplicate classes for modification.

MG2R:
This function otherwise fails during `puppet preview` on Puppet 3.8.X systems.


MG2R:
Failing check seems to be failing in an unrelated function.  How do I fix this? Do I fix this?

DavidS:
@MG2R no worries, that's an unrelated failure that we're working on.

PR #690, opened 2016-11-15
MODULES-4008: clarify deprecation language

jbondpdx:
Updates to clarify (I hope) deprecation language and note which functions are deprecated.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -712,45 +712,67 @@ See the [`assert_type()`](https://docs.puppetlabs.com/references/latest/function
 
 #### `is_absolute_path`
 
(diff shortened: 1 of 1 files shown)


DavidS:
@jbondpdx I like the language changes. And, now that I've read the note you added to the functions, I do realize that your original suggestion of pointing to validate_legacy as a way to address these deprecations is a very good one. Apologies.

Finally, all the is_* functions (except is_a()) from https://github.com/puppetlabs/puppetlabs-stdlib/tree/master/lib/puppet/functions are also deprecated.

jbondpdx:
@DavidS Awesome, thank you for looking at this. I've updated the PR if you'd give it another look and merge, if appropriate.

PR #691, opened 2016-11-21
Fixing broken link to #validate_legacy docs

idnorton:
Doc link to the validate_legacy section is missing an underscore which is a little annoying.


DavidS:
Thanks @idnorton !

cc @jbondpdx

PR #693, opened 2016-11-23
(MODULES-3393) Deprecation - Use puppet stacktrace if available

HelenCampbell:


HelenCampbell:
A previous PR (https://github.com/puppetlabs/puppetlabs-stdlib/pull/685) was raised on this issue, however once it was merged it was discovered that Puppet 3 with future parser enabled was calling the Puppet 4 version of the deprecation function. The Puppet stacktrace is not available until Puppet 4.6, so this was breaking existing setups. The solution was to check is the stacktrace was defined, and if it was to use it as part of the message output.

DavidS:
Please put the info into the commit message, so it is also available offline.

lib/puppet/functions/deprecation.rb
```diff
@@ -8,6 +8,14 @@
   end
 
   def deprecation(key, message)
+    if defined? Puppet::Pops::PuppetStack.stacktrace()
+      stacktrace = Puppet::Pops::PuppetStack.stacktrace()
+      file = stacktrace[0]
+      line = stacktrace[1]
+      output_message = "#{message} at #{file}:#{line}"
+    else
+      output_message = message
```
bmjen: Don't you need to use `output_message` in the failure msg and deprecation warning msg below?

DavidS: true. as I said on the original PR, re-using the `message` var would obviate the need to modify the code below, and - in this case - also remove the need fo the `else`.

PR #695, opened 2016-11-30
Indicate that the type function is preferred

npwalker:
Prior to this commit, users coming to the type_of function would not realize that the type function in puppet does the same thing and is preferred over type_of. 

After this commit, we have a comment indicating the above.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -1190,6 +1190,8 @@ Returns a string description of the type when passed a value. Type can be a stri
 
 #### `type_of`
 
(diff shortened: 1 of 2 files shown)


npwalker:
This is also related to this documentation ticket.  

https://tickets.puppetlabs.com/browse/DOCUMENT-617

DavidS:
Please also add that info to the README - we are not (YET!) using puppet-strings.

npwalker:
Updated to indicate the same in the README

DavidS:
Thanks! We should also probably deprecate this at some point, y/n?

PR #698, opened 2016-12-05
Release prep for 4.13.2

HelenCampbell:


tphoney:
this should be 4.14.0 if features were added

PR #700, opened 2016-12-07
(MODULES-4188) Add UUID generation function

petems:
* Generates UUID based on a given string and 
optional seed

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -578,6 +578,15 @@ fqdn_rotate([1, 2, 3], 'custom seed')
 
 *Type*: rvalue.
 
+#### `fqdn_uuid`
+
+Returns a rfc4122 valid version 5 UUID based on an FQDN string under the DNS namespace
+
+  * fqdn_uuid('puppetlabs.com') returns '9c70320f-6815-5fc5-ab0f-debe68bf764c'
+  * fqdn_uuid('google.com') returns '64ee70a4-8cc1-5d25-abf2-dea6c79a09c8'
+
+*Type*: rvalue.
+
 #### `get_module_path`
 
 Returns the absolute path of the specified module for the current environment.
--- a/lib/puppet/parser/functions/fqdn_uuid.rb
+++ b/lib/puppet/parser/functions/fqdn_uuid.rb
@@ -1,0 +1,92 @@
+require 'digest/sha1'
+
+module Puppet::Parser::Functions
+  newfunction(:fqdn_uuid, :type => :rvalue, :doc => <<-END) do |args|
+
+    Creates a UUID based on a given string, assumed to be the FQDN
+
+    For example, to generate a UUID based on the FQDN of a system:
+
+    Usage:
+
+      $uuid = fqdn_uuid($::fqdn)
+
+    The generated UUID will be the same for the given hostname
+
+    The resulting UUID is returned on the form:
+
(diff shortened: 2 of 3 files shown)


lib/puppet/parser/functions/uuid_generate.rb
```diff
@@ -0,0 +1,79 @@
...
+
+    Creates a UUID based on a given string and an optional seed value.
+
+    For example, to generate a UUID based on the FQDN of a system:
+
+    Usage:
+
+      $uuid = uuid_generate($::fqdn)
+
+    The seed is optional. If given it must be a string.
```
hunner: It should mention if the seed should be numeric, hex, or an arbitrary string.

lib/puppet/parser/functions/uuid_generate.rb
```diff
@@ -0,0 +1,79 @@
...
+
+    END
+
+    if args.length == 0
+      raise(ArgumentError, "uuid_generate: No arguments given")
+    elsif args.length == 1
+      seed = nil
+      fqdn = args[0]
+    elsif args.length == 2
+      fqdn = args[0]
```
hunner: Technically it doesn't have to be an FQDN so it should be called `input_string` or something.

lib/puppet/parser/functions/uuid_generate.rb
```diff
@@ -0,0 +1,79 @@
...
+      seed = args[1]
+      unless seed.is_a?(String)
+        raise(Puppet::ParseError, 'uuid_generate(): seed argument must be a string')
+      end
+    else
+      raise(ArgumentError, "uuid_generate: Too many arguments given (#{args.length})")
+    end
+
+    # The UUID for our namespace (6ba7b810-9dad-11d1-80b4-00c04fd430c8)
+    ns_uuid = "\x0b\x7a\x81\xff\xdb\x8d\x42\xfe\x8d\x9f\x76\x8e\xa5\xb8\xed\x1a"
```
hunner: This UUID name space ID is only for prefixing FQDNs and as this is an "uuid_generate puppet function arbitrary string" UUID, we should create a new namespace specifically for this function (I imagine anything will do).

As per RFC-4122:
> Allocate a UUID to use as a "name space ID" for all UUIDs generated from names in that name space; see Appendix C for some pre-defined values.

lib/puppet/parser/functions/uuid_generate.rb
```diff
@@ -0,0 +1,79 @@
...
+
+    hash_input = ns_uuid + fqdn
+    unless seed.nil?
+      hash_input += "\x00" + seed
```
hunner: If the seed can be any string, why prefix with `'\x00'`?

lib/puppet/parser/functions/uuid_generate.rb
petems: Yeah, this was a holdover from the older FQDN specific code

lib/puppet/parser/functions/uuid_generate.rb
petems: TBH I'm not sure about this bit, I was leveraging an existing bit of code

petems: What would need to change here?

PR #701, opened 2016-12-13
WIP use rsync instead of scp for copying the module puppethack

tphoney:


tphoney:
there is always something :)

PR #707, opened 2017-01-08
Also remove values that are nil

mattkenn4545:


wilson208:
@mattkenn4545 I am thinking that this change may be slightly misleading, nil != Undef and the naming of this function is `delete_undef_values `. Perhaps this should be broken out to its own function `delete_nil_values` or perhaps add an additional parameter `delete_nil_values` or similar to `delete_undef_values` which defaults to false to ensure we do not modify current behaviour.

mattkenn4545:
However there is no nil value in the DSL.

hlindberg:
This is the same as filtering for match on `Undef` in puppet.
```
function without_undef(Collection $val) { filter |$k, $v| { $v !~ Undef and $k !~ Undef}) }
```
The matching with `Undef` will handle both Ruby `nil` and Ruby `:undef`. It checks if key is of Undef type as well as the value. It works for both arrays and hashes (i.e. a Collection)

I think this function is better implemented as a puppet language function as it removes the concern of having to deal with the Ruby symbol `:undef`.

PR #714, opened 2017-02-01
(FM-6019) - i18N tests for Spike

pmcmaw:
WIP - Tests for internationalization. Sending out for feedback before I continue to write tests for all of the stdlib functions.  I plan on writing only unit tests as I am only testing that the function is compatible and can handle UTF8 characters etc.

This PR covers 2 functions:
1) ensure_resource
2) any2array


DavidS:
Please add a little bit of verbiage around what we control for by running this on both unit and acceptance stages?

tphoney:
+1

DavidS:
#720 has a workaround for the test failure

PR #716, opened 2017-02-03
(Modules 4377) Causes ensure_packages to accept concurrent declarations with ensure => 'present' and 'installed'

EmersonPrado:
If user declares ensure_packages(<Package>, { ensure => 'present' }) and ensure_packages(<Package>, { ensure => 'installed' }) concurrently, funcion fails at different 'ensure' values. This change causes function to change 'installed' to 'present', avoiding this error.
Maybe similar treatment is needed when arguments[0] is a hash, but I'm an absolute beginner in Ruby.

### DIFF
--- a/lib/puppet/parser/functions/ensure_packages.rb
+++ b/lib/puppet/parser/functions/ensure_packages.rb
@@ -20,6 +20,9 @@ third argument to the ensure_resource() function.
     if arguments[0].is_a?(Hash)
       if arguments[1]
         defaults = { 'ensure' => 'present' }.merge(arguments[1])
+        if defaults['ensure'] == 'installed'
+          defaults['ensure'] = 'present'
+        end
       else
         defaults = { 'ensure' => 'present' }
       end
@@ -31,6 +34,9 @@ third argument to the ensure_resource() function.
 
       if arguments[1]
         defaults = { 'ensure' => 'present' }.merge(arguments[1])
+        if defaults['ensure'] == 'installed'
+          defaults['ensure'] = 'present'
+        end
       else
         defaults = { 'ensure' => 'present' }
       end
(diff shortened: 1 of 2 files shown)


EmersonPrado:
I would appreciate some help regarding the failed checks. They don't seem to be related to my commits, but I'm not sure.
Both point to line 3 of spec/acceptance/zip_spec.rb, which is `require 'puppet'`, and complain that 'Hiera is not a class (TypeError)'.

I also found 2 excerpts that called my attention regarding the errors:
#### Gemfile
```
# Only explicitly specify Facter/Hiera if a version has been specified.
# Otherwise it can lead to strange bundler behavior...
gem 'hiera', *location_for(ENV['HIERA_GEM_VERSION']) if ENV['HIERA_GEM_VERSION']
```
#### lib/puppet/parser/functions/ensure_resources.rb
```
From Hiera Backend:
...
Call:
ensure_resources('user', hiera_hash('userlist'), {'ensure' => 'present'})
```
Does this tell anything?

Thanks a lot.

EmersonPrado:
Still regarding the test error, there is a question in Ask Puppet reporting the exact same error, and also related to the exact same test:
https://ask.puppet.com/question/28542/hiera-is-not-a-class-typeerror/
The only answer so far mentions auto-generated Rake tests as a possible cause for the error. Maybe this should go for a new ticket.

EmersonPrado:
Seems like a bug in Beaker:
https://tickets.puppetlabs.com/browse/BKR-969

esalberg:
+1 - We have run into this too.

EmersonPrado:
Seems like commit 2c6c2ff62db3cb9ee3b8bf755f66da4918783d28 fixed the errors.

tphoney:
Thanks for the PR @EmersonPrado good work !!!

PR #717, opened 2017-02-04
Add new file_line option append_on_no_match

ripclawffb:
Currently, there doesn't exist a way to leave a file untouched if no regex matches are found. 

I have a use case where I'm searching for a specific regex in multiple files and I only want to update the files that contain the regex pattern.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,6 +115,18 @@ file_line { 'bashrc_proxy':
 
 In the example above, `match` looks for a line beginning with 'export' followed by 'HTTP_PROXY' and replaces it with the value in line.
 
+Match Example:
+
+    file_line { 'bashrc_proxy':
+      ensure             => present,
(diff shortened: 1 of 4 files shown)


tphoney:
@ripclawffb There have been a number of changes in the file_line can you see if this PR is still necessary. 

More importantly apologies for the slowness to getting to this PR

ripclawffb:
@tphoney I reviewed the changes for file_line and there still doesn't seem to be an option to not add a line if a match is not found in a file.

I also went ahead and rebased as requested.

eputnam:
rebased again @tphoney, looks ok pending travis. @ripclawffb take a look if you'd like to make sure it still looks ok to you.

ripclawffb:
@eputnam thanks! looks good to me.

PR #719, opened 2017-02-04
Allow test module metadata.json to be read

domcleal:
Since puppetlabs/puppet@f2e8e66, the test module's metadata.json is also
read causing expectation failures. Like 5c51463, permit it to be read by
Puppet.


DavidS:
#720 has a workaround for the test failure

PR #721, opened 2017-02-21
MODULES-4380 Tests for ensure_resource

pmcmaw:
This is adding tests to ensure that the ensure_resource function can process UFT8 strings. 
Tests will be added for each function as a separate PR to maintain traceability.


pmcmaw:
Closed as this is already in from a previous PR: https://github.com/puppetlabs/puppetlabs-stdlib/pull/714

PR #725, opened 2017-02-22
(FM-6054) - Unit tests for low effort functions

pmcmaw:
This is unit tests for the low effort function tests for i18n compatibility. 

This includes tests for the following functions:
delete_undef_values
difference
flatten
fqdn_rand_string
fqdn_rotate
grep
hash
intersection
join
join_keys_to_values
lstrip
member
merge
pick_default
prefix
reject
reverse
rstrip
strip
suffix
union
unique

### DIFF
--- a/spec/functions/join_keys_to_values_spec.rb
+++ b/spec/functions/join_keys_to_values_spec.rb
@@ -11,6 +11,12 @@ describe 'join_keys_to_values' do
   it { is_expected.to run.with_params({}, ':').and_return([]) }
   it { is_expected.to run.with_params({ 'key' => 'value' }, '').and_return(['keyvalue']) }
   it { is_expected.to run.with_params({ 'key' => 'value' }, ':').and_return(['key:value']) }
+
+  context 'should run with UTF8 and double byte characters' do 
+    it { is_expected.to run.with_params({ 'ҝẽγ' => '√ạĺűē' }, ':').and_return(['ҝẽγ:√ạĺűē']) }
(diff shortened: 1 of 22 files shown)


spec/functions/join_keys_to_values_spec.rb
```diff
@@ -10,7 +10,12 @@
   it { is_expected.to run.with_params({}, '').and_return([]) }
   it { is_expected.to run.with_params({}, ':').and_return([]) }
   it { is_expected.to run.with_params({ 'key' => 'value' }, '').and_return(['keyvalue']) }
+  it { is_expected.to run.with_params({ 'ҝẽγ' => '√ạĺűē' }, '').and_return(['ҝẽγ√ạĺűē']) }
   it { is_expected.to run.with_params({ 'key' => 'value' }, ':').and_return(['key:value']) }
+  it { is_expected.to run.with_params({ 'ҝẽγ' => '√ạĺűē' }, ':').and_return(['ҝẽγ:√ạĺűē']) }
+  it { is_expected.to run.with_params({ 'ҝẽγ' => '√ạĺűē' }, '万').and_return(['ҝẽγ万√ạĺűē']) }
+
```
tphoney: i cant see the difference between these added tests, why 3 of them ?

tphoney: otherwise looks good

pmcmaw: The first test is testing the actual functionality of the function, the second test is making sure that the function can cope with UTF8 strings, then the final test is ensuring that the double byte character can be used.

pmcmaw: I have now updated the context for the above tests to make it more clear what the test is doing and its purpose.

PR #726, opened 2017-02-27
(#FM-6068) allow file encoding to be specified

GeoffWilliams:
Add a new parameter `encoding` to allow non UTF-8 files to specify a file encoding.  This prevents receiving the error message "invalid byte sequence in UTF-8" when special characters that are not UTF-8 encoded appear in the input stream, such as the copyright symbol.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -118,6 +118,22 @@ In this code example, `match` looks for a line beginning with export
 followed by HTTP_PROXY and delete it.  If multiple lines match, an
 error will be raised unless the `multiple => true` parameter is set.
 
+Encoding example:
+
+    file_line { "XScreenSaver":
+      ensure   => present,
+      path     => '/root/XScreenSaver'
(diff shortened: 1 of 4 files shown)


tphoney:
Thanks for the PR, would it be possible to add some tests around this new functionality, and add a little bit of text into the readme. This looks really useful 👍

tphoney:
@pmcmaw is this something that you could look over

pmcmaw:
@tphoney looks good to me, happy for merge when tests and docs are in.

GeoffWilliams:
@tphoney and @pmcmaw - tests added and documentation copy pasted from source code to readme.  I'm not sure what I can test for other then the ability to accept the parameter so I just tested that - thanks

pmcmaw:
@GeoffWilliams Thank you for your contribution, it is much appreciated. 
:-)

PR #730, opened 2017-03-02
(MODULES-4294) WIP adds i18n decorator for ruby

eputnam:
DEPENDS ON #729 

All appropriate strings in lib/facter decorated with the _() function
All appropriate strings in lib/puppet/type and lib/puppet/provider wrapped in _() function
All appropriate strings in lib/puppet/parser/functions wrapped with _() function
All appropriate strings in lib/puppet/functions wrapped with _() function

### DIFF
--- a/lib/facter/facter_dot_d.rb
+++ b/lib/facter/facter_dot_d.rb
@@ -49,7 +49,7 @@ class Facter::Util::DotD
       end
     end
   rescue StandardError => e
-    Facter.warn("Failed to handle #{file} as text facts: #{e.class}: #{e}")
+    Facter.warn(_("Failed to handle #{file} as text facts: #{e.class}: #{e}"))
   end
 
   def json_parser(file)
@@ -66,7 +66,7 @@ class Facter::Util::DotD
       end
     end
   rescue StandardError => e
-    Facter.warn("Failed to handle #{file} as json facts: #{e.class}: #{e}")
+    Facter.warn(_("Failed to handle #{file} as json facts: #{e.class}: #{e}"))
   end
 
   def yaml_parser(file)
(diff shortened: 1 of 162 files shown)


tphoney:
looks good,  how did you change so much ? through an ide or sed/awk ?

pmcmaw:
@tphoney The changes were carried out using the following script:
https://github.com/eputnam/string-wrapper

lib/facter/facter_dot_d.rb
```diff
@@ -49,7 +49,7 @@ def txt_parser(file)
       end
     end
   rescue StandardError => e
-    Facter.warn("Failed to handle #{file} as text facts: #{e.class}: #{e}")
+    Facter.warn(_("Failed to handle #{file} as text facts: #{e.class}: #{e}"))
```
DavidS: This will not create a translatable string, as it interpolates `file` and `e.class` before looking up the translation.

eputnam: is this true for all the Facter messaging?

DavidS: This is true for all strings.

DavidS: See https://github.com/puppetlabs/gettext-setup-gem#interpolation for background.

eputnam: i've been told interpolated strings are a non-issue which is why i thought i could get away with it. i'll be looking into this today.

glennsarti:
PR #729 has been merged. Needs a rebase

tphoney:
@eputnam is this still relevant ? can it be closed

eputnam:
@TP this is just all strings decorated using the scripts we have currently, but if it's going to get us caught doing something we don't want to do then i guess it's not worth keeping...

eputnam:
failures are expected on Puppet 4.2.3 which does not know how to handle the _() function

PR #735, opened 2017-03-09
Should only try to apply the resource if it not defined

elmobp:
As there is a number of modules that use ensure_package without wrapping a defined around them many modules have conflicts. This will help work around them by default, Puppet will not let you define a resource twice so it should be done here as well


wilson208:
LGTM 👍

PR #736, opened 2017-03-13
Addition of new length function

HelenCampbell:


lib/puppet/functions/length.rb
```diff
@@ -0,0 +1,16 @@
+#A function to eventually replace the old size() function for stdlib - The original size function did not handle Puppets new type capabilities, so this function is a Puppet 4 compatible solution. 
+Puppet::Functions.create_function(:length) do
+	dispatch :length do
+		param 'Any', :value
```
DavidS: Using the proper type declaration here, using `Variant`, will avoid the need for the `raise` in L12 below.

Using the proper type declarations here, using different dispatches for Strings, and others, will avoid the need for the `if` chain in L7.

DavidS:
Title needs changing, and still waiting for travis&appveyor

PR #737, opened 2017-03-14
(MODULES-4528) Use versioncmp to check Puppet version for 4.10.x compat

domcleal:
`Puppet.version.to_f` on Puppet 4.10.0 will evaluate to `4.1`, causing
test and behavioural changes when conditionals check that the version is
equal or greater than versions such as `4.3`.

Version comparisons that are vulnerable to this have been changed to use
Puppet's versioncmp implementation, while most others only check for
for major version boundaries which is safe.


wilson208:
👍 LGTM, Thanks!

PR #738, opened 2017-03-14
(MODULES-6051) Add UTF-8 support to capitalize

HAIL9000:
Prior to this commit, the capitalize function did not support
UTF-8 characters. Now it has been updated using the Ruby Unicode
library so that it properly parses and capitalizes strings containing
UTF-8 characters.


HAIL9000:
Right now this PR only fixes one of the several functions in stdlib that have this issue. If this fix looks good, then I'll go back and add fixes for the remaining functions.

HAIL9000:
Closing in favor of #741

PR #745, opened 2017-03-22
[MODULES-4528] Replace Puppet.version.to_f version comparison from spec_helper.rb

wilson208:


igalic:
uhm… what does the title / commit message of this PR have to do with the changes?

wilson208:
@igalic I know this is somewhat confusing but what happened is that as a result of MODULES-4528 I have needed to do a modulesync run to replace Puppet.version.to_f in the spec/spec_helper.rb file. However when I went to do the msync run it turns out that there are pending changes put into the modulesync configs that have not yet been modulesync'd out to the modules. Rather than doing this in 2 different runs across all our modules I thought its better to combine both changes.

PR #747, opened 2017-03-24
(MODULES-4322) pre-loc edit on stdlib README

jbondpdx:
Hello good module people!

Please carefully review my changes here. I've noted all questions or places I'm in doubt about with TODO, so you can search for that to find those spots easily.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -2,7 +2,6 @@
 
 #### Table of Contents
 
-1. [Overview](#overview)
 2. [Module Description - What the module does and why it is useful](#module-description)
 3. [Setup - The basics of getting started with stdlib](#setup)
 4. [Usage - Configuration options and additional functionality](#usage)
@@ -10,36 +9,31 @@
 5. [Limitations - OS compatibility, etc.](#limitations)
 6. [Development - Guide for contributing to the module](#development)
 
-## Overview
-
-Adds a standard library of resources for Puppet modules.
 
 ## Module Description
 
-This module provides a standard library of resources for the development of Puppet modules. Puppet modules make heavy use of this standard library. The stdlib module adds the following resources to Puppet:
+This module provides a standard library of resources for Puppet modules. Puppet modules make heavy use of this standard library. The stdlib module adds the following resources to Puppet:
 
  * Stages
  * Facts
  * Functions
- * Defined resource types
- * Data Types
+ * Defined types
+ * Data types
  * Providers
 
 > *Note:* As of version 3.7, Puppet Enterprise no longer includes the stdlib module. If you're running Puppet Enterprise, you should install the most recent release of stdlib for compatibility with Puppet modules.
 
 ## Setup
 
-Installing the stdlib module adds the functions, facts, and resources of this standard library to Puppet.
+[Install](https://docs.puppet.com/puppet/latest/modules_installing.html) the stdlib module to add the functions, facts, and resources of this standard library to Puppet.
+
+If you are authoring a module that depends on stdlib, be sure to [specify dependencies](https://docs.puppet.com/puppet/latest/modules_metadata.html#specifying-dependencies) in your metadata.json.
 
 ## Usage
 
-After you've installed stdlib, all of its functions, facts, and resources are available for module use or development.
+Most of stdlib's features are automatically loaded by Puppet. To use standardized run stages in Puppet, declare this class in your manifest with `include stdlib`.
 
-If you want to use a standardized set of run stages for Puppet, `include stdlib` in your manifest.
-
-* `stdlib`: Most of stdlib's features are automatically loaded by Puppet. To use standardized run stages in Puppet, declare this class in your manifest with `include stdlib`.
-
-  When declared, stdlib declares all other classes in the module. The only other class currently included in the module is `stdlib::stages`.
+When declared, stdlib declares all other classes in the module. The only other class currently included in the module is `stdlib::stages`.
 
 The `stdlib::stages` class declares various run stages for deploying infrastructure, language runtimes, and application layers. The high level stages are (in order):
 
@@ -52,87 +46,96 @@ The `stdlib::stages` class declares various run stages for deploying infrastruct
   * deploy_app
   * deploy
 
-  Sample usage:
+Sample usage:
 
-  ~~~
-  node default {
-    include stdlib
-    class { java: stage => 'runtime' }
-  }
-  ~~~
+```puppet
+node default {
+  include stdlib
+  class { java: stage => 'runtime' }
+}
+```
 
 ## Reference
 
+* [Public classes][]
+* [Private classes][]
+* [Defined types][]
+* [Data types][]
+* [Facts][]
+* [Functions][]
+
 ### Classes
 
-#### Public Classes
+#### Public classes
 
-  The stdlib class has no parameters.
+The `stdlib` class has no parameters.
 
-#### Private Classes
+#### Private classes
 
-* `stdlib::stages`: Manages a standard set of run stages for Puppet. It is managed by the stdlib class and should not be declared independently.
+* `stdlib::stages`: Manages a standard set of run stages for Puppet.
 
-### Resource Types
+### Defined types
 
 #### `file_line`
 
-Ensures that a given line is contained within a file. The implementation matches the full line, including whitespace at the beginning and end. If the line is not contained in the given file, Puppet appends the line to the end of the file to ensure the desired state.  Multiple resources can be declared to manage multiple lines in the same file.
+Ensures that a given line is contained within a file. The implementation matches the full line, including whitespace at the beginning and end. If the line is not contained in the given file, Puppet appends the line to the end of the file to ensure the desired state. Multiple resources can be declared to manage multiple lines in the same file.
 
 Example:
 
-    file_line { 'sudo_rule':
-      path => '/etc/sudoers',
-      line => '%sudo ALL=(ALL) ALL',
-    }
+```puppet
+file_line { 'sudo_rule':
+  path => '/etc/sudoers',
+  line => '%sudo ALL=(ALL) ALL',
+}
 
-    file_line { 'sudo_rule_nopw':
-      path => '/etc/sudoers',
-      line => '%sudonopw ALL=(ALL) NOPASSWD: ALL',
-    }
+file_line { 'sudo_rule_nopw':
+  path => '/etc/sudoers',
+  line => '%sudonopw ALL=(ALL) NOPASSWD: ALL',
+}
+```
 
-In this example, Puppet ensures that both of the specified lines are contained in the file `/etc/sudoers`.
+In the example above, Puppet ensures that both of the specified lines are contained in the file `/etc/sudoers`.
 
 Match Example:
 
-    file_line { 'bashrc_proxy':
-      ensure => present,
-      path   => '/etc/bashrc',
-      line   => 'export HTTP_PROXY=http://squid.puppetlabs.vm:3128',
-      match  => '^export\ HTTP_PROXY\=',
-    }
+```puppet
+file_line { 'bashrc_proxy':
+  ensure => present,
+  path   => '/etc/bashrc',
+  line   => 'export HTTP_PROXY=http://squid.puppetlabs.vm:3128',
+  match  => '^export\ HTTP_PROXY\=',
+}
+```
 
-In this code example, `match` looks for a line beginning with export followed by HTTP_PROXY and replaces it with the value in line.
+In the example above, `match` looks for a line beginning with 'export' followed by 'HTTP_PROXY' and replaces it with the value in line.
 
-Match Example With `ensure => absent`:
+Match Example with `ensure => absent`:
 
-    file_line { 'bashrc_proxy':
-      ensure            => absent,
-      path              => '/etc/bashrc',
-      line              => 'export HTTP_PROXY=http://squid.puppetlabs.vm:3128',
-      match             => '^export\ HTTP_PROXY\=',
-      match_for_absence => true,
-    }
+```puppet
+file_line { 'bashrc_proxy':
+  ensure            => absent,
+  path              => '/etc/bashrc',
+  line              => 'export HTTP_PROXY=http://squid.puppetlabs.vm:3128',
+  match             => '^export\ HTTP_PROXY\=',
+  match_for_absence => true,
+}
+```
 
-In this code example, `match` looks for a line beginning with export
-followed by HTTP_PROXY and delete it.  If multiple lines match, an
-error will be raised unless the `multiple => true` parameter is set.
+In the example above, `match` looks for a line beginning with 'export' followed by 'HTTP_PROXY' and deletes it. If multiple lines match, an error is raised, unless the `multiple => true` parameter is set.
 
 Encoding example:
 
-    file_line { "XScreenSaver":
-      ensure   => present,
-      path     => '/root/XScreenSaver'
-      line     => "*lock: 10:00:00",
-      match    => '^*lock:',
-      encoding => "iso-8859-1",
-    }
+```puppet
+file_line { "XScreenSaver":
+  ensure   => present,
+  path     => '/root/XScreenSaver'
+  line     => "*lock: 10:00:00",
+  match    => '^*lock:',
+  encoding => "iso-8859-1",
+}
+```
 
-Files with special characters that are not valid UTF-8 will give the 
-error message "invalid byte sequence in UTF-8".  In this case, determine
-the correct file encoding and specify the correct encoding using the
-encoding attribute, the value of which needs to be a valid Ruby character
-encoding.
+Files with special characters that are not valid UTF-8 give the error message "Invalid byte sequence in UTF-8". In this case, determine the correct file encoding and specify it with the `encoding` attribute.
 
 **Autorequires:** If Puppet is managing the file that contains the line being managed, the `file_line` resource autorequires that file.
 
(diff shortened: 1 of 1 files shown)


README.markdown
```diff
@@ -940,92 +1316,110 @@ Merges two or more hashes together and returns the resulting hash.
...
-* Run `catalog` to see the contents currently compiling catalog
-* Run `cd catalog` and `ls` to see catalog methods and instance variables
-* Run `@resource_table` to see the current catalog resource table
+In a pry session, useful commands include:
+
+* Run `catalog` to see the contents currently compiling catalog.
+* Run `cd catalog` and `ls` to see catalog methods and instance variables.
+* Run `@resource_table` to see the current catalog resource table.
 
 #### `assert_private`
```
DavidS: This is not in alphabetic order here.

README.markdown
```diff
@@ -940,92 +1316,110 @@ Merges two or more hashes together and returns the resulting hash.
...
 
-Sets the current class or definition as private. Calling the class or definition from outside the current module will fail.
+Sets the current class or definition as private. Calling the class or definition from outside the current module fails.
```
DavidS: "defined type" instead of "definition"

README.markdown
```diff
@@ -1037,198 +1431,252 @@ The second argument to this function is which type of hash to use. It will be co
...
-    * `%x`: Preferred representation for the date alone, no time
-    * `%X`: Preferred representation for the time alone, no date
-    * `%y`: Year without a century (00..99)
-    * `%Y`: Year with century
-    * `%z`: Time zone as hour offset from UTC (e.g. +0900)
-    * `%Z`: Time zone name
-    * `%%`: Literal '%' character
+Returns formatted time.
+
+For example, `strftime("%s")` returns the time since Unix epoch, and `strftime("%Y-%m-%d")` returns the date.
```
DavidS: link to upstream documentation for full details: https://ruby-doc.org/core-2.1.9/Time.html#method-i-strftime

README.markdown
```diff
@@ -1037,198 +1431,252 @@ The second argument to this function is which type of hash to use. It will be co
...
+For example, "4 kB" becomes "4096". 
+
+Arguments: A single string.
 
 *Type*: rvalue.
 
-DEPRECATED: replaced by `dig()`.
+#### `try_get_value`
+
+**DEPRECATED:** replaced by `dig()`.
```
DavidS: lolsob

README.markdown
```diff
@@ -886,51 +1252,61 @@ Loads a JSON file containing an array, string, or hash, and returns the data in
...
-  $metadata = load_module_metadata('archive')
-  notify { $metadata['author']: }
-  ~~~
+```puppet
+$metadata = load_module_metadata('archive')
+notify { $metadata['author']: }
+```
 
-If you do not want to fail the catalog compilation when a module's metadata file is absent:
+When a module's metadata file is absent, the catalog compilation fails. To stop this failure:
```
jbondpdx: Yeah, I almost went with avoid in the first place.

jbondpdx:
I've updated the PR based on @DavidS 's comments; will squash commits before the real PR.

README.markdown
```diff
@@ -52,167 +46,374 @@ The `stdlib::stages` class declares various run stages for deploying infrastruct
...
+
+  Specifies the line after which Puppet adds any new lines using a regular expression. (Existing lines are added in place.)
+
+  Values: String containing a regex.
+  
+  Default value: `undef`.
+  
+* `encoding`
+  
+  Specifies the correct file encoding. TODO: verify information, default value?
```
HelenCampbell: https://github.com/puppetlabs/puppetlabs-stdlib/blob/46e3a2349e9eef4001db8be9581fae9dfbccaf14/lib/puppet/type/file_line.rb#L128
Default value is 'UTF-8'.

README.markdown
```diff
@@ -52,167 +46,374 @@ The `stdlib::stages` class declares various run stages for deploying infrastruct
...
+
+Returns the value of the Puppet vardir setting for the node running Puppet or Puppet agent.
+
+#### `puppet_environmentpath`
+
+Returns the value of the Puppet environment path settings for the node running Puppet or Puppet agent.
+
+#### `puppet_server`
+
+TODO: What does this return? It was unclear to me.
```
HelenCampbell: The changelog describes it as: "Addition of puppet_server fact to return agents server."

README.markdown
```diff
@@ -52,167 +46,374 @@ The `stdlib::stages` class declares various run stages for deploying infrastruct
...
+
+#### `root_home`
+
+Determines the root home directory.
+
+TODO: The comments say: "This varies on PE supported platforms and may be reconfigured by the end user." How would they reconfigure it? How does it vary? What information will the user want here? (thank you!)
```
HelenCampbell: The following is what I can tell from this spec test: https://github.com/puppetlabs/puppetlabs-stdlib/blob/dce8d7b194e3f6d9bb24e076c6e835aeb671099a/spec/unit/facter/root_home_spec.rb

What is expected from each os/ how they vary:
Solaris: "/" 
Linux: "/root"
Windows:  nil
Macosx: "/var/root"
aix: "/root"

Because it is a facter fact, we can override (reconfigure) this like so: https://www.puppetcookbook.com/posts/override-a-facter-fact.html
I'm not sure if/ how much you'd want to include about this information as I would assume that anyone using Facter facts would know that they can be overridden.

README.markdown
```diff
@@ -52,167 +46,374 @@ The `stdlib::stages` class declares various run stages for deploying infrastruct
...
+Returns the default provider Puppet uses to manage services on this system
 
 ### Functions
 
 #### `abs`
 
-Returns the absolute value of a number; for example, '-34.56' becomes '34.56'. Takes a single integer and float value as an argument. *Type*: rvalue.
+Returns the absolute value of a number. For example, '-34.56' becomes '34.56'.
+
+Argument: A single integer and a float value as an argument.
```
HelenCampbell: This can be misleading - checked the function and its in there too as a comment which we should change. But it should be more: Accepts a single argument of either an integer or float value.

README.markdown
```diff
@@ -247,131 +449,197 @@ Returns the `basename` of a path (optionally stripping an extension). For exampl
...
+Arguments: Boolean.
+
+*Type*: rvalue.
 
 #### `camelcase`
 
-Converts the case of a string or all strings in an array to camel case. *Type*: rvalue.
+Converts the case of a string or all strings in an array to mixed case (both uppercase and lowercase).
+
+Arguments: TODO how do I use this function?
```
HelenCampbell: Since camel case is (as far as I know) a technical term I would leave it in here?
Requires inputs of either array or string. Returns the same type of argument as it received but in CamelCase form.

README.markdown
```diff
@@ -247,131 +449,197 @@ Returns the `basename` of a path (optionally stripping an extension). For exampl
...
+* `delete(['ab', 'b'], 'b')` returns ['ab'].
 
 *Type*: rvalue.
 
 #### `delete_at`
 
-Deletes a determined indexed value from an array. For example, `delete_at(['a','b','c'], 1)` returns ['a','c']. *Type*: rvalue.
+Deletes a determined indexed value from an array.
+
+For example: `delete_at(['a','b','c'], 1)` returns ['a','c']. *Type*: rvalue.
```
HelenCampbell: Missed the newline break for the type.

README.markdown
```diff
@@ -529,30 +836,34 @@ If the resource already exists, but does not match the specified parameters, thi
...
 
-  `ensure_resource('user', ['dan','alex'], {'ensure' => 'present'})`
+`ensure_resource('user', ['dan','alex'], {'ensure' => 'present'})`
 
 *Type*: statement.
 
 #### `ensure_resources`
 
-Takes a resource type, title (only hash), and a hash of attributes that describe the resource(s).
+TODO: I'm not sure I understand this one exactly. Need to clarify and rewrite.
```
HelenCampbell: I'll take a bash describing here, forgive me if it doesn’t help!
So if I want to ensure a resource is exists I can use this function. I pass in the type of the resource I'm looking for, it's title and a hash of attributes that will be checked against the resource, and if the resource isn’t present it will attempt creating it. 
So with the example:
`ensure_resource('user', 'jean', {'ensure' => 'present'})`
I am trying to ensure there is a user resource called jean, and I want to make sure it has attributes 'ensure' => 'present'.
If the resource does not exist it will create it:  user { ‘jean’:
	ensure => present,
}

If the (jean) resource already exists but does not match the specified parameters:
user { ‘jean’:
	ensure => absent,
}
this function will attempt to recreate the resource leading to a duplicate
resource definition error. Rough example: Error: Duplicate declaration: User[jean] is already declared in file /

In the other example it uses a hash of titles - this simply loops through the titles performs the aforementioned functionality on each of them.

README.markdown
```diff
@@ -650,233 +976,293 @@ Looks up a variable in a remote namespace.
...
+```puppet
+has_interface_with("lo")                        => true
+```
 
 *Type*: rvalue.
 
 #### `has_ip_address`
 
-Returns 'true' if the client has the requested IP address on some interface. This function iterates through the `interfaces` fact and checks the `ipaddress_IFACE` facts, performing a simple string comparison. *Type*: rvalue.
+Returns `true` if the client has the requested IP address on some interface.
```
HelenCampbell: I would keep the second sentence here as it's important to know where it's checking against and how. Otherwise you'd have to look at the function code itself to find out.

README.markdown
```diff
@@ -650,233 +976,293 @@ Looks up a variable in a remote namespace.
...
+
+Arguments: A string specifying an IP address.
+
+*Type*: rvalue.
 
 #### `has_ip_network`
 
-Returns 'true' if the client has an IP address within the requested network. This function iterates through the `interfaces` fact and checks the `network_IFACE` facts, performing a simple string comparision. *Type*: rvalue.
+Returns `true` if the client has an IP address within the requested network.
```
HelenCampbell: Again would leave in the second sentence for clarity to what it's being checked against.

README.markdown
```diff
@@ -1253,65 +1705,91 @@ $value = try_get_value($data, 'a/b/c/d', 'not_found')
...
 3. **'not_found'** The default value. It will be returned if nothing is found.
-   (optional, defaults to *undef*)
+   (optional, defaults to *`undef`*)
 4. **'/'** The path separator character.
    (optional, defaults to *'/'*)
 
-#### `type3x`
+*Type*: rvalue.
+
+#### `type3x` TODO: is this deprecated at this point?
```
HelenCampbell: It claims to be deprecated in the docs however there are no warnings which would make me worry about removing it. Most likely it will stay in and we will add a deprecation warning, then in the release after it will be fully removed. 
Arguments can be any of the following:

* string
* array
* hash
* float
* integer
* boolean

hunner: It was deprecated when it was added, and documented as such:
- https://github.com/puppetlabs/puppetlabs-stdlib/pull/382/files#diff-67df06d66a84fa8e9b63ecf6dc90c761R7
- https://github.com/puppetlabs/puppetlabs-stdlib/pull/382/files#diff-0a369498a5a8db3ac8fa606b544c9810R467

It can be removed when we drop puppet 3, as stated. It doesn't have to print a warning to be deprecated.

README.markdown
```diff
@@ -1253,65 +1705,91 @@ $value = try_get_value($data, 'a/b/c/d', 'not_found')
...
+
+Returns a string description of the type of a given value. The type can be a string, array, hash, float, integer, or Boolean. This function will be removed in a future release. For Puppet 4, use the new type system instead.
 
-Returns a string description of the type when passed a value. Type can be a string, array, hash, float, integer, or boolean. This function will be removed when Puppet 3 support is dropped and the new type system can be used. *Type*: rvalue.
+Arguments: TODO: what is the argument to use?
```
hunner: *arguments*: `type3x(Any $arg)`

README.markdown
```diff
@@ -1320,257 +1798,267 @@ $my_path3 = ['C:/Program Files (x86)/Puppet Labs/Puppet','C:/Program Files/Puppe
...
-  validate_integer(1, 3, 2)
-  validate_integer(1, 3, true)
-  ~~~
-
-  * Plus all of the above, but any combination of values passed as strings ('false' or "false").
-  * Plus all of the above, but with incorrect combinations of negative integer values.
-  * Plus all of the above, but with non-integer items in arrays or maximum / minimum argument.
-
-  *Type*: statement.
+Validates an integer or an array of integers. Terminates catalog compilation if any of the checks fail.
```
hunner: Can we just remove all documentation except the line stating that it is deprecated? We don't need to explain how to use deprecated functionality.

README.markdown
```diff
@@ -1712,61 +2211,73 @@ Validates that all passed values are string data structures. Aborts catalog comp
...
-  validate_string(true)
-  validate_string([ 'some', 'array' ])
-  ~~~
+```puppet
+validate_string(true)
+validate_string([ 'some', 'array' ])
+```
 
-*Note:* validate_string(undef) will not fail in this version of the functions API (incl. current and future parser).
+*Note:* validate_string(`undef`) will not fail in this version of the functions API (including current and future parser). TODO: Is this true for Puppet 3 only?
```
hunner: `undef` also validates as a string on my puppet 4.9, so I'd say on both 3 and 4.

README.markdown
```diff
@@ -1320,257 +1798,267 @@ $my_path3 = ['C:/Program Files (x86)/Puppet Labs/Puppet','C:/Program Files/Puppe
...
-  * Plus all of the above, but with incorrect combinations of negative integer values.
-  * Plus all of the above, but with non-integer items in arrays or maximum / minimum argument.
-
-  *Type*: statement.
+Validates an integer or an array of integers. Terminates catalog compilation if any of the checks fail.
+
+Arguments:
+
+* An integer or an array of integers, as the first argument.
+* Optionally, a maximum, as the second argument. All elements of the first argument must be equal to or less than this maximum. [TODO: Does (All elements of) mean the sum of all elements has to be equal to or less than the max, or that each element has to be equal to or less than the max. I can fix the instances after this too, don't worry about tagging them.]
```
hunner: "All elements of the first argument..." means "If the first argument is an array, all values in the array..."

hunner: Arrays in computer science are based on set theory, and "element" is a set theory word for "object in a set."

README.markdown
jbondpdx: Sounds legit to me, thank you!

README.markdown
jbondpdx: Yes, there's some effort in here to standardize some of that, but it might be that we do more of that as we transition stuff to Puppet 4.

README.markdown
```diff
@@ -232,13 +433,14 @@ base64('encode', 'https://puppetlabs.com', 'urlsafe')
...
 *Type*: rvalue.
 
 #### `basename`
 
-Returns the `basename` of a path (optionally stripping an extension). For example:
+Returns the `basename` of a path. An optional argument strips the extension. For example:
+
   * ('/path/to/a/file.ext') returns 'file.ext'
   * ('relative/path/file.ext') returns 'file.ext'
   * ('/path/to/a/file.ext', '.ext') returns 'file'
```
jbondpdx: Hah, yeah, I do not blame you.

README.markdown
```diff
@@ -52,167 +46,374 @@ The `stdlib::stages` class declares various run stages for deploying infrastruct
...
+
+*Type*: rvalue.
 
 #### `any2array`
 
-Converts any object to an array containing that object. Empty argument lists are converted to an empty array. Arrays are left untouched. Hashes are converted to arrays of alternating keys and values. *Type*: rvalue.
+Converts any object to an array containing that object. Converts empty argument lists are to empty arrays. Hashes are converted to arrays of alternating keys and values. Arrays are not touched.
+
+*Type*: rvalue.
+
```
jbondpdx: nice catch, thank you!

README.markdown
```diff
@@ -381,30 +649,52 @@ Prints deprecation warnings and logs a warning once for a given key:
...
-DEPRECATED: This function has been replaced in Puppet 4.5.0, use dig44() for backwards compatibility or use the new version.
+* `difference(["a","b","c"],["b","c","d"])` returns ["a"].
 
 *Type*: rvalue.
 
+#### `dig`
+
+> DEPRECATED: This function has been replaced with a built-in [`dig`](https://docs.puppet.com/puppet/latest/function.html#dig) function as of Puppet 4.5.0. Use [`dig44()`](#dig44) for backwards compatibility or use the new version.
+
 Retrieves a value within multiple layers of hashes and arrays via an array of keys containing a path. The function goes through the structure by each path component and tries to return the value at the end of the path.
```
jbondpdx: I'd rather not, because my perspective is that even though we are warning them, if it still works, they should be able to find out how to use it. Open to discussion, of course.

README.markdown
```diff
@@ -561,30 +872,41 @@ userlist:
     gid: 'mygroup'
 
 ensure_resources('user', hiera_hash('userlist'), {'ensure' => 'present'})
```
jbondpdx: @hunner , should I separate that line into a separate code block then?

README.markdown
jbondpdx: Oh, that helps me, thank you.

jbondpdx:
@HelenCampbell and @hunner , thank you!

README.markdown
```diff
@@ -247,131 +429,197 @@ Returns the `basename` of a path (optionally stripping an extension). For exampl
...
+Deletes a determined indexed value from an array.
+
+For example: `delete_at(['a','b','c'], 1)` returns ['a','c']. *Type*: rvalue.
 
 #### `delete_regex`
 
-Deletes all instances of a given element from an array or hash that match a provided regular expression. A string will be treated as a one-item array.
+Deletes all instances of a given element from an array or hash that match a provided regular expression. A string is treated as a one-item array.
+
+*Note:* This function is an implementation of a Ruby class and might not be UTF8 compatible. To ensure compatibility, use this function with Ruby 2.4.0 or greater.
```
jbondpdx: If we know when Puppet will ship the newer Ruby version, I can add that in.

README.markdown
jbondpdx: Added a link a couple lines down, with the arguments for this function.

jbondpdx:
Requested changes are made, except for my one outstanding question for @hunner above. If that code sample looks ok, this can be merged; if I need to fix it, I can do that. Oh, I can squash these too, if you want me to.

README.markdown
hunner: Yep.
```yaml
---
yaml: "goes here"
```
```puppet
ensure_resources('goes','here')
```

jbondpdx:
OK, this is squashed and updated, so maybe ready for merge now?

PR #749, opened 2017-03-28
Add an Ensure type

npwalker:


ekohl:
The problem with this is that in some cases the ensure is overloaded. Think of package where it can also be installed or a specific version. I wonder if this should be EnsurePresence would be a better name. Yay bikeshedding ;)

npwalker:
@ekohl I think the resolution to that is to have a Package_ensure type that allows for latest and versions.  But I think it's better to validate against the smallest set of values instead of anything that ensure will accept.  For example, ensure accepts true and false but they shouldn't be used.

ekohl:
@npwalker I agree, but my main worry is that people start using ```Ensure``` on ```package``` and accidentally break things. Another name could be ```BaseEnsure```. Then a ```PackageEnsure``` could extend this: ```Enum[BaseEnsure, 'installed', Pattern['\d+(\.\d+)*']]```.

alexjfisher:
@ekohl Further complicated by some package providers (ie puppet_gem) allowing things like `~> 1.0`

tphoney:
Hi everyone, I think there are a number of issues raised here which prevent this from being merged as is. I am closing this PR for now. If you can address the issues outlined above please feel free to create a new PR.  Many thanks for the work and reviewing people have done

PR #750, opened 2017-03-28
Add a type for ensure on service resources

npwalker:


ekohl:
Would ServiceEnsure or EnsureService be easier to read?

jonhattan:
Only first character uppercase, for consistency with existing types (f.e: `Stdlib::Absolutepath`).

`Ensureservice` seems better to me, so it's consistent with adding `Ensurewhatever` in a future.

ekohl:
A third option is ```Stdlib::Ensure::Service``` and ```Stdlib::Ensure::Whatever```.

types/serviceensure.pp
```diff
@@ -0,0 +1 @@
+type Stdlib::Serviceensure = Enum['stopped', 'running']
```
b4ldr: should this not be 
```ruby
type Stdlib::Serviceensure = Varient[Boolean,Enum['stopped', 'running']]
```

npwalker: IMO, using `ensure => true` or `ensure => false` on a service is an anti-pattern.  It works but it's just not as clear to new users what is going on.  As a convention I don't support it.

b4ldr: I tend to agree with you however if this is going into stdlib then i think it should conform to the documentation 
https://puppet.com/docs/puppet/5.3/type.html#service-attribute-ensure

david22swan:
Add the needed docs change and I will merge.

david22swan:
@npwalker Moving towards a clean release on this module to allow a full rubocop run, so would love to get this in before that. Please add the docs so I can do o, otherwise a full-rebase will be needed.

npwalker:
@david22swan okay updated.

PR #752, opened 2017-03-29
Fix placement of chaining arrow for style guide/linting

domcleal:
Corrects lint warning under puppet-lint 2.2.0:

    WARNING: arrow should be on the right operand's line on line 5


domcleal:
Fixed within 55d113ff.

PR #756, opened 2017-04-14
(maint) Stdlib::Compat::Integer accepts numbers with newlines apparently

hunner:


rnelson0:
https://github.com/puppetlabs/puppetlabs-stdlib/blob/e4ea1c46e7408f6878819dc08905be445a9c7bc8/lib/puppet/parser/functions/is_integer.rb#L36 - in `123\456`, the `$` matches the `\n` and ignores everything after that. Hooray!

eputnam:
@hunner would you mind adding a commit message? something like "#thisisfine so we're just updating tests"

PR #757, opened 2017-04-14
Add a stdlib::data_to_yaml function

npwalker:


npwalker:
@hlindberg I believe I made the function so that it only allows an Array or Hash parameter.  [Data](https://docs.puppet.com/puppet/4.10/lang_data_abstract.html#data) allows for undef and I'm not clear why I'd want to allow that.  

I added spec tests.

hlindberg:
Your specification of `Array` or `Hash` means that those can contain any kind of values, including any Ruby runtime object returned by a function. If you instead use `Data` that is taken care of for you as it only allows things that can be representation in JSON / YAML.

npwalker:
Okay validating Data instead of Hash and Array separately.

david22swan:
Is there any movement on this pr??

david22swan:
I apologize but given the amount of time that has passed since this pr was last altered, and the fact that the tests are routinely failing I feel that it must be closed.
I apologize if this inconveniences you and look forward to reviewing PRs from you in the future.

PR #758, opened 2017-04-18
Ruby 1.8 doesn't support open_args

sathieu:
Regression from #726


sathieu:
As Debian 6 is in the support matrix.

PR #759, opened 2017-04-21
Value function

binford2k:
A simple and safe way to get a value from a data structure. Like $facts.
This uses the same dot syntax that Facter cli uses and will return undef
if any part of the resolved value doesn't exist (instead of failing with 
`Operator '[]' is not applicable to an undef value` as using the raw data
structure would).

Example usage:

    $docker_ip = $facts.value('networking.interfaces.docker0.ip')

### DIFF
--- a/lib/puppet/functions/value.rb
+++ b/lib/puppet/functions/value.rb
@@ -1,0 +1,29 @@
+# A simple and safe way to get a value from a data structure. Like $facts.
+# This uses the same dot syntax that Facter cli uses and will return undef
+# if any part of the resolved value doesn't exist (instead of failing as
+# using the raw data structure would).
+#
+# @example Display a deeply nested fact
+#
+#       $docker_ip = $facts.value('networking.interfaces.docker0.ip')
+#
+#       if($docker_ip) {
+#         notify { "The Docker interface is: ${docker_ip}": }
+#       }
+#
+#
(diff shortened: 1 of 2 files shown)


hlindberg:
There is an issue with this approach as all parts of the "dotted string" become string keys. It is therefore not possible to use it with a mix of arrays and hashes, and it will not work with numeric keys in a hash. Given that the entire functionality is this:
```
call_function('dig', data, *lookup_key.split('.'))
```
the same can be done directly in puppet (similar example):
```
$facts.dig(*"os.release.major".split(/\./))
```
It would perhaps be of greater value to have a function that expands dot-notation-string to an array of indexes (a function that also allows escapes etc. just like what is supported in lookup - it could probably reuse that logic). As an example - imagine that function to be named `dotted_path` - you would then write:
```
$facts.dig(*dotted_path("os.release.major"))
```

binford2k:
@hlindberg Do you think that someone who struggles with the difference between `facter os.release.major` and `$facts['os']['release']['major']` would grok `$facts.dig(*"os.release.major".split(/\./))`?

Or even `$facts.dig(*dotted_path("os.release.major"))`?

tphoney:
Just saying something here, to re-ignite this pr. @hlindberg does what @binford2k says, make sense ?

binford2k:
Closed in favor of #787

PR #761, opened 2017-04-25
(PE-20308) Pass a literal type and not a string to findresource

hunner:
- `defined_with_params` calls `findresource(reference.to_s)`
- `findresource` is
  https://github.com/puppetlabs/puppet/blob/4.8.1/lib/puppet/parser/compiler.rb#L407
  and points to
  https://github.com/puppetlabs/puppet/blob/4.8.1/lib/puppet/resource/catalog.rb#L352
- This calls `Puppet::Resource.new` with the type
  https://github.com/puppetlabs/puppet/blob/4.8.1/lib/puppet/resource/catalog.rb#L366
- This ends up calling `resource_type` via
  https://github.com/puppetlabs/puppet/blob/4.8.1/lib/puppet/resource.rb#L317-L319
  and that ends up declaring the type via the autoloader api at
  https://github.com/puppetlabs/puppet/blob/4.8.1/lib/puppet/resource.rb#L390
- But why does the autoloader API fail to find it in the current
  environment?
- Okay, so when the autoloader is trying to find the type, it uses the
  typeloader to look it up in the current environment
  https://github.com/puppetlabs/puppet/blob/4.8.1/lib/puppet/metatype/manager.rb#L171
- this calls `get_file` and `mark_loaded`
  https://github.com/puppetlabs/puppet/blob/4.8.1/lib/puppet/util/autoload.rb#L64-L67

Suggested workaround is to pass a literal type instead of a string to
`findresource` to fix in stdlib, and also to fix loading/requiring of
type in core puppet.

### DIFF
--- a/lib/puppet/parser/functions/defined_with_params.rb
+++ b/lib/puppet/parser/functions/defined_with_params.rb
@@ -24,7 +24,25 @@ ENDOFDOC
     params = {}
   end
   ret = false
-  if resource = findresource(reference.to_s)
+
+  if Puppet::Util::Package.versioncmp(Puppet.version, '4.5.0') >= 0
+    # Workaround for PE-20308
+    if reference.is_a?(String)
+      type_name, title = Puppet::Resource.type_and_title(reference, nil)
+      type = Puppet::Type.type(type_name)
+    elsif reference.is_a?(Puppet::Resource)
+      type = reference.resource_type
+      title = reference.title
+    else
+      raise(ArgumentError, "Reference is not understood: '#{reference.class}'")
+    end
+    #end workaround
+  else
+    type = reference.to_s
+    title = nil
+  end
+
+  if resource = findresource(type, title)
     matches = params.collect do |key, value|
       # eql? avoids bugs caused by monkeypatching in puppet
       resource_is_undef = resource[key].eql?(:undef) || resource[key].nil?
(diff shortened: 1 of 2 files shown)


lib/puppet/parser/functions/defined_with_params.rb
```diff
@@ -24,7 +24,19 @@
     params = {}
   end
   ret = false
-  if resource = findresource(reference.to_s)
+
+  # Workaround for PE-20308
+  if reference.is_a?(String)
+    type_name, title = Puppet::Resource.type_and_title(reference, nil)
```
thallgren: The `Puppet::Resource#type_and_title` class method was introduced in Puppet 4.5.0. Perhaps that's what causing appveyor to fail for 4.2.3?

lib/puppet/parser/functions/defined_with_params.rb
```diff
@@ -24,7 +24,19 @@
...
+  elsif reference.is_a?(Puppet::Resource)
+    type_name, title = Puppet::Resource.type_and_title(reference.type, reference.title)
```
thallgren: Would it not be more efficient to just ask the resource for it's type here?
```
type = reference.resource_type
```

hunner: I had been looking for something like that originally, but overlooked it. Thanks!

lib/puppet/parser/functions/defined_with_params.rb
hunner: I'll just fallback to the old behavior pre 4.5.0 as the ticket only reports the behavior on newer puppet versions.

domcleal:
This change doesn't work with defined types, only native types - I filed https://tickets.puppetlabs.com/browse/MODULES-4790 ([tests](https://github.com/domcleal/puppetlabs-stdlib/commit/c8441b20d39ceee47739dc18695095e7b4fa3293)). Unfortunately this affects `ensure_resource`, causing duplicate resources and catalog build failures.

PR #762, opened 2017-04-26
Fix issue where the following causes obscure catalog compilation errors:

dacron:
```
file { '/tmp/somefile':
  ensure => 'file',
}

File['/tmp/somefile'] -> Package <| |>

ensure_packages($somearray)
```

If $somearray is undefined or one of the elements contains an empty
string, an error like the following is thrown:

Could not find resource 'Package[]' for relationship from
'File[/tmp/somefile]' on node $::fqdn


hunner:
Manually merged to release branch via https://github.com/puppetlabs/puppetlabs-stdlib/pull/764 and will go out with the next release, then merged back to master. Thanks!

PR #765, opened 2017-04-28
(PE-20308) Fix defined_with_params() for defined type strings & references

hunner:


lib/puppet/parser/functions/defined_with_params.rb
```diff
@@ -30,6 +30,7 @@
     if reference.is_a?(String)
       type_name, title = Puppet::Resource.type_and_title(reference, nil)
       type = Puppet::Type.type(type_name)
+      type = type_name if type.nil?
```
thallgren: Perhaps the best way would be to resolve the type the same way it's resolved in the `is_defined` function. I.e, replace the two lines with:
```
type = Puppet::Pops::Evaluator::Runtime3ResourceSupport.find_resource_type_or_class(scope, val)
```
That will ensure some level of consistency. Settling for a string here will cause yet another call to `Puppet::Type#type` from the `Puppet::Resource#resource_type` method.

hunner: Ah, I never knew how to do this. Thanks!

buzzdeee:
Is this also going into the master branch as well?

tphoney:
Yes this will be merged back into master once we release 4.17.0, which should happen with in the next few days.  @buzzdeee thank you for your patience.

PR #766, opened 2017-05-02
Update alias spec error message expectation for PUP-6694

domcleal:
The alias data type name is now included in the error message for data
type mismatches, since puppetlabs/puppet@aef26614. This fixes the
following class of failures against Puppet's master branch:

    1) test::float rejects other values true should fail to compile and raise an error matching /parameter 'value' expects a value of type Float or Pattern(\[.*\]+)?/
       Failure/Error: it { is_expected.to compile.and_raise_error(/parameter 'value' expects a value
         of type Float or Pattern(\[.*\]+)?/) } error during compilation: Evaluation Error: Error while evaluating a Resource Statement,
         Class[Test::Float]: parameter 'value' expects a Stdlib::Compat::Float = Variant[Float, Pattern[/^-?(?:(?:[1-9]\d*)|0)(?:\.\d+)(?:[eE]-?\d+)?$/]] value, got Boolean at line 1:1 on node [..]
       # ./spec/aliases/float_spec.rb:23:in `block (5 levels) in <top (required)>'

### DIFF
--- a/spec/aliases/absolutepath_spec.rb
+++ b/spec/aliases/absolutepath_spec.rb
@@ -40,7 +40,7 @@ if Puppet::Util::Package.versioncmp(Puppet.version, '4.5.0') >= 0
         ].each do |value|
           describe value.inspect do
             let(:params) {{ value: value }}
-            it { is_expected.to compile.and_raise_error(/parameter 'value' expects a match for Variant/) }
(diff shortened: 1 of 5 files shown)


domcleal:
715c40d and 33922a4 were merged instead.

PR #767, opened 2017-05-03
(MODULES-3432) Adds getvar_emptystring function

petems:
* To be combined with pick
* Allows a degree of backwards compatibility with
pick() and getvar()
* For example:
`$redis_version_real = pick(getvar_emptystring('$::redis_server_version'), $minimum_version)`

This will work with:
* Puppet 3
* Puppet 4
* Puppet 3 with strict_variables
* Puppet 4 with strict_variables

### DIFF
--- a/lib/puppet/parser/functions/getvar_emptystring.rb
+++ b/lib/puppet/parser/functions/getvar_emptystring.rb
@@ -1,0 +1,35 @@
+module Puppet::Parser::Functions
+
+  newfunction(:getvar_emptystring, :type => :rvalue, :doc => <<-'ENDHEREDOC') do |args|
+    Lookup a variable in a remote namespace.
+    Return an empty string rather than undef/nil if it could not be found
+
+    For example:
+
+        $foo = getvar('site::data::foo')
+        # Equivalent to $foo = $site::data::foo
+
+    This is useful if the namespace itself is stored in a string:
+
+        $datalocation = 'site::data'
+        $bar = getvar("${datalocation}::bar")
+        # Equivalent to $bar = $site::data::bar
+    ENDHEREDOC
+
+    unless args.length == 1
+      raise Puppet::ParseError, ("getvar_emptystring(): wrong number of arguments (#{args.length}; must be 1)")
+    end
+
+    begin
+      result = self.lookupvar("#{args[0]}")
+
+      result = '' if result.nil?
+
+      result
+    rescue Puppet::ParseError # Eat the exception if strict_variables = true is set and return an empty string
+      return ''
+    end
+
+  end
+
+end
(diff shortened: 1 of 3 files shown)


lib/puppet/parser/functions/getvar_emptystring.rb
```diff
@@ -0,0 +1,35 @@
...
+      raise Puppet::ParseError, ("getvar_emptystring(): wrong number of arguments (#{args.length}; must be 1)")
+    end
+
+    begin
+      result = self.lookupvar("#{args[0]}")
+
+      result = '' if result.nil?
+
+      result
+    rescue Puppet::ParseError # Eat the exception if strict_variables = true is set and return an empty string
```
hlindberg: This is not safe - it will catch other kinds of errors as well as "non existing variable".

lib/puppet/parser/functions/getvar_emptystring.rb
```diff
@@ -0,0 +1,35 @@
...
+    ENDHEREDOC
+
+    unless args.length == 1
+      raise Puppet::ParseError, ("getvar_emptystring(): wrong number of arguments (#{args.length}; must be 1)")
+    end
+
+    begin
+      result = self.lookupvar("#{args[0]}")
+
+      result = '' if result.nil?
```
hlindberg: This is not nice - it translates puppet `undef` to empty string (the result if you render the `undef` as a string. IMO this brings back the same kind of old confusion over `undef` `:undef` and empty-string.

lib/puppet/parser/functions/getvar_emptystring.rb
```diff
@@ -0,0 +1,35 @@
...
+        $bar = getvar("${datalocation}::bar")
+        # Equivalent to $bar = $site::data::bar
+    ENDHEREDOC
+
+    unless args.length == 1
+      raise Puppet::ParseError, ("getvar_emptystring(): wrong number of arguments (#{args.length}; must be 1)")
+    end
+
+    begin
+      result = self.lookupvar("#{args[0]}")
```
hlindberg: This turns whatever args[0] is into a string. The function should check that the argument is a string of length >= 1. If also supporting numeric variables (match vars) it could allow integer values.

tphoney:
@petems can you have another look at this please. also 👺

david22swan:
Is this still active?

hlindberg:
@david22swan waiting on contributor to make changes - will not be merged as it stands now. I also question the need for this function in stdlib in the first place and if circumventing strict_variables and mixing up undef and ''  again is such a bright idea...

davidmalloncares:
@david22swan Hi David, can we move ahead with our planned changes please? Peter can rebase at a later date if he wants to continue with his proposed changes

david22swan:
Closing ticket now

david22swan:
We are currently aiming for a clean release so that the entire module can be put through the Rubocop process. If the ticket is important please rebase and reupload it after.

PR #777, opened 2017-05-14
Fix filenames of two function spec tests

alexjfisher:
These tests weren't being run.  Total tests increase from 2742 to 2769.


tphoney:
@alexjfisher yo yo, totally agree with the rename, can you look at the failure, if not i will jump on it next week

alexjfisher:
@tphoney I fixed up one of the tests and now the travis build passes.  Appveyor is still failing for puppet 4.2.  Looks like the error messages aren't the same for older and new versions of puppet.  These are the tests that fail. https://github.com/puppetlabs/puppetlabs-stdlib/blob/5a17bf1c338c74bd6c8dbacc9a05b4cd501ecc23/spec/functions/length.rb#L5-L8

spec/functions/length_spec.rb
```diff
@@ -26,10 +26,6 @@
   it { is_expected.to run.with_params('āβćđ').and_return(4) }
 
   context 'when using a class extending String' do
-    it 'should call its size method' do
-      value = AlsoString.new('asdfghjkl')
-      value.expects(:length).returns('foo')
```
alexjfisher: I'm not sure it ever made any sense to mock returning a string when the function should only ever return an int.

tphoney:
Thanks for the changes @alexjfisher ,agreed with your conclusion. I am looking into weird appveryor ness at the minute. Hopefully i will get something merged soon.

tphoney:
Thanks for the PR, stdlib is failing but not because of this. merging

PR #783, opened 2017-06-16
Fix headers in CHANGELOG.md so that headers render correctly

davewongillies:
Just a quick fix to correct the formatting on headers in `CHANGELOG.md`. Because of the missing space after the hashes, headers weren't rendering.


tphoney:
Thanks for the fix ! @davewongillies

PR #786, opened 2017-06-28
(MODULES-5113) Make line support Sensitive

reidmv:
Do this by making it a property. The idea is to get rid of the warning
that appears when a Sensitive value is passed to the line attribute.


tphoney:
apologies for closing and re-opening, i was testing a theory around appveyor, previously this was green. After re-running the tests it is red in appveyor. i am investigating and will get this merged asap

PR #792, opened 2017-07-12
add additional aliasis

b4ldr:
Add additional types to standard lib for 

* `Stdlib::Port`: all valid TCP/UDP ports
* `Stdlib::Privilegedport`:  ports which need root power to bind
* `Stdlib::Unprivilegedport`:  ports which do not need root power
* `Stdlib::Fileuri`: match file:///path/to/file
* `Stdlib::Puppeturi`: match puppet:///modules/path/to/file
* `Stdlib::Puppetsource`: matches values that can be used for a file type source parameter
* `Stdlib::Puppetcontent`: matches values that can be used for a file type content parameter
* `Stdlib::Base64`: matches base64 strings
* `Stdlib::Base32`: matches base64 strings
* `Stdlib::Fqdn`: loose check for a fully qualified domain name 
* `Stdlib::Host`: fqdn or ip address

i think the port types originally came from voxpupuli/puppet-tea so can remove them if desired.  it seems i [icann-dns/puppet-tea] and voxpupuli/puppet-tea have also implemented some syslog types would you be intrested in adding them or a variation their of?  

I would also be keen to move the IP and other compat functions to the Stdlib level.  i.e. `Stdlib::Ip_address` instead of `Stdlib::Compat::Ip_address` .  just simply because the latter is getting on the large side.  would be happy to to the work just wondered if there is philosophical objections to doing something like

```puppet
type Stdlib::Ip_address = Stdlib::Compat::Ip_address
```

### DIFF
--- a/spec/aliases/base32_spec.rb
+++ b/spec/aliases/base32_spec.rb
@@ -1,0 +1,54 @@
+require 'spec_helper'
+
+if Puppet.version.to_f >= 4.5
+  describe 'test::base32', type: :class do
+    describe 'valid handling' do
+      %w(
+        ASDASDDASD3453453
+        ASDASDDASD3453453=
+        ASDASDDASD3453453==
+        ASDASDDASD3453453===
+        ASDASDDASD3453453====
+        ASDASDDASD3453453=====
+        ASDASDDASD3453453======
+        asdasddasd3453453
+        asdasddasd3453453=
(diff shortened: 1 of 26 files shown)


b4ldr:
any feedback on this?

b4ldr:
i have split this up so that its easier to review.  please see #839 #840 #841 #842 #843

PR #794, opened 2017-07-13
(MODULES-5003) file_line does not change multiple lines when one matches

tkishel:
The exists? method is called to determine the need to call the create and
destroy methods. When the :ensure, :match, and :multiple attributes are defined,
the exists? method needs to return false unless all the lines that match the
:match attribute are equal to the :line attribute.

The first commit is minimal, and implements this change.
The second commit normalizes the code, which I needed for comprehension.
Either resolves the issue.


tphoney:
Thanks @tkishel superb work

PR #795, opened 2017-07-14
enable parallel testing for appveyor

tphoney:


tphoney:
@glennsarti @jpogran  what do you think ?

glennsarti:
Yup that's you do it.  You'd probably want to update sync.yml otherwise modsync will overwrite it.  Example is in chocolatey (https://github.com/puppetlabs/puppetlabs-chocolatey/blob/master/.sync.yml#L45-L46)

tphoney:
question being, is this something we want to roll out to other modules / all modules ?
or stage it

tphoney:
@glennsarti changed to use release_check ala travis once this goes green we can look at modulesync

appveyor.yml
```diff
@@ -42,7 +42,7 @@ build: off
 test_script:
 - bundle exec puppet -V
 - ruby -v
-- bundle exec rake spec SPEC_OPTS='--format documentation'
+- bundle exec rake release_check SPEC_OPTS='--format documentation'
```
glennsarti: typo - Should be `bundle exec rake release_checks`

tphoney: yeah :( lateness does not equal greatness...

glennsarti:
@tphoney If you want to rerun jobs, you can just do that from the AppVeyor console

PR #797, opened 2017-07-21
Update README for fact() function

reidmv:
Add README update for `fact()` function. This was missed in #787.


reidmv:
@npwalker README update

PR #810, opened 2017-08-28
(MODULES-5546) add check for pw_hash

eputnam:
Older versions of Puppet end up choking on the sensitive data type check in the pw_hash function added in 42d4ea7. this adds a check for the Sensitive class and then the type of the arg, so that if the class has not been declared, it will just move on.


eputnam:
yeah i like that better.

PR #811, opened 2017-08-31
Allow root as valid UNIX path

kofrezo:
Since the purpose of this validation type is to check if the given parameter is a valid UNIX path it should also cover the root path "slash" since it might be possible that it is specified as a target directory.

### DIFF
--- a/types/unixpath.pp
+++ b/types/unixpath.pp
@@ -1,2 +1,2 @@
 # this regex rejects any path component that is a / or a NUL
-type Stdlib::Unixpath = Pattern[/^\/([^\/\0]+\/*)+$/]
+type Stdlib::Unixpath = Pattern[/^\/([^\/\0]+\/*)*$/]


pmcmaw:
Hey @kofrezo I have had a look at your change. I see you have 2 commits for one change, if you could squash the commit and add a test to test the change I will get this merged. 

Many Thanks for your contribution, it is much appreciated.

kofrezo:
Hi @pmcmaw
Sorry for late response I was on vaccation I sqaushed the commits to one so if tests pass I think everything should be fine :-)

kofrezo:
@pmcmaw I don't really get why it is failing could you please help me with it.

pmcmaw:
@kofrezo There was an issue with a few tests. This has been resolved by @tphoney and now the build is passing. Many Thanks for your contribution. 👍

PR #813, opened 2017-09-04
Make Unixpath accept "/"

hasegeli:
This is a perfectly valid path.


hasegeli:
I haven't noticed that my colleague already submit #811.

PR #816, opened 2017-09-13
use single quotes in validate_legacy example code

mutante:
I used the example code from this README file with our puppet code but i got downvotes from CI / jenkins / puppet-lint because it doesn't conform to Puppet style guides to use double quotes here.

It produces "double quoted string containing no variables" warnings from puppet-lint.

I wanted to fix the example locally but it was pointed out to me we should fix it upstream stdlib instead, so here it is. Just use single quotes here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2214,7 +2214,7 @@ Arguments:
 Example:
 
 ```puppet
-validate_legacy("Optional[String]", "validate_re", "Value to be validated", ["."])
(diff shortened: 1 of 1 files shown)


pmcmaw:
@mutante I have re kicked Travis and it is now passing. 
If you could squash your 2 commits I will get your change merged.

PR #819, opened 2017-09-27
(MODULES-5651) append_on_no_match method fix

eputnam:
This renames append_on_no_match to simply append to relieve lots of double negatives and adds a fix to the method. Previously, if the param was set to false then it always returns true before the replace param is evaluated, so nothing happens. This being the case, I added an additional condition , so that it only acts when there's no match.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,11 +80,11 @@ The `stdlib` class has no parameters.
 
 * `stdlib::stages`: Manages a standard set of run stages for Puppet.
 
-### Defined types
+### Custom types
 
 #### `file_line`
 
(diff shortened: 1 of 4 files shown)


tphoney:
yo @eputnam  this looks to be a backwards breaking change.. can we leave the current functionality and mark it as deprecated.

README.md
```diff
@@ -241,6 +241,16 @@ Boolean.
...
 
+##### `append`
+
+Specifies whether or not the line is appended to the end of the file when no match is found.
+
+Boolean.
+
+Values: `true`, `false`
+
+Default value: `false`.
```
eputnam: defaults to `true` silly

hunner:
@tphoney `append_on_no_match` was not a documented parameter in the README and was not mentioned in the changelog, and is so poorly named that I would advocate "bugfixing" it in master now and calling it a new feature in the next release.

PR #820, opened 2017-09-29
(maint) Clarify docs and add new tests

alexharv074:
Based on a Stack Overflow question, it was noted that the documentation
for `file_line` isn't complete and the underlying behaviour somewhat
confusing.

https://stackoverflow.com/questions/46468922/how-to-change-the-contents-of-a-file-by-using-puppet/46473458

In this patch I add additional unit tests and better examples and
complete documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,29 +117,55 @@ In the example above, `match` looks for a line beginning with 'export' followed 
 
 Match Example:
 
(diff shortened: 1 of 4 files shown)


tphoney:
Awesome !! thanks @alexharv074

PR #821, opened 2017-09-29
(MODULES-5687) Merge duplicate key that is an array

cropalato:
If the idea is to merge values, why not merge the arrays, when both values are array?


david22swan:
Modifying the method in this manner will cause backwards compatibility errors. If it is no bother please separate your changes into a new method entirely.
This was previously discussed in https://github.com/puppetlabs/puppetlabs-stdlib/pull/304

david22swan:
@cropalato I apologize but given that this PR makes changes that are backwards incompatible I feel that I must close it. If you still wish for your changes to be merged in please separate them into a new method and then resubmit in a new PR.

PR #823, opened 2017-10-04
(MODULES-5679) Add a new function ifelse to match ruby's tenary operator

vStone:
Allows for more compact, clean code.


david22swan:
LGTM

hlindberg:
Why on earth do we need a function like this in stdlib?!? This is written like this in the language `if $val { $iftrue } else { $iffalse }`.
I understand the motivation (somwhat), but please, do this in a separate module.

david22swan:
I apologize for the inconvenience but it has bee asked that I undo this merge, as stated above if you wish to keep this change please do so in a separate module.
Once again I apologize.

PR #788, opened 2017-07-03
(MODULES-5003) file_line fix all broken lines

tphoney:
If the following lines exist
hard core 90
hard core 10
it correctly corrects both lines to
hard core 0
hard core 0
However if these lines exist in the file
hard core 90
hard core 0
It incorrectly does nothing 
It should correct the first line

PR #825, opened 2017-10-12
(MODULES-5651) Do not append infinitely

hunner:
https://tickets.puppetlabs.com/browse/MODULES-5003 gave rise to #788 and #794 which caused different behavior based on whether the line value was matched by the match regex or not.

The change in behavior was both breaking and broken, though that was hard to tell because the behavior was ill-described in general.

[bugfix] This commit resolves the breaking behavior by reverting the behavior of "replacing matches when a line matching `line` exists even when `multiple` is set to `true`".

[feature] This commit adds a new parameter to make file_line replace all matches universally with the `line` value, even when the line exists elsewhere in the file. This feature only affects modifying multiple lines in a file when the `line` value already exists.

[bugfix] This commit more strictly defines the various interactions of `ensure`, `match`, `append_on_no_match`, `replace`, `multiple`, and `replace_all_matches_not_matching_line`. It also more clearly documents and tests these interactions.

### DIFF
--- a/spec/unit/puppet/provider/file_line/ruby_spec.rb
+++ b/spec/unit/puppet/provider/file_line/ruby_spec.rb
@@ -1,159 +1,370 @@
 #! /usr/bin/env ruby -S rspec
 require 'spec_helper'
-require 'tempfile'
+
 provider_class = Puppet::Type.type(:file_line).provider(:ruby)
 # These tests fail on windows when run as part of the rake task. Individually they pass
 describe provider_class, :unless => Puppet::Util::Platform.windows? do
-  context "when adding" do
-    let :tmpfile do
-      tmp = Tempfile.new('tmp')
-      path = tmp.path
-      tmp.close!
-      path
-    end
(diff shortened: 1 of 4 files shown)


hunner:
Needs some commenting clean up, tests added & cleaned up, and docs proof read.

The logic in the provider and error-checking in the type can be proof-read though.

spec/unit/puppet/provider/file_line/ruby_spec.rb
```diff
@@ -171,244 +382,252 @@
...
             fh.write("foo1\nfoo = blah\nfoo2\nfoo = baz")
           end
+          expect(@provider.exists?).to be true
+          expect(File.read(tmpfile).chomp).to eql("foo1\nfoo = blah\nfoo2\nfoo = baz")
         end
+      end
+    end
+  end
+  context "when match+replace+append_on_no_match" do
+  end
```
eputnam: oops?

PR #831, opened 2017-10-26
Add merge_resource_hash function

raphink:
Takes any number of hashes of hashes (in the format fit for create_resources()).
For each key as hash, merges all subkey values (i.e. resource parameters) into arrays.

E.g. given these three hashes:

```puppet
   $h1 = {
     k1 => { p1 => 'v1' },
     k2 => { p1 => ['v211', 'v212'] }
   }

   $h2 = {
     k2 => {
       p1 => 'v213',
       p2 => 'v221'
     }
   }

   $h3 = {
     k2 => {
       p1 => ['v214', 'v215'],
       p3 => 'v231'
     },
   }
```

it will produce:

```puppet
   {
     k1 => {p1 => 'v1'},
     k2 => {
       p1 => ['v211', 'v212', 'v213', 'v214', 'v215'],
       p2 => 'v221',
       p3 => 'v231'
     }
   }
```

### DIFF
--- a/lib/puppet/parser/functions/merge_resource_hash.rb
+++ b/lib/puppet/parser/functions/merge_resource_hash.rb
@@ -1,0 +1,63 @@
+#
+# merge_resource_hash.rb
+#
+
+module Puppet::Parser::Functions
+  newfunction(:merge_resource_hash, :type => :rvalue, :doc => <<-EOS
+    Takes any number of hashes of hashes (in the format fit for create_resources()).
+    For each key as hash, merges all subkey values (i.e. resource parameters) into arrays.
+
+    E.g. given these three hashes:
+
+       $h1 = {
+         k1 => { p1 => 'v1' },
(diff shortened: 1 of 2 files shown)


rico89:
Exactly what we need!

ripienaar:
I'd be against this, `create_resources` should go away, this looks like something that should not be a in a "standard library".  But to my mind this is true for almost everything going in here recently.

General call for better curation of what might realistically go in this repos and if the thigns being added is really helping people

david22swan:
Rekicked tests and they passed.
Once docks are added I will merge

hlindberg:
-1 - please to not merge to stdlib. Put this in a module of your own if you like to use this.

david22swan:
It has been requested that I close this merge, I apologize for any wasted time and if you wish to continue to use it please keep it in a separate module.

raphink:
No problem. Just wanted to suggest it in case it would be interesting. We can definitely host it somewhere else.

PR #839, opened 2017-11-13
add Stdlib::Port, Stdlib::Privilegedport & Stdlib::Unprivilegedport

b4ldr:


b4ldr:
might be better to have 

```puppet
Stdlib::Port
Stdlib::Port::Privlaged
Stdlib::Port::Unprivlaged
```

happy to change if thats prfered

b4ldr:

pegasd:
I'd love to see some progress on this.
Is there a specific hold-up preventing something in here?

FWIW, I'd love to see the alternative naming version implemented. Namely, these would be great:

```
Stdlib::Port
Stdlib::Port::Privileged
Stdlib::Port::Unprivileged
```

b4ldr:
I did a rebase to fix the README conflict and renamed the types as suggested.  `Stdlib::Port::Privileged` seems more inline with other types

PR #840, opened 2017-11-13
add Stdlib::base64 and Stdlib::Base32 types

b4ldr:


b4ldr:

david22swan:
@b4ldr Once rebased I would fell happy to merge.

b4ldr:
@david22swan rebased, thanks

PR #841, opened 2017-11-13
add Stdlib::Filesource

b4ldr:


b4ldr:

pegasd:
As per #856, you don't need the test::* manifest anymore, so it can be safely deleted from PR.

b4ldr:
thanks @pegasd i missed that one, updated now

spec/type_aliases/filesource_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+    describe 'valid handling' do
+      %w[
+        https://hello.com
+        https://notcreative.org
+        https://canstillaccepthttps.co.uk
+        http://anhttp.com
+        http://runningoutofideas.gov
+        file:///hello/bla
+        file:///foo/bar.log
+        puppet:///modules/foo/bar.log
```
binford2k: You should also test `puppet://someserver/modules/foo/bar.log`

b4ldr: @binford2k thanks for the pointer i didn't see that varient in the docs.  Have updeated the type pattern and spec tests

david22swan:
@binford2k Are you happy that the changes you requested have been made?

spec/type_aliases/filesource_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+          [nil],
+          [nil, nil],
+          { 'foo' => 'bar' },
+          {},
+          '',
+          '*/Users//nope',
+          '\\Users/hc/wksp/stdlib',
+          'C:noslashes',
+          '\\var\\tmp',
+          'puppet:///foo/bar.log',
```
alexjfisher: @binford2k This is actually valid though?  https://puppet.com/docs/puppet/5.5/file_serving.html#whats-a-mount-point-in-a-puppet-uri
We're using this type in puppet-zabbix and a user has complained the regex is rejecting valid puppet fileserver mount points. https://github.com/voxpupuli/puppet-zabbix/issues/580

binford2k: Yes, that should be valid.

PR #843, opened 2017-11-13
update Stdlib::*::ip* types

b4ldr:
The current `Stdlib::Compat::IPv6` type has a bug in that it would matches the following incorrect strings

```shell
foobbar2001::db8:1
2001::db8::1
```
This is because it does not surround the pattern match the beginning of the string with the '^'.   Further to this the current pattern allows the parameter to be surrounded with whitespace which is different behaviour to the `Stdlib::Compat::Ipv4` type.  

My understanding is that the compat functions should exactly match the legacy validate_* functions as such i have moved the fix to this type out of the compat 
folder.  I have also created a new `Stdlib::Ip_address` type which references the new `Stdlib::Ipv6` type and finally created a `Stdlib::Ipv4` type for consistency which is currently just an alias to the compat type.


b4ldr:

david22swan:
@b4ldr Could you rebase this please.

b4ldr:
@david22swan fixed conflicts

PR #851, opened 2017-11-22
Add test for https://github.com/puppetlabs/puppetlabs-stdlib/pull/850

sean797:
Thanks for fixing it quickly! Just adding a test :-)


pmcmaw:
@sean797 fantastic! 
As soon as this runs through Travis CI I will get it merged. 
Thank you for alerting us so quickly.

We are also currently working on a release which should be out in the next day with this change. 👍

PR #855, opened 2017-11-28
Ability to skip undef values in to_json_pretty()

pegasd:
Note that I also added a trailing newline at the end of `to_json_pretty()` output. Seems like a desirable change since usually we output to a file.


pmcmaw:
@pegasd thank you for your contribution, it is much appreciated 👍

PR #856, opened 2017-11-28
Updated type alias tests and dropped superfluous wrapper classes

pegasd:
Since newer versions of rspec-puppet support testing type aliases directly, let's drop superfluous wrapper classes and start testing properly.


b4ldr:
+1

pmcmaw:
Looks good to me! 👍 
Thank you for taking the time to create this PR, it is much appreciated.

PR #860, opened 2017-12-05
(MODULES-6216) Add acceptance tests for type3x() function

ghoneycutt:


ghoneycutt:
This should fail because `type3x()` is broken. See https://tickets.puppetlabs.com/browse/MODULES-6216

pmcmaw:
Including these changes in #861.
Many thanks for your contribution, it is much appreciated.

PR #862, opened 2017-12-05
fixed wrong comment in unixpath.pp

c33s:
https://tickets.puppetlabs.com/browse/MODULES-6221

### DIFF
--- a/types/unixpath.pp
+++ b/types/unixpath.pp
@@ -1,2 +1,2 @@
-# this regex rejects any path component that is a / or a NUL
+# this regex rejects any path component that does not start with "/", contain a "." or is NUL
 type Stdlib::Unixpath = Pattern[/^\/([^\/\0]+\/*)*$/]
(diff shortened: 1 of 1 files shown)


pegasd:
Should the `contain a "."` part be here?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

c33s:
oh this was a wrong assumption. i digged through an puppetlabs-lvm error to `stdlib::absolutepath` which is using `unixpath`. so a assumed it has to be a absolute unix path which also does not contain any relative parts (`/home/../etc`)

i remove the dots part from the PR

tested the regex with https://regex101.com/

is/should this be a valid unixpath?

```/./......../'''ääÄÄÄvffoo/ü/:\\-__##../bar/com```

eputnam:
That _is_ a valid unix path, yes.

tphoney:
Thanks @c33s  !

PR #869, opened 2017-12-25
Added ensure_* functions

pegasd:
After seeing lots of these

```puppet
if $ensure == present {
  $ensure_directory = directory
} else {
  $ensure_directory = absent
}

file { '/somedir':
  ensure => $ensure_directory,
}
```

I decided to implement helper functions and abstract this logic away.

Usage is as follows:

```puppet
file { '/somedir':
  ensure => ensure_directory($ensure),
}

file { '/somedir/somelink':
  ensure => ensure_link($ensure),
}
```

### DIFF
--- a/lib/puppet/functions/ensure_directory.rb
+++ b/lib/puppet/functions/ensure_directory.rb
@@ -1,0 +1,37 @@
+# Takes a generic 'ensure' parameter (or its boolean equivalent)
+# and converts it to an appropriate value for use with directory
+# declaration.
+Puppet::Functions.create_function(:ensure_directory) do
+  # @param ensure_param Pass ensure value here
+  # @return [String] 'directory' or 'absent'
+  #
+  # @example Calling the function with 'present'
+  #   ensure_directory('present') # returns 'directory'
+  #
+  # @example Calling the function with boolean parameter
+  #   ensure_directory(true) # returns 'directory'
+  #
+  # @example
+  #   class myservice::config (
+  #     Enum[present, absent] $ensure = present,
+  #   ) {
+  #     file { '/etc/myservice':
+  #       ensure => ensure_directory($ensure),
+  #       mode   => '0755',
+  #     }
+  #   }
+  dispatch :ensure do
+    param 'Variant[String, Boolean]', :ensure_param
+  end
+
+  def ensure(ensure_param)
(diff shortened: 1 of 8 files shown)


pegasd:
Hm, doesn't seem like broken tests were caused by my changes.
Did something break on Travis / rspec side?

glennsarti:
@pegasd I'm seeing similar failures in my windows modules on Travis :-(  Suspect Travis just us a christmas present :-(

glennsarti:
@pegasd Tracking this in https://tickets.puppetlabs.com/browse/MODULES-6339

pegasd:
Can we retest this please? I'm assuming the issue has been fixed.

lib/puppet/functions/ensure_directory.rb
```diff
@@ -0,0 +1,37 @@
...
+  #   class myservice::config (
+  #     Enum[present, absent] $ensure = present,
+  #   ) {
+  #     file { '/etc/myservice':
+  #       ensure => ensure_directory($ensure),
+  #       mode   => '0755',
+  #     }
+  #   }
+  dispatch :ensure do
+    param 'Variant[String, Boolean]', :ensure_param
```
hlindberg: If you change the type to `Enum[present, absent]` you would not need to check that in an else since you are guaranteed your method gets called with only one of those two values.

hlindberg:
The original can be written: `$ensure_directory = if $ensure == present { directory } else { $ensure }`.
If you want to use these in your codebase consider that the functions are simple enough that they can be written in the puppet language - but do put them in a module of your own and namespace them.

A more useful general purpose function would be something like a simple switch/swap value like:
`swap_matching($ensure, present => directory)` that would swap the value in $ensure to 'directory' if value is 'present', else return the value of $ensure. This function would be comparable to using a case expression (you give the function a hash of matches instead of writing a case expression). I am however not convinced that such a function is that desirable since you have to learn that it exists and what it does - in contrast to `$ensure_directory = if $ensure == present { directory } else { $ensure }` which is general logic that you need to learn anyway.

Closing this for now since commits and PR also must be marked with the ticket number.
In any case, please open a Jira ticket for discussing the merits of adding this (the original, or the idea of having a general swap/switch) to stdlib.

PR #872, opened 2018-01-04
(MODULES-6366) Add data types for IP validation

ghoneycutt:
The types are from https://github.com/thrnio/puppet-ip which was released
under the Apache-2.0 license.


ghoneycutt:
I purposely left `Stdlib::Compat::Ip*` alone as it is meant to deal with the old `validate_*` functions. Because of this, the new code here does not mention it and has new tests separate from it so when it is time for the `Stdlib::Compat` bits to be taken out of the code, it can be done easily. I did however copy the test cases used in the associated types added here.

Test cases were also taken from the README in https://github.com/thrnio/puppet-ip. For the types that did not have tests, I used the examples from the RFC. https://www.ietf.org/rfc/rfc2373.txt

binford2k:
Hey @ghoneycutt! Could you add @thrnio as a co-author? (I'm sure you know how, but here's [docs](https://help.github.com/articles/creating-a-commit-with-multiple-authors/), if not.)

ghoneycutt:
I did not, thanks for teaching me!

Added Ryan to the commit.

b4ldr:
just mentioning https://github.com/puppetlabs/puppetlabs-stdlib/pull/843 as its similar.  I prefer this PR as its much more complete however i think we can strip out the `Address` word from all types to make readability easier e.g.

```
Stdlib::IP::Address -> Stdlib::IP::Address
Stdlib::IP::Address::v4 -> Stdlib::IP::v4
Stdlib::IP::Address::v6 -> Stdlib::IP::v6
```

ghoneycutt:
@binford2k Thanks for your +1. Could you approve or merge this PR? I would like to get a release going as it is a dependency for me to release other modules that need this functionality.

ghoneycutt:
I prefer with the `Address` portion as it is more exact. I'm happy to defer to whatever Puppet would like in order to get this merged.

b4ldr:
@ghoneycutt Im not sure what you mean by it being more exact.  if we want to be exact then we should really use something like

```
Stdlib::IP::4::Address   # ipv4
Stdlib::IP::6::Address   # ipv6
Stdlib::IP::50::Address  # ESP
```

i.e. the IP protocol should come before the address.  This becomes useful is you want to have something more then `Address` for instance

```
Stdlib::IP::4::Header   # ipv4
Stdlib::IP::6::Header   # ipv6
Stdlib::IP::50::Header  # ESP
```

however practically speaking i don't think puppet will ever need more then address as such having it in the type definition becomes superfluous and even if more are added it still makes a sane default.  what about having a pointer such that

```puppet
type Stdlib::IP = Stdlib::IP::Address
type Stdlib::IP::V4 = Stdlib::IP::Address::V4
type Stdlib::IP::V6 = Stdlib::IP::Address::V6
```

thrnio:
Just a note, `Stdlib::IP::4::Address` isn't valid because each component has to start with a letter, which is why the existing names all use `V4` or `V6`. `Stdlib::IP::V4::Address` would work though.

ghoneycutt:
I'm not attached to `Stdlib::IP::Address::x`, so it could be `Stdlib::IP::x` as you mentioned. Whatever Puppet wants to merge is fine with me.

Would not want to have an alias from one to the other as this adds unneeded complexity.

Puppet, what color should this shed be?

bodgit:
Any reason these have to be imported into Stdlib? If it's for the purposes of deprecating the original source module then I'd vote for importing in a way that needs the minimum of work for module authors to adapt to it, i.e. if it's just a case of prefixing everything with `Stdlib::` rather than having to shuffle things about.

eputnam:
@ghoneycutt just needs a rebase and we'll get this moving.

ghoneycutt:
Thanks @eputnam, got this rebased.

ghoneycutt:
Appreciate the merge @eputnam !

Can you please let us know here when a release is made that contains this PR? Will it be soon? Seems there hasn't been a release in quite some time relative to merged PR's.

PR #875, opened 2018-02-19
get rid of fixnum|bignum deprecation warning

tuxmea:
stdlib/lib/puppet/parser/functions/type3x.rb:25: warning: constant ::Fixnum is deprecated
stdlib/lib/puppet/parser/functions/type3x.rb:25: warning: constant ::Bignum is deprecated

### DIFF
--- a/lib/puppet/parser/functions/type3x.rb
+++ b/lib/puppet/parser/functions/type3x.rb
@@ -21,7 +21,7 @@ module Puppet::Parser::Functions
 
     klass = value.class
 
-    unless [TrueClass, FalseClass, Array, Bignum, Fixnum, Float, Hash, String].include?(klass) # rubocop:disable Lint/UnifiedInteger
+    unless %w[TrueClass FalseClass Array Bignum Fixnum Float Hash String].include?(klass.to_s)
       raise(Puppet::ParseError, 'type3x(): Unknown type')
     end
 


lib/puppet/parser/functions/type3x.rb
```diff
@@ -21,7 +21,7 @@ module Puppet::Parser::Functions
 
     klass = value.class
 
-    unless [TrueClass, FalseClass, Array, Bignum, Fixnum, Float, Hash, String].include?(klass) # rubocop:disable Lint/UnifiedInteger
+    unless %w[TrueClass FalseClass Array Bignum Fixnum Float Hash String].include?(klass.to_s)
```
eputnam: We should probably add Integer to this list since Bignum and Fixnum are indeed deprecated. When the deprecation warning is bypassed, we get the Puppet::ParseError instead because this function doesn't know what an Integer is.

tuxmea:
Hi Eric,

> On 25. Feb 2018, at 01:40, Eric Putnam <notifications@github.com> wrote:
> 
> @eputnam commented on this pull request.
> 
> In lib/puppet/parser/functions/type3x.rb:
> 
> > @@ -21,7 +21,7 @@ module Puppet::Parser::Functions
>  
>      klass = value.class
>  
> -    unless [TrueClass, FalseClass, Array, Bignum, Fixnum, Float, Hash, String].include?(klass) # rubocop:disable Lint/UnifiedInteger
> +    unless %w[TrueClass FalseClass Array Bignum Fixnum Float Hash String].include?(klass.to_s)
> 
> We should probably add Integer to this list since Bignum and Fixnum are indeed deprecated. When the deprecation warning is bypassed, we get the Puppet::ParseError instead because this function doesn't know what an Integer is.

We have that fix in place at a customer and it is working without the mentioned PareError.
Can you provide an example which generates the error?

And: why is your comment not visible at github issue?

Best,
Martin

PR #876, opened 2018-02-20
add remove_duplicate_whitspace function

bastelfreak:


bastelfreak:
Hey team. Please let me know if you're interested in this function. If so, I can try to add some tests (if desired/needed).

lib/puppet/parser/functions/remove_duplicate_whitespace.rb
```diff
@@ -0,0 +1,35 @@
...
+    raise(Puppet::ParseError, "remove_duplicate_whitespace(): Wrong number of arguments given (#{arguments.size} for 1)") if arguments.empty?
+
+    value = arguments[0]
+
+    unless value.is_a?(Array) || value.is_a?(String)
+      raise(Puppet::ParseError, 'remove_duplicate_whitespace(): Requires either array or string to work with')
+    end
+
+    result = if value.is_a?(Array)
+               value.map { |i| i.is_a?(String) ? i.gsub(/\s+/, ' ') : i }
```
ekohl: How about keeping the original whitespace char. Something like ```gsub(/(\s)\s+/, '\1')```

bastelfreak: Would that make a huge difference?

ekohl: I'm wondering if people like using tabs or perhaps tabs are needed somehow?

lib/puppet/parser/functions/remove_duplicate_whitespace.rb
```diff
@@ -0,0 +1,35 @@
...
+#  remove_duplicate_whitespace.rb
+#
+module Puppet::Parser::Functions
+  newfunction(:remove_duplicate_whitespace, :type => :rvalue, :doc => <<-DOC
+    This function replaces one or multiple whitespaces with a single one from a string or from
+    every string inside an array.
+
+    *Examples:*
+
+        strip("    aaa   bla")
```
binford2k: The name doesn't match the docs

bastelfreak:
I ruined my git history in this feature branch and cleaned it up.

lib/puppet/parser/functions/remove_duplicate_whitespace.rb
bastelfreak: fixed

hlindberg:
I think this could be a useful addition to the String formatting that is built into puppet - see [specification for String(String)](https://github.com/puppetlabs/puppet-specifications/blob/master/language/types_values_variables.md#string-to-string) - it has a %t format that trims left/right, but none that compacts internal whitespace. A %T could trim all spaces - so users can write `$result = String($input, '%T')`.

hlindberg:
There is already a `squeeze` function that can do this. Closing this.

PR #877, opened 2018-02-20
seeded_rand_string() function

pegasd:
This function is useful for generating consistent random strings (read "passwords") across different nodes.

### DIFF
--- a/lib/puppet/functions/seeded_rand_string.rb
+++ b/lib/puppet/functions/seeded_rand_string.rb
@@ -1,0 +1,29 @@
+# Generates a consistent random string of specific length based on (optionally) provided seed.
+#
+# @example Generate a consistently random string of length 8 with a seed:
+#   seeded_rand_string(8, "${module_name}::redis_password")
+#
+# @example Generate a random string from a specific set of characters:
+#   seeded_rand_string(5, '', 'abcdef')
+Puppet::Functions.create_function(:seeded_rand_string) do
+  # @param length Length of string to be generated.
+  # @param seed Seed string.
+  # @param charset String that contains characters to use for the random string.
+  #
+  # @return [String] Random string.
+  dispatch :rand_string do
+    param 'Integer[1]', :length
+    param 'String', :seed
+    optional_param 'Optional[String[2]]', :charset
+  end
+
+  def rand_string(length, seed, charset = nil)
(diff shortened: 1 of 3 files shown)


lib/puppet/functions/seeded_rand_string.rb
```diff
@@ -0,0 +1,29 @@
...
+#   seeded_rand_string(5, '', 'abcdef')
+Puppet::Functions.create_function(:seeded_rand_string) do
+  # @param length Length of string to be generated.
+  # @param seed Seed string.
+  # @param charset String that contains characters to use for the random string.
+  #
+  # @return [String] Random string.
+  dispatch :rand_string do
+    param 'Integer[1]', :length
+    param 'String', :seed
```
hlindberg: The text description states that seed is optional, but this declaration is for a required param.

hlindberg:
I think it is worth opening a ticket in puppet's Jira in the MODULES project to discuss the merit of this function. There are already other functions related to 'random' and there may be opportunity to do something about sprawling additions of "random this" and "random that"...

eputnam:
@hlindberg we already have an fqdn_rand() and fqdn_rand_string() but only seeded_rand() with no seeded_rand_string(). This method doesn't seem like sprawl with that in mind.

hlindberg:
`fqdn_rand_string` can be given a `seed` - how is this different? Why are so many different functions needed for almost the same functionality? Why not a simpler API and more intuitive names? Say `random_number` and `random_string` and then use named options to make those produce the different variations of using fqdn as seed or some other seed, length of string, an Integer or a Float etc. Come to think of it, even just `random` would work and then give it a data type that expresses a range - for example `random(Integer[1,100])`, `random(String[10])`.

pegasd:
@hlindberg There is currently no way to generate the same randomized string on two different hosts.
I do agree with your naming approach though. Let's pick whichever one is more suitable for this.

david22swan:
@pegasd Is there any movement on this PR?

pegasd:
@david22swan How should I proceed here? I'm open to suggestions. Or, if the community finds this insecure, I'm OK with closing this for good.

hlindberg:
Hm, I had objections to this, those concerns were not addressed. We could do much better than what was now merged. Suggest reverting this.

PR #880, opened 2018-03-03
Remove unneeded execute permission

smortex:
This set of commits do not have functional changes and remove execute permission and shebang on a set of tests that actually cannot be used as shell scripts.


david22swan:
@smortex Please rebase your changes onto the master

smortex:
Done, thanks for the head-up!

eputnam:
@smortex , see this commit https://github.com/puppetlabs/puppetlabs-stdlib/commit/09f892023c726eabaa85a308c2dbffd8e8b71fbd where they were added. Would you mind testing your changes with [rpmlint|https://linux.die.net/man/1/rpmlint] to see if it still complains?

smortex:
@eputnam do you have some pointers about building a RPM from a puppet module?

I could only find this [pull request](https://github.com/puppetlabs/puppetlabs-stdlib/pull/237/files), but when I build a package using this patch, `rpmlint(1)` only produce a few unrelated warnings.  If I revert some of my commits (to have an inconsistent state with executable files without shebang), I have the same result.  My guess is that this RPM building thing is not the one used by @averi when he proposed this change.

ping @averi in case you can help here 😉

eputnam:
@smortex ok, interesting, thanks for looking into that. Perhaps it isn't an issue any longer. Would be interested to see if @averi has any thoughts.

averi:
@smortex, @eputnam, we're badly lagging behind upstream in EPEL [1], I'll try to get the latest upstream release packaged later this week. Will you be making a new release based on the above fixes? Also can you confirm puppetlabs-stdlib is still compatible with Puppet 2.x series?

[1] https://src.fedoraproject.org/rpms/puppetlabs-stdlib/blob/master/f/puppetlabs-stdlib.spec

eputnam:
@averi stdlib still supports Puppet 2.7.20 and newer. let us know how the packaging goes. if there is no longer an issue, i'll get this merged and release asap.

averi:
rpmlint is now clean, feel free to merge, release so I can rebase the package to the latest available release. Thanks!

eputnam:
thanks @smortex and @averi !

averi:
@eputnam, would you mind double checking the 4.25.0 tag again? Seems it doesn't contain all the commits we were expecting (@smortex's ones on top) and as a consequence the Github generated tarball still doesn't contain the relevant execute permissions and shebangs fixes.

eputnam:
@averi yeah this went in just after the 4.25.0 tag. I'll see what I can do about a patch release.

averi:
@eputnam, any update on this? will a minor release happen to include the aforementioned patches?

eputnam:
@averi yeah we're working on it currently. you can actually follow the forge RSS feed to get an update as soon as it's out: https://forge.puppet.com/puppetlabs/stdlib/rss

PR #881, opened 2018-03-05
(MODULES-6332) - PDK convert

pmcmaw:
Firstly running pdk convert to make the module pdk compliant.

Second commit is disabling Style/HashSyntax around code within the spec/spec_helper.rb.

Reason: stdlib is a special case where we enforce the usage of hash_rockets, this avoids the module becoming backwards incompatible. The spec_helper.rb is written by a pdk template so even if I implemented hash rockets on the next `pdk update` run it will be overwritten. This is something to consider before the next `pdk update` is ran and on the next run these changes will be reverted.


.rubocop_todo.yml
```diff
@@ -0,0 +1,4 @@
+Style/HashSyntax:
+  EnforcedStyle: hash_rockets
```
pmcmaw: This is required to avoid making the module backwards compatible.

.rubocop_todo.yml
```diff
@@ -0,0 +1,4 @@
...
+RSpec/NamedSubject:
+  Enabled: false
```
pmcmaw: Unsure how ok this is?
It used to be disabled and there are a serious amount of violations of this cop. It would result in some files being completely rewritten.

spec/spec_helper.rb
```diff
@@ -1,9 +1,32 @@
...
-  require 'spec_helper_local'
+  require 'spec_helper_local' if File.file?(File.join(File.dirname(__FILE__), 'spec_helper_local.rb'))
 rescue LoadError => loaderror
-  puts "Could not require spec_helper_local: #{loaderror.message}"
+  warn "Could not require spec_helper_local: #{loaderror.message}"
+end
+
+include RspecPuppetFacts
+
+# rubocop:disable Style/HashSyntax
```
pmcmaw: This is a work around that will need to be resolved, the change will be overwritten on the next `pdk update`.

.rubocop_todo.yml
HelenCampbell: I would say this is fine for now. Plus it won't get lost if it's in the todo, so seems perfect solution for the meantime.

spec/spec_helper.rb
pmcmaw: I have raised an internal bug for tracking this and it has been added onto the backlog.
We will need to come up with some type of solution going forward.

PR #882, opened 2018-03-05
Support 'relative' paths for Stdlib::Unixpath type.

fatmcgav:


ripienaar:
this fundamentally changes the behaviour of the type, it would be a major breaking change so something I'd imagine only be shipped in stdlib 5.

fatmcgav:
Whilst I agree this is a behaviour change, I'm not sure I'd consider it a fundamental one... 

The type is called 'Unixpath', and a relative path is still a valid unix path. 

A guess the actual breaking change will be in 'Stdlib::Absolutepath', which variants this type... 

I'm happy to defer to PL though...

david22swan:
@fatmcgav Please alter the docs to reflect your changes and add some test coverage.

b4ldr:
I dont like that this would change the bahaviour of `Stdlib::Absolutepath`.  If there needs to be types for relative paths then i think more types are needed to preserver current behaviour perhaps

```puppet
Stdlib::Path
Stdlib::Path::Absolute
Stdlib::UnixPath
Stdlib::UnixPath::Absolute
Stdlib::WindowsPath
Stdlib::WindowsPath::Absolute
```

spec/type_aliases/unixpath_spec.rb
```diff
@@ -11,6 +11,7 @@
         /var/ネット
         /var//tmp
         /var/../tmp
+        ./dir
```
b4ldr: what relative dirs of the form `relative/dir`

fatmcgav:
@david22swan / @b4ldr Yeh, the change in behaviour in `Stdlib::Absolutpath` is undesirable at best, and as @ripienaar pointed out is a change in the [documented](https://github.com/puppetlabs/puppetlabs-stdlib/#stdlibabsolutepath) behaviour. 

I like the idea of creating additional types that do enforce `Absolute`ness, so I'll rework this PR...

eputnam:
thanks for your work here @fatmcgav ! seems like it would be good to have types for relative and absolute forms of unix and windows paths. then we could use those absolute forms in absolutepath.

david22swan:
@fatmcgav Sorry to bother you but I was wondering whether this PR was still active???

david22swan:
@fatmcgav Apologies, but due to the amount of time that this PR has been left to lie fallow and your lack of response when prompted on it I feel that I must now close it.
Thank you for the work that you put into it and if you ever decided to return to it, or open a separate PR, we will of course be more than happy to review it again.
Best Wishes
The Modules Team

PR #885, opened 2018-03-09
(MODULES-6771) - Updates to README.

pmcmaw:
These changes includes:
1.  The removal of 'Unacceptable Examples' to avoid confusion
2. The removal of hard line breaks
3. The addition of 'valid values'
4. Changing IP Addresses wrapped in back ticks to wrap them in single quotes

These changes are being PR'd directly against the release branch as these are required before the module can be released.

### DIFF
--- a/README.md
+++ b/README.md
@@ -407,7 +407,7 @@ C:\\
 \\\\host\\windows
 ```
 
-Unacceptable input example:
+Valid values: A windows filepath.
 
 #### `Stdlib::Filesource`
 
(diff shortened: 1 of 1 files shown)


pmcmaw:
Hey @jbondpdx,

I have requested your review as these are the changes that we were chatting about last night. Hopefully they are up to scratch. The last few entries I have added the URL to the page for examples, hoping this is sufficient. 👍

README.md
```diff
@@ -795,48 +628,40 @@ Examples:
...
 ```
 
+Valid values: An IPv6 address.
+
 #### `Stdlib::IP::Address::Nosubnet`
 
-Match the same things as the `Stdlib::IP::Address` alias, except it will not
-match an address that includes an address prefix (e.g., it will match
-`192.168.0.6` but not `192.168.0.6/24`).
+Match the same things as the `Stdlib::IP::Address` alias, except it will not match an address that includes an address prefix (e.g., it will match '192.168.0.6' but not '192.168.0.6/24').
```
jbondpdx: e.g. is hard for translation; use 'for example' instead please

jbondpdx: I see some other ones, so I think you could just do a search and replace.

PR #887, opened 2018-03-13
WIP - (maint) - Removing duplicates from template and .sync.yml

pmcmaw:


pmcmaw:
@HelenCampbell these seem to be the only options I can remove without any introducing failures in travis for acceptance tests.

PR #890, opened 2018-03-16
Allow pick() to work with strict variables

binford2k:
If `strict_variables` is enabled, then the standard usage of `pick()`
will fail. This simply adds the calling convention that strings that
look like variables will be resolved if they exist and ignored if they
don't.

### DIFF
--- a/lib/puppet/parser/functions/pick.rb
+++ b/lib/puppet/parser/functions/pick.rb
@@ -24,13 +24,12 @@ module Puppet::Parser::Functions
 
 DOC
              ) do |args|
-    args = args.compact
-               
-    # look up the values of any strings that look like '$variables'
-    args.map! do |item|
-      next unless item.is_a? String
+    # look up the values of any strings that look like '$variables' w/o mutating args
+    args = args.map do |item|
+      next item unless item.is_a? String
       item.start_with?('$') ? call_function('getvar', [item.slice(1..-1)]) : item
     end
+    args.compact!
     args.delete(:undef)
     args.delete(:undefined)
     args.delete('')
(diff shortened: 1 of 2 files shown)


lib/puppet/parser/functions/pick.rb
```diff
@@ -15,9 +15,22 @@ module Puppet::Parser::Functions
...
+      $real_jenkins_version = pick('$::jenkins_version', '1.449')
+
 DOC
              ) do |args|
-    args = args.compact
+
+    # look up the values of any strings that look like '$variables'
+    args.map! do |item|
+      next unless item.is_a? String
+      item.start_with?('$') ? function_getvar([item.slice(1..-1)]) : item
```
hlindberg: Use `call_function` instead of calling with `function_xxx` as that only works for 3.x functions.

lib/puppet/parser/functions/pick.rb
```diff
@@ -15,9 +15,22 @@ module Puppet::Parser::Functions
...
+    variable exists.
+
+      $real_jenkins_version = pick('$::jenkins_version', '1.449')
+
 DOC
              ) do |args|
-    args = args.compact
+
+    # look up the values of any strings that look like '$variables'
+    args.map! do |item|
```
hlindberg: Should absolutely not mutate the given args!! Must work on a copy. earlier the `args.compact` provided the copy now this logic will destroy the caller's array of arguments.

david22swan:
@binford2k This is currently failing unit and syntax test's

hlindberg:
*sigh* - I think this should be reverted as the change is not backwards compatible.

mwhahaha:
revert https://github.com/puppetlabs/puppetlabs-stdlib/pull/927

mwhahaha:
Simple test to exhibit the issue is:

$g = pick('$something', '$somethingelse')

Will fail with,
Error: Evaluation Error: Error while evaluating a Function Call, pick(): must receive at least one non empty value at /home/aschultz/foo.pp:1:6 on node aschultz.localdomain

hlindberg:
For those that want to avoid errors with strict variables, there are other functions to use besides `pick`. For example the `getvar` function which will return `undef` for a variable that does not exist - simply combine that with `pick`.

PR #896, opened 2018-03-22
(MODULES-6894) - Puppet 5.5.0 function deprecation

pmcmaw:
https://puppet.com/docs/puppet/5.5/release_notes.html#new-features

The following functions have been implemented in puppet 5.5.0. As these functions in puppet take precedence over stdlib functions. We are deprecating these functions and addressing additional failures introduced with the release.

1. `flatten`
2. `empty`
3. `join`
4. `keys`
5. `values`
6. `length`

### DIFF
--- a/README.md
+++ b/README.md
@@ -1163,7 +1163,7 @@ Converts the case of a string or of all strings in an array to lowercase.
 
 #### `empty`
 
-**Deprecated. Puppet 5.5.0 has introduced a built in `empty` function, which will take precedence over the stdlib function. See [Puppet 5.5.0 Release Notes](https://puppet.com/docs/puppet/5.5/release_notes.html#new-features).**
+**Deprecated:** This function has been replaced with a built-in [`empty`](https://docs.puppet.com/puppet/latest/function.html#empty) function as of Puppet 5.5.0.
 
 Returns `true` if the argument is an array or hash that contains no elements, or an empty string. Returns `false` when the argument is a numerical value.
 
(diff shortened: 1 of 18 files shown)


lib/puppet/functions/length.rb
```diff
@@ -11,4 +11,5 @@ def length(value)
     end
     result
   end
+  Puppet.deprecation_warning('The length function is deprecated. This function is now shipped with Puppet in versions 5.5.0 and later. For more information please see Puppet 5.5.0 Release Notes.')
```
hunner: These functions should only be deprecated for users using puppet >=5.5.0, not for users running puppet <5.5.0. So it shouldn't print the deprecation warnings on puppet <5.5.0.

Users running >=5.5.0 will not be able to call these functions without specially trying, so I don't think printing a deprecating warning for >=5.5.0 will actually be helpful.

So we probably don't need to print them at all, and thus don't need to test for them.

README.md
```diff
@@ -1163,6 +1163,8 @@ Converts the case of a string or of all strings in an array to lowercase.
 
 #### `empty`
 
+**Deprecated. Puppet 5.5.0 has introduced a built in `empty` function, which will take precedence over the stdlib function. See [Puppet 5.5.0 Release Notes](https://puppet.com/docs/puppet/5.5/release_notes.html#new-features).**
```
hunner: The functions should only be deprecated for users of puppet >=5.5.0, but not for users of puppet <5.5.0. We can remove them from stdlib entirely when we drop compatibility for puppet <6.0.0 from stdlib in 2050 or whatever :P.

PR #899, opened 2018-04-02
(MODULES-6951) Updating translations for readmes/README_ja_JP.md

ehom:
Manually checking this file in since triggering it from TX did not work.


ehom:
Wrong branch. Should be a PR to the release branch.

PR #902, opened 2018-04-18
(MODULES-6881) - Removing duplication in .sync.yml

pmcmaw:


.sync.yml
```diff
@@ -8,9 +8,9 @@
   secure: ""
   branches:
     - release
-  extras:
-    - env: CHECK=release_checks
-      rvm: 2.1.9
+extras:
```
pmcmaw: This section is required. When removing it, it runs release_checks against Ruby 2.4.1 instead of Ruby 2.1.9.

PR #903, opened 2018-04-19
(MODULES-6881) - Removing duplication in .sync.yml

pmcmaw:
```extras:
  - env: CHECK=release_checks
    rvm: 2.1.9
```

release_checks is included in the pdk template however it runs against ruby 2.4.1 instead of ruby 2.1.9. With this config, release_checks runs against ruby 2.1.9.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,68 +1,42 @@
 ---
-sudo: false
-dist: trusty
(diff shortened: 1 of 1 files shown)


pmcmaw:
Incorrect, restarting PR.

PR #905, opened 2018-04-24
(MODULES-7024) Add 20-octet MAC addresses

ananace:
Not quite sure if the best thing is to extend the MAC matcher, but since the 20-octet hardware address is the exact same format it feels like a good start.

### DIFF
--- a/lib/puppet/parser/functions/is_mac_address.rb
+++ b/lib/puppet/parser/functions/is_mac_address.rb
@@ -14,6 +14,7 @@ module Puppet::Parser::Functions
     mac = arguments[0]
 
     return true if %r{^[a-f0-9]{1,2}(:[a-f0-9]{1,2}){5}$}i =~ mac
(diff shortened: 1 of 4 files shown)


pmcmaw:
Closing and reopening to kick tests from an updated Travis test configuration that is now in puppetlabs-stdlib master.

pmcmaw:
Would it be possible to add a test to confirm that this functionality works as expected. An example of testing this can be found here: https://github.com/puppetlabs/puppetlabs-stdlib/blob/master/spec/acceptance/is_mac_address_spec.rb so if you extend them to include a 20-octet MAC address. 

We would also like to thank you for taking time to contribute, it is much appreciated.

ananace:
Just to throw a ping at this, are there any examples for testing an actual type - and not a function? Or is it just a case of it being enough to test the is_mac_address function?

pegasd:
@ananace There are! Check out `spec/type_aliases` directory - this is the best approach for testing custom types imo.

PR #907, opened 2018-05-04
Update documentation of reject() function to mention filter() function

hlindberg:
People finding the reject function may need to perform other types of rejection than using a regexp (which only works on strings).
For that reason it is worth including a small example using filter().


hlindberg:
AppVeyor says there are trailing whitespace - cannot see those in this file...

alexjfisher:
Lines 16 and 19?

hlindberg:
This is superseded by #922 - closing this.

PR #909, opened 2018-05-16
(MODULES-7181) Remove Stdlib::(Ipv4|IPv6|Ip_address)

baurmatt:
There was 3 different data types for ip validation before this commit.
This removes the Stdlib::(Ipv4|IPv6|Ip_address) data type in favour of
Stdlib::IP::* as those are way more complete.


b4ldr:
as the one that introduced those types, i agree the we should only have one interface for this  [and the compat one].  The `Stdlib::IP::Address` types are much nicer then the ones i introduced so we should keep them

pegasd:
The upcoming 5.0.0 release could be a good time to introduce these backwards-incompatible changes.

pmcmaw:
@david22swan when merging can you ensure to squash and merge. There is a typo in my second commit. (Pressed enter and submitted by accident)

HelenCampbell:
Thank you for the PR @baurmatt !

PR #912, opened 2018-05-25
[FM-6953] Removal of OS support on stdlib

david22swan:


david22swan:
related pr: https://github.com/puppetlabs/ci-job-configs/pull/4537

metadata.json
```diff
@@ -38,30 +38,26 @@
...
       "operatingsystemrelease": [
-        "7",
         "8",
         "9"
       ]
     },
     {
       "operatingsystem": "Ubuntu",
       "operatingsystemrelease": [
-        "14.04",
```
b4ldr: how come ubuntu 14.04 being removed, its not EOL until April 2019[1]

[1]https://wiki.ubuntu.com/Releases

davinhanlon: Hi @b4ldr. Ubuntu 14.04 is no longer supported as a master platform by Puppet. We are going through a process for adding Ubuntu 18 support to modules that are supported on Ubuntu, and in preparation for that are removing support for Ubuntu 14.04. Our pipelines will continue to test against Ubuntu 16. We're not making any actual code changes, so it should still work on 14.04. Let me know if you've any concerns. Thanks.

b4ldr: ok thanks for the explanation

hunner: @b4ldr Perhaps of interest to you, we changed to keeping 14.04 based on Puppet *agent* platform support rather than Puppet master platform support (so may be more along the April 2019 timeline that you mentioned).

PR #913, opened 2018-05-30
(maint) - Cause dirname to fail when passed undef

b4ldr:
currently the dirname function returns `'.'` if passed `undef`.  this PR causes dirname to fail in this instance, it will also cause dirname to fail if an empty string is passed.

test before patch
```bash
# puppet apply -e 'notify{dirname(undef): }'                                          
Notice: Compiled catalog for jobo-5685.local in environment production in 0.04 seconds
Notice: .
Notice: /Stage[main]/Main/Notify[.]/message: defined 'message' as '.'
Notice: Applied catalog in 0.02 seconds
```

test after patch
```bash
# puppet apply -e 'notify{dirname(undef): }' 
Error: Evaluation Error: Error while evaluating a Function Call, dirname(): Requires a none empty string (line: 1, column: 8) on node jobo-5685.local
```


lib/puppet/parser/functions/dirname.rb
```diff
@@ -16,6 +16,10 @@ module Puppet::Parser::Functions
     unless arguments[0].is_a?(String)
       raise(Puppet::ParseError, 'dirname(): Requires string as argument')
     end
+    # undef is converted to an empty string ''
+    if arguments[0].empty?
+      raise(Puppet::ParseError, 'dirname(): Requires a none empty string as argument')
```
pmcmaw: can you update:

`none empty` to `non-empty` and I will get your change merged.

b4ldr:
@pmcmaw i some how messed up the rebase so i have created a new clean PR here https://github.com/puppetlabs/puppetlabs-stdlib/pull/921

pmcmaw:
Thats great! Ill have a look at it, when it passes all tests I will work on getting it merged. Thank you for the quick response.

PR #918, opened 2018-06-12
Load https file into loadjson() and loadyaml()

jonnytdevops:
This functionality is useful if one needs to interface their puppet manifest with external systems that expose hashes via JSON or YAML over HTTP(S).


david22swan:
Running through adhoc: https://jenkins-master-prod-1.delivery.puppetlabs.net/view/modules/view/linux/view/_adhoc/view/stdlib/

PR #925, opened 2018-06-19
(maint) - Skip sort when puppet agent is version 6 or greater

pmcmaw:
This function will be present in core puppet from version 6.0.0 and later.


pmcmaw:
Nightly adhoc tests have passed.

PR #927, opened 2018-06-22
Revert "Allow pick() to work with strict variables"

mwhahaha:
This reverts commit 2a493adcd01ece80cb70fab004e2fb8b01261775.

This change was not backwards compatible. If a user was using pick with
a string that include a $, it would be evaluated and could throw an
error.

See https://bugs.launchpad.net/tripleo/+bug/1778201


mwhahaha:
@hlindberg @david22swan @binford2k please review

PR #928, opened 2018-06-22
Allow pick() to handle strings starting with $

binford2k:
Escape a prefixed $ with a \ to prevent it from being evaluated.


david22swan:
Adhoc Run: https://jenkins-master-prod-1.delivery.puppetlabs.net/view/modules/view/cross-platform/view/_adhoc/view/stdlib/

david22swan:
Ran green on adhoc

pmcmaw:
Closing and reopening to rekick Appveyor for testing purposes.

hlindberg:
I still have the same concerns I had on June 25th, and think this should not be merged.

hlindberg:
My concern was simply that the PR is not backwards compatible since someone may actually have strings that start with `$`, and if this is merged, then they must change their code to escape a leading `$`.  If this change is wanted anyway (risk is probably low) then by all means go head and merge.

A backwards compatible way could be to add a function that returns the first defined variable with a value other than undef, say `first_defined_var()` - your long example would then be:
```
pick(first_defined_var('foo_var1', 'foo_var2', 'foo_var3', 'foo_var4'),  'default')
```

I looked at `getvar` and it cannot be extended to return first defined, since it now supports a default value - hence proposing adding a new function that simply checks for first defined variable name not having the value `undef` and returning `undef` if none of the variable names were defined. 

I doubt the use case is important/frequent enough to warrant syntax, but if it was, then maybe something like `$[foo_var1, foo_var2, foo_var3, foo_var4]` could mean the same as `first_defined_var(...)` - but for me a function is better since it describes what is does in a way that punctuation does not.

binford2k:
the `first_defined_var()` idea seems workable. I'm not a fan of magically overloading strings in the first place. If you think that's the best way, then I'll write the function.

I'd still really like a way to specify that a variable was optional too (for something like the following use case that is still super common), but that's a separate conversation.

``` puppet
if $virtual? {
    # do stuff for virtual machines and don't blow up if the fact doesn't exist and strict_variables is on
}
```

hlindberg:
For the last one - isn't the recommendation to always use`$facts` ?
```
if $facts['virtual'] { }
```
Or, if wanting just one formula that works also for digging (puppet >= 6.0.0):
```
if $facts.get('virtual')
if $facts.get('os.family')
```

I do think it is best to write a new function - the name I suggested may not be the best (`first_defined_var()`), but I have no better suggestion atm.  I did consider `pick_var` since that name makes sense for someone used to `pick`, but it looks a bit silly with `pick(pick_var('a', 'b', 'c'), 'default')`. You could make this the "new pick" though, but it requires doing something special for either the default value or the list of variables - for example, using a hash for options `pick_var('a', 'b', 'c',  default => 'default')`, or by using a marker value - `pick_var('a', 'b', 'c', default, 'default')` (the value given after a literal `default` is the default value, or "last value is the (required) default" `pick_var('a', 'b', 'c', 'default')`. Just some thoughts...

PR #931, opened 2018-06-29
(maint) Add deprecation call to `sprintf_hash`

hlindberg:
The functions was already documented as being deprecated but an actual
call was missing. This commit adds a call.

Note that the sprintf_hash function is a 4.x function and therefore uses
call_function (all other functions use the deprecated `function_<name>`
convention. They also send Ruby Symbol as the name of the function.


david22swan:
@hlindberg Currently failing a syntax check

hlindberg:
Some nasty style nazi thing going on there… I am on my third attempt now trying to satisfy the rules…

- henrik

PR #933, opened 2018-07-25
5.0.0 - Release Prep

david22swan:


david22swan:
@clairecadman Could you take a look at the summary for this changelog and tell me what you think??

pmcmaw:
@david22swan  You will want this against release.

david22swan:
Changes made

pmcmaw:
I have updated your PR title, can you update the summary to be like the following:
'This is a major release which removes support for Scientific 5 and Debian 7. This also contains a lot of work around updating functions that have now been migrated into Puppet itself. This release will be used as a pinning point as there are plans to remove support for all puppet versions less than 4 in the next release.'

david22swan:
updated

PR #949, opened 2018-08-24
add Function extname()

cocker-cc:


lib/puppet/parser/functions/extname.rb
```diff
@@ -0,0 +1,18 @@
+#
+# extname.rb
+#
+module Puppet::Parser::Functions
```
hlindberg: Please implement using the 4.x function API and namespace the function to stdlib

lib/puppet/functions/extname.rb
```diff
@@ -0,0 +1,25 @@
+# Returns the Extension (the Portion of Filename in Path starting from the
+# last Period).
+#
+# If Path is a Dotfile, or starts with a Period, then the starting Dot is not
+# dealt with the Start of the Extension.
+#
+# An empty String will also be returned, when the Period is the last Character
+# in Path.
+
+Puppet::Functions.create_function(:extname) do
```
hlindberg: please make that `:'stdlib::extname'` to place it in the `stdlib` namespace.
Then, the examples need to be updated to use `stdlib::extname(...)`

cocker-cc: Done.  
Also I moved it into `functions/stdlib/` according to the [Documentation](https://puppet.com/docs/puppet/4.10/functions_ruby_overview.html#location).
I adapted also `README.md`, although this one might be auto-generated.

PR #950, opened 2018-08-28
Update docs for 'concat' to be correct

rhowe-gds:
Looks like there was a copy & paste error in the documentation for concat.
Fix up the example to be correct.


tphoney:
Lovely catch. great work !!!

PR #954, opened 2018-09-08
(MODULES-7768) Handle nil in delete_undef_values() function

hlindberg:
PUP-9112 changed use of `:undef` inside structured values to instead
using `nil` in Puppet 6.0.0. The `delete_undef_values()` function
was not prepared to handle this and would not delete `nil` from
`Array` or `Hash` values.

This commit fixes this problem.

All 3.x functions in stdlib were reviewed for the same problem, and only `delete_undef_values()` had this problem.

### DIFF
--- a/lib/puppet/parser/functions/delete_undef_values.rb
+++ b/lib/puppet/parser/functions/delete_undef_values.rb
@@ -30,9 +30,10 @@ module Puppet::Parser::Functions
     end
     result = args[0].dup
     if result.is_a?(Hash)
-      result.delete_if { |_key, val| val.equal? :undef }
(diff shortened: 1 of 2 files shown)


MikaelSmith:
`pick` doesn't? Oh, I guess `args.compact` takes care of it.

davidmalloncares:
Hey @hlindberg / @MikaelSmith - the Modules team should really be involved in these changes to stdlib as maintainers of the repo - we have no visibility of the changes, or had a chance to review and feedback - esp around any related changes to acceptance or unit tests that may be needed to prevent the module going red in our pipelines

PR #955, opened 2018-09-20
Fix `pick` function docs

alexjfisher:
In puppet, (since version 4 or 3 with future parser), the only values
that are falsey are `undef` and `false`.


tphoney:
cheers @alexjfisher  !

PR #961, opened 2018-09-26
(FM-7388) Fix failing tests on Puppet 6

hlindberg:
This fixes failing tests for Puppet 6.0.0. There are several causes:

* arguments to functions may be copies of given values - mocking must take that into account
* the rspec_puppet method of calling `subject.call` for a 4x function does not work - the `subject.execute` method should be used instead (the `call` method requires a scope and ends up taking given arguments as the scope - with certain failure as a consequence when called with just arguments).
* it is an error do declare user defined resources that do not exist
* Handling of `undef` is different in Puppet 6.0.0 - calling functions from Ruby will now get the same transformation of arguments as if the call came from the puppet language.
* Tests were marked as pending expecting Puppet to not raise errors for illegal arguments - that problem is fixed with the current combination of rspec_puppet and puppet

This PR also has a few maintenance changes - adding parentheses around calls.

See individual commits for more details regarding the changes made to each failing test.

### DIFF
--- a/spec/functions/abs_spec.rb
+++ b/spec/functions/abs_spec.rb
@@ -12,6 +12,10 @@ describe 'abs', :if => Puppet::Util::Package.versioncmp(Puppet.version, '6.0.0')
   end
 
   describe 'signature validation in puppet4', :if => RSpec.configuration.puppet_future do
+    before(:each) do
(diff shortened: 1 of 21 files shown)


tphoney:
closed in favour of #962

PR #966, opened 2018-11-01
(MODULES-8219) - Fix to unit test's to account for Puppet 5.5.7

david22swan:
https://tickets.puppetlabs.com/browse/MODULES-8219

### DIFF
--- a/spec/functions/deprecation_spec.rb
+++ b/spec/functions/deprecation_spec.rb
@@ -11,7 +11,7 @@ if Puppet::Util::Package.versioncmp(Puppet.version, '4.5.0') >= 0
     it { is_expected.to run.with_params.and_raise_error(ArgumentError) }
 
     it 'displays a single warning' do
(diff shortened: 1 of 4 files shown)


tphoney:
@hlindberg is this expected behaviour 5.X vs 6.X ?

hlindberg:
@tphoney In Puppet >= 6.0.0 undef / nil values in nested values (arrays/hashes) are not transformed (to empty string or `:undef` symbol. Also, if function returns data with `:undef` that gets transformed to nil / undef. This change was merged into Puppet 5.5.7, but was reverted in 5.5.8 because of too much breakage found in functions with illegal method definitions inside of the functions. I think that the reverted logic also reverted that changes to transformation of '' and :undef - but I don't remember without looking.

david22swan:
Updated change's to account for 5.5.8 reverts

pmcmaw:
@david22swan Can you update this PR to link to a ticket with more info?
This will avoid confusion with the fact changes have been made in 5.5.7 and then reverted.

PR #953, opened 2018-09-07
fix IPAddress type to allow /0 CIDR as this is valid

jameseck:
Currently, if you use the Stdlib::IP::Address type, it fails to validate 0.0.0.0/0 as a valid IP.  This PR fixes that.

PR #975, opened 2018-12-04
(MODULES-8322) Consider IPs with /0 as valid

simondeziel:
0.0.0.0/0 and ::/0 should be considered valid as they are useful to represent the default route.

This is what people expect as seen in:

https://github.com/zehweh/puppet-netplan/commit/164b0a0575da1634ba47b09f478dbec95d78517f
https://github.com/puppetlabs/puppetlabs-stdlib/pull/953 (which is superseded by this pull request)


david22swan:
screen shot 2019-01-10 at 4 52 52 pm

david22swan:
@simondeziel Thank you for the work you did :)

PR #978, opened 2018-12-18
(MODULES-8137) - Addition of support for SLES 15

david22swan:


david22swan:
screen shot 2018-12-18 at 3 52 27 pm

PR #981, opened 2018-12-23
(MODULES-8404) - Relax `Stdlib::Filesource` type

alexjfisher:
The previous regex only supported the special `modules` `puppet:///`
mount point.  Custom mount points should also be accepted.

https://puppet.com/docs/puppet/5.5/file_serving.html#whats-a-mount-point-in-a-puppet-uri

### DIFF
--- a/spec/type_aliases/filesource_spec.rb
+++ b/spec/type_aliases/filesource_spec.rb
@@ -3,9 +3,30 @@ require 'spec_helper'
 if Puppet::Util::Package.versioncmp(Puppet.version, '4.5.0') >= 0
   describe 'Stdlib::Filesource' do
     describe 'valid handling' do
-      ['https://hello.com', 'https://notcreative.org', 'https://canstillaccepthttps.co.uk', 'http://anhttp.com', 'http://runningoutofideas.gov', 'file:///hello/bla', 'file:///foo/bar.log',
-       'puppet:///modules/foo/bar.log', 'puppet://pm.example.com/modules/foo/bar.log', 'puppet://192.0.2.1/modules/foo/bar.log', '/usr2/username/bin:/usr/local/bin:/usr/bin:.', 'C:/', 'C:\\',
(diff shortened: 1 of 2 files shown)


spec/type_aliases/filesource_spec.rb
```diff
@@ -25,8 +46,6 @@
           '\\Users/hc/wksp/stdlib',
           'C:noslashes',
           '\\var\\tmp',
-          'puppet:///foo/bar.log',
-          'puppet:///pm.example.com/modules/foo/bar.log',
```
alexjfisher: At first glance this might not look valid, but I can't see anything that says `pm.example.com` couldn't be a valid mount point.  Unlikely and unusual, yes, but not technically invalid.

spec/type_aliases/filesource_spec.rb
```diff
@@ -3,9 +3,30 @@
 if Puppet::Util::Package.versioncmp(Puppet.version, '4.5.0') >= 0
   describe 'Stdlib::Filesource' do
     describe 'valid handling' do
-      ['https://hello.com', 'https://notcreative.org', 'https://canstillaccepthttps.co.uk', 'http://anhttp.com', 'http://runningoutofideas.gov', 'file:///hello/bla', 'file:///foo/bar.log',
-       'puppet:///modules/foo/bar.log', 'puppet://pm.example.com/modules/foo/bar.log', 'puppet://192.0.2.1/modules/foo/bar.log', '/usr2/username/bin:/usr/local/bin:/usr/bin:.', 'C:/', 'C:\\',
-       'C:\\WINDOWS\\System32', 'C:/windows/system32', 'X:/foo/bar', 'X:\\foo\\bar', '\\\\host\\windows', '//host/windows', '/var/tmp', '/var/opt/../lib/puppet'].each do |value|
+      [
```
bastelfreak: oh thanks for reformatting

alexjfisher: I know it creates noise, but it really was pretty hard to read as was.

PR #1003, opened 2019-03-07
Add a stdlib::ip_in_range() function

iglov:
It's very useful function from https://github.com/deanwilson/puppet-ip_in_range . Why you haven't include it so far?

### DIFF
--- a/lib/puppet/functions/stdlib/ip_in_range.rb
+++ b/lib/puppet/functions/stdlib/ip_in_range.rb
@@ -1,0 +1,30 @@
+# Returns true if the ipaddress is within the given CIDRs
+#
+# @example ip_in_range(<IPv4 Address>, <IPv4 CIDR>)
+#   stdlib::ip_in_range('10.10.10.53', '10.10.10.0/24') => true
+Puppet::Functions.create_function(:'stdlib::ip_in_range') do
+  # @param [String] ipaddress The IP address to check
+  # @param [Variant[String, Array]] range One CIDR or an array of CIDRs
+  #   defining the range(s) to check against
+  #
+  # @return [Boolean] True or False
+  dispatch :ip_in_range do
+    param 'String', :ipaddress
+    param 'Variant[String, Array]', :range
+    return_type 'Boolean'
+  end
+
+  require 'ipaddr'
+
+  def ip_in_range(ipaddress, range)
+    ip = IPAddr.new(ipaddress)
+
+    if range.is_a? Array
+      ranges = range.map { |r| IPAddr.new(r) }
+      ranges.any? { |rng| rng.include?(ip) }
+    elsif range.is_a? String
+      ranges = IPAddr.new(range)
+      ranges.include?(ip)
+    end
+  end
+end
--- a/README.md
+++ b/README.md
@@ -1399,6 +1399,17 @@ stdlib::extname('.profile')      => ''
 
 *Type*: rvalue.
 
+#### `stdlib::ip_in_range`
+
+A Puppet function to determine if an IPv4 address is within the IPv4 CIDR. Returns true if the ipaddress is within the given CIDRs.
+
+```puppet
+$ranges = ['192.168.0.0/24', '10.10.10.0/24']
+$valid_ip = stdlib::ip_in_range('10.10.10.53', $ranges) # $valid_ip == true
+```
+
+*Type*: rvalue.
+
 #### `fact`
 
 Return the value of a given fact. Supports the use of dot-notation for referring to structured facts. If a fact requested does not exist, returns Undef.
--- a/spec/functions/ip_in_range_spec.rb
+++ b/spec/functions/ip_in_range_spec.rb
@@ -1,0 +1,17 @@
+require 'spec_helper'
+
+describe 'stdlib::ip_in_range' do
+  describe 'signature validation' do
+    it { is_expected.not_to eq(nil) }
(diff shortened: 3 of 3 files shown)


hlindberg:
I think this needs to be refactored to use the modern (4.x) function API, and that the function should be namespaced to `stdlib` to avoid potentially clashing with similar function in other modules (namespacing cannot be done in the 3.x API).
There is also the need to clean up the commits.

Others need to chime in on the usefulness of the function and if it should be added to stdlib in the first place. I will make some review comments meanwhile.

lib/puppet/parser/functions/ip_in_range.rb
```diff
@@ -0,0 +1,57 @@
...
+    unless Puppet::Parser::Functions.function('is_ip_address')
+      raise Puppet::ParseError, 'ip_in_range(): requires the puppetlabs-stdlib functions'
+    end
+
+    # 2nd arg can be an array so normalise it into one array for processing
+    arguments.flatten!
+
+    # error out if any of the arguments given are not valid IP addresses
+    arguments.each do |ipaddress|
+      unless function_is_ip_address([ipaddress])
```
hlindberg: This is old API that does not work with modern functions - use `call_function(name, ...)` instead as that can call both 3x and 4x functions.

lib/puppet/parser/functions/ip_in_range.rb
```diff
@@ -0,0 +1,57 @@
...
+             ) do |arguments|
+
+    require 'ipaddr'
+
+    unless arguments.length == 2
+      raise ArgumentError, "ip_in_range(): wrong number of arguments (#{arguments.length} must be 2)"
+    end
+
+    # load in the ip validation function from puppetlabs-stdlib
+    unless Puppet::Parser::Functions.function('is_ip_address')
```
hlindberg: This is old API and does not work with modern functions. Suggest simply calling the function and catch the error that results if it is not there.

lib/puppet/parser/functions/ip_in_range.rb
```diff
@@ -0,0 +1,57 @@
...
+    or you can pass an array of CIDR ranges and return true if the IP is inside
+    any of them.
+
+    $ranges = ['192.168.0.0/24', '10.10.10.0/24']
+    $valid_ip = ip_in_range('10.10.10.53', $ranges) # $valid_ip == true
+
+    # or use it in a conditional
+
+    if ip_in_range($ipaddress, '10.10.200.0/24') == true {
+      notify { 'Do subnet based resources here': }
```
hlindberg: In the modern API, the documentation is written using YARDOC in a comment that goes above the function. That is much better as the documentation text is then not loaded at runtime (waste of memory and processing power).

lib/puppet/parser/functions/ip_in_range.rb
hlindberg: Also, do you actually need to typecheck the input first since you would be getting an error from `IPAddr.new`

hlindberg:
This documentation may be of value: https://puppet.com/docs/puppet/6.3/functions_refactor_legacy.html

iglov:
Hey, @hlindberg ! Thank u, for ur revision, i appreciate it! I've done some fixes. Could u plz check it and says what should i doing next?
P.S. How i should clear commits? Create new pull request with 1 commit? :)

lib/puppet/functions/stdlib/ip_in_range.rb
```diff
@@ -0,0 +1,29 @@
...
+  #
+  # @return [Boolean] True or False
+  dispatch :ip_in_range do
+    param 'String', :ipaddress
+    param 'Variant[String, Array]', :range
+    return_type 'Boolean'
+  end
+
+  def ip_in_range(ipaddress, range)
+    require 'ipaddr'
```
hlindberg: calling `require` every time the function is called is unnecessary - better to call this in the main body - for example move this to line 15. That way it is only called once.

lib/puppet/functions/stdlib/ip_in_range.rb
```diff
@@ -0,0 +1,29 @@
+# Returns true if the ipaddress is within the given CIDRs
+#
+# @example ip_in_range(<IPv4 Address>, <IPv4 CIDR>)
+#   stdlib::ip_in_range('10.10.10.53', '10.10.10.0/24') => true
+Puppet::Functions.create_function(:'stdlib::ip_in_range') do
+  # @param [String] ipaddress Get an IP address
```
hlindberg:
```suggestion
  # @param [String] ipaddress The IP address to check
```

lib/puppet/functions/stdlib/ip_in_range.rb
```diff
@@ -0,0 +1,29 @@
...
+  # @param [Variant[String, Array]] range Get a one or many ranges of IP addresses
```
hlindberg:
```suggestion
  # @param [Variant[String, Array]] range One CIDR or an array of CIDRs defining the range(s) to check against
```

lib/puppet/functions/stdlib/ip_in_range.rb
iglov: got it!

iglov:
Hey, @hlindberg ! Thank a lot, ur so helpful! Now, how i think, i totally done with this task :)

eimlav:
Hi @iglov . Thanks for your contribution (and @hlindberg for reviewing). This change looks good to me. Cheers!

PR #1004, opened 2019-03-08
(MODULES-8732) Cron types

Rathios:
This is a feature request for cron types in stdlib.

I'd like to reason there is a need for good cron datatypes in stdlib that raise the bar when it comes to validating cron parameters.

Some points:

- The cron resource is built-in and is a use case in all Puppet code
- Modules already dependent on stdlib will have these types available without having to depend on other third party modules
- Modules sometimes don't fully support the same datatype combinations the `cron` resource does, eg `Variant[Integer[n,n], String[1], Array[Integer[n,n], String[1]]]`, nor do most of them bother to validate the values of the fields (quite understandably)
- Lacking validation of field values is sometimes made up for in spec tests, when a spec test for each cron type could instead handle invalid values more simply

I've found there are already good cron types in modules like [Vox Pupuli's cron](https://github.com/voxpupuli/puppet-cron/tree/master/types). It would probably be a good idea to not ignore lessons learned while creating them.

This obviously needs a bit more testing, but it works well on vixie cron (CentOS 7) to start. It would be good to find out if `\g` is supported in regexes on all Puppet versions that support types.

### DIFF
--- a/spec/type_aliases/cron_hour_spec.rb
+++ b/spec/type_aliases/cron_hour_spec.rb
@@ -1,0 +1,70 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe 'Stdlib::Cron::Hour' do
+  describe 'accept integers' do
+    [0, 1, 2, 15, 23].each do |value|
+      it { is_expected.to allow_value(value) }
+    end
+  end
+  describe 'accept integers as strings' do
+    ['0', '1', '2', '15', '23'].each do |value|
+      it { is_expected.to allow_value(value) }
+    end
+  end
+  describe 'accept ranges' do
+    ['0-23', '0-5', '18-23', '5-11', '19-22'].each do |value|
+      it { is_expected.to allow_value(value) }
+    end
+  end
+  describe 'accept step values' do
+    ['0-23/2', '0-23/24', '9-15/3'].each do |value|
+      it { is_expected.to allow_value(value) }
+    end
+  end
+  describe 'accept wildcards' do
+    ['*', '*-7', '*/7'].each do |value|
+      it { is_expected.to allow_value(value) }
+    end
+  end
+  describe 'accept comma-separated integers, ranges and step values' do
+    ['0,1,2,15,23', '0-5,11-23', '0-11/2,12-23/4', '0,2-5,12-23/2'].each do |value|
+      it { is_expected.to allow_value(value) }
+    end
+  end
+  describe 'accept arrays of valid expressions' do
+    [
+      [0, 3, 6, 9, 12, 15, 18, 21],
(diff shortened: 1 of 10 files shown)


david22swan:
@Rathios The code look's good, a little hesitant since it's taken from a vox module but it should be fine. It could use some unit test's however to cover the use of the types, if you could add them I think I would be happy to merge.

Rathios:
@david22swan I wrote these from scratch before I thought to look for a cron module, which is perhaps even less assuring. Thinking it might be good to get them tested "in the field" on a wider range of OSes? I have limited opportunity myself beyond Ubuntu, CentOS and RHEL though. That aside, they should be fairly solid.

Will get some spec tests up soon.

david22swan:
@Rathios Apologies, I thought they were simply taken from the module you listed. I'm sorry if I offended you.

Rathios:
Oh, no offense taken whatsoever. :) I just meant to point out that they weren't pulled from any modules and so haven't been tested particularly widely. I've been a bit busy but will get tests added soon.

pegasd:
Hey there, @Rathios!
I’ve been maintaining my own Cron type aliases for a while here: https://github.com/pegasd/puppet-cron

Feel free to borrow the specs: https://github.com/pegasd/puppet-cron/tree/master/spec/type_aliases

pegasd:
Any progress here?

I also wanted to note that, since `puppet-strings` supports type aliases now, it'd be great to remove newlines between the comment and the type definition. Otherwise `puppet-strings` considers those types undocumented.

lionce:
Hello @Rathios ,

Please let us know if you're still working on this PR and if you need help. Thanks! 

Cheers!

Rathios:
Surprise! I'm not dead. The rest of 2019 just went insane.

@pegasd Thanks! Those spec tests helped me get going with my own. :) I've also corrected the newline issue you mentioned concerning proper support for puppet-strings.

I could use some feedback on whether or not to allow leading zeroes. Cron has no issue with them, but due to the potential for bugs in Puppet-related code (such as https://tickets.puppetlabs.com/browse/MODULES-7786) I decided to remove support for them.

These tests should pass and if they do, please let me know how this is looking. I'd appreciate some help testing this more extensively before it's merged, just because it's stdlib.

carabasdaniel:
Hi @Rathios,

Thank you for your contribution, this looks really great!

This would be better suited to the core cron module: https://github.com/puppetlabs/puppetlabs-cron_core

If you submit it there, please ping me on the PR so I can make sure it gets merged quickly.

michaeltlombardi:
This PR has been quiet for several months and belongs best in the puppetlabs/puppetlabs-cron_core project - I am closing this for now.

PR #1005, opened 2019-03-09
add Stdlib::Syslogfacility type

bastelfreak:


types/syslogfacility.pp
```diff
@@ -0,0 +1,25 @@
+type Stdlib::Syslogfacility = Enum[
+  'user',
```
alexjfisher: There's also a `kern` facility.  I can't remember why I left it out of the nrpe data type.  Probably because it wouldn't make sense (or be illegal) to log nrpe with facility code 0 (not that many of the others would make much sense either).

bastelfreak: updated it

types/syslogfacility.pp
```diff
@@ -0,0 +1,26 @@
...
+  'solaris-cron',
+  'local0',
+  'local1',
+  'local2',
+  'local3',
+  'local4',
+  'local5',
+  'local6',
+  'local7',
+  'kern'
```
alexjfisher: For consistency, probably best to put this first in the list, (so they're all in facility code order).
https://en.wikipedia.org/wiki/Syslog#Facility

sheenaajay:
Hi @bastelfreak ,Hasn't seen any activity in this PR, do you think you'll be able to work on it in the near future? No pressure, just checking in.

alexjfisher:
@sheenaajay Is there any extra work required?  I don't think my nitpick should be a blocker.

sheenaajay:
Thank you @alexjfisher . 
Hi @bastelfreak , Could you please resolve the conflicts and rebase with the latest master to run the acceptance tests.Thank you.

bastelfreak:
@sheenaajay I rebased it

sheenaajay:
Screen Shot 2019-09-18 at 09 26 16

sheenaajay:
Thank you for submitting the PR @bastelfreak.

sheenaajay:
Great work...

PR #1006, opened 2019-03-12
Add more ensure types

runejuhl:
Adds the following:

+ `Stdlib::Ensure::File`
+ `Stdlib::Ensure::Package`
+ `Stdlib::Ensure::PackageVersion`

### DIFF
--- a/README.md
+++ b/README.md
@@ -305,6 +305,69 @@ Unacceptable input example:
 ../relative_path
 ```
 
+#### `Stdlib::Ensure::File`
+
+Matches acceptable ensure values for file resources.
+
+Acceptable input examples:
+
+```shell
+present
+absent
+file
+directory
+link
+```
+
+Unacceptable input example:
+
+```shell
+true
+false
+```
+#### `Stdlib::Ensure::Package`
+
+Matches acceptable keyword ensure values for package resources. Does **not**
+match package versions; see `Stdlib::Ensure::PackageVersion`.
+
+Acceptable input examples:
+
+```shell
+present
+installed
+absent
+purged
+held
+latest
+```
+
+Unacceptable input example:
+
+```shell
+true
+false
+"v0.2.4"
+```
+#### `Stdlib::Ensure::PackageVersion`
+
+Matches acceptable ensure values for package resources, plus strings and
+`SemVer`.
+
+Acceptable input examples are the ones in `Stdlib::Ensure::Package` and:
+
+```shell
+"v2.5.7"
+"0.2.1"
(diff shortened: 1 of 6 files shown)


hlindberg:
There is a problem with having these in stdlib as what they describe is defined by Puppet and may be different in different puppet versions.
We have discussed adding the data type `Like` to puppet as that would enable describing that you want a type that is like the type of some known definition  - for example `Like[file, ensure]`

runejuhl:
@hlindberg that would be better, but seeing as that feature hasn't been implemented, would it not be alright to add these types to Stdlib in the meantime?

The downside to not having them in a widely used library, like Stdlib, is that you miss out on some configurability. It's a common pattern to take a parameter like `Enum['present', 'absent'] $config_file_ensure` (like in [`postgresql::client`](https://github.com/puppetlabs/puppetlabs-postgresql/blob/master/manifests/client.pp#L3)), but it's rare to see all the file `ensure` options spelled out -- like in `postgresql::client` it's often just `present` or `absent`.

A better example might be the `voxpupuli-puppet` module: here the `ensure` in the `Package[python]` resource comes from `::puppet` which defines it like this:

```puppet
  Enum['absent', 'present', 'latest'] $ensure     = $python::params::ensure,
```

This means that it's not possible to pin Python against a particular version without jumping through hoops, even if it would have been possible if `Stdlib::Ensure::PackageVersion` had been used.

hlindberg:
Agree that it is better to have them done right statically in something like stdlib that having multiple bad/limiting implementations all over the place. Will let others decide on the merits of adding the suggested data types to stdlib - my point was just that these values do change with Puppet version even if it is not frequently.

runejuhl:
Ah, gotcha. Yes, that is something to keep in mind :+1:

sheenaajay:
Thanks, @runejuhl for submitting the PR. Apologize the delay.
Please check if the later releases of puppet solve the issue.

michaeltlombardi:
Since this hasn't seen any activity in about four months, I'm going to go ahead and close this pull request for the time being. Please re-open this pull request once the next actions are addressed, new information is available, or you have a question related to this pull request. We've become aware of difficulties re-opening pull requests, in the event you cannot please mention `puppetlabs/modules`  with an @ sign in front and we'll re-open this pull request.

Closing the pull request doesn't mean we don't consider this change valuable, just that there are things that need to be addressed before it can be merged. If you have any questions or concerns, please don't hesitate to ping us in slack.puppet.com.

alexjfisher:
Looks like something very similar was merged in https://github.com/puppetlabs/puppetlabs-stdlib/pull/1129

PR #1014, opened 2019-04-16
(MODULES-8949) add stdlib::has_value() function

chadh:
as the subject says.  Adds ruby's `has_value()` functionality.

### DIFF
--- a/lib/puppet/functions/stdlib/has_value.rb
+++ b/lib/puppet/functions/stdlib/has_value.rb
@@ -1,0 +1,23 @@
+# Returns whether or not a value is present in a hash
+#
+# Note that this is a relatively expensive operation versus a key lookup
+#
+
+Puppet::Functions.create_function(:'stdlib::has_value') do
+  # @param h the hash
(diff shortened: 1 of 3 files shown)


lib/puppet/functions/stdlib/has_value.rb
```diff
@@ -0,0 +1,23 @@
+# Returns whether or not a value is present in a hash
+#
+# Note that this is a relatively expensive operation versus a key lookup
+#
```
smortex: Should'nt this information be in / be copied to the `README.md` file?

chadh: added.  thanks.  btw, henrik noted that we can just do:

```
'value' in $hash.values
```

to accomplish this, so maybe we don't need it.  I'm not sure if that discussion should be here or in the ticket.  The requestor was testing that method.

chadh:
Since there is a relatively easy way to accomplish this in the language, I am going to close the PR.

PR #1017, opened 2019-05-08
(MODULES-9049) Add type alias for 'yes' and 'no'.

ghoneycutt:


ghoneycutt:
https://tickets.puppetlabs.com/projects/MODULES/issues/MODULES-9049

ghoneycutt:
Could someone please take a look, thank you!

hlindberg:
I think this is fine - it would be nice to make them case independent which could be done by using a `Pattern` data type with a regexp where the case independent flag is turned on.
For example:
```
type YesNo = Pattern[/\A(?i:(yes|no))\z/]
```
This works hand in hand with `Boolean($val)` since it turns a `yes/no` into boolean in the same case independent way.

ghoneycutt:
Good idea Henrik. I'll get the PR amended.

ghoneycutt:
@hlindberg look good?

tphoney:
Looks good to me, Merging. 

Thanks for the PR @ghoneycutt

PR #1022, opened 2019-05-22
(FM-8048) Add RedHat 8 support

eimlav:


eimlav:
Experimental adhoc with RedHat 8 ran successfully:

Screen Shot 2019-05-22 at 13 16 48

PR #1023, opened 2019-05-22
Limit the maximum array size produced by range().

mbaynton:
It is trivially easy to accidentally DOS a puppet master by requesting
a range larger than was intended, for example
`range('host00/index.html', 'host25/index.html')`. Because jruby's
implementation of unrolling the range is parallelized, this sort of
thing takes down the whole box.

The solution here is a simple sanity-check on it; we could get fancier
with a new argument specifying a larger size etc. if someone thinks it
needed.

### DIFF
--- a/lib/puppet/parser/functions/range.rb
+++ b/lib/puppet/parser/functions/range.rb
@@ -80,7 +80,7 @@ module Puppet::Parser::Functions
             when '...' then (start...stop) # Exclusive of last element
             end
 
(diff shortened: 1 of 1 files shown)


mbaynton:
This broke our 24-core 128gb puppetmaster today, for what it's worth.

tphoney:
Hi @mbaynton thanks for the fix, i think this is a sensible change. Merged to master ! 🎆

PR #1024, opened 2019-05-23
Add localhost to defined Type Stdlib::Host

cocker-cc:


alexjfisher:
Is this actually needed?  I think `Stdlib::Fqdn` will already allow `localhost`.

cocker-cc:
`Stdlib::Fqdn` does not accept `localhost`, as there is no dot in `localhost`.

alexjfisher:
There's no dot in `example`, but according to the tests, that is a valid `Stdlib::Fqdn`
https://github.com/puppetlabs/puppetlabs-stdlib/blob/3310305e2cecb44e986c95706f512099829ec733/spec/type_aliases/fqdn_spec.rb#L6

cocker-cc:
Ayea, indeed You're right.  I must have overseen something…  Sorry for the noise.  I'll close this one.

PR #1031, opened 2019-06-20
(FM-8230) Convert testing to litmus

eimlav:

PR #1032, opened 2019-06-20
(MODULES-9432) Remove Style/HashSyntax rubocop rule

eimlav:
This rule is used to enforce pre Ruby 1.9 hash syntax:

	{:a => 2}

Since the module no longer supports Puppet versions which correspond
with anything pre Ruby 1.9, this should be removed to enforce the modern
hash syntax:

	{a: 2, b: 1}

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -2,9 +2,5 @@ RSpec/NamedSubject:
   Enabled: false
 Lint/ScriptPermission:
   Enabled: false
-Style/HashSyntax:
-  Exclude:
-  - spec/spec_helper.rb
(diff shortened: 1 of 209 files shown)


joshcooper:
I don't think we should do this, because puppetserver 5.x uses JRuby 1.7 which is language compatible with MRI 1.9. So installing this module on puppetserver 5/PE 2018.1 will cause compilation to fail. See alsoMODULES-6139 and MODULES-8695

eimlav:
Hi @joshcooper thanks for the info. For now I'll disable the rule in the affected block in #1031 and we can look at this issue another time. Cheers.

PR #1036, opened 2019-07-18
Minor typo.

vStone:


REFERENCE.md
```diff
@@ -16,7 +16,7 @@ the stdlib class, and should not be declared independently.
 
 **Functions**
 
-* [`abs`](#abs): **Deprected:** Returns the absolute value of a number
+* [`abs`](#abs): **Deprecated:** Returns the absolute value of a number
```
igalic: i wonder if this will need to be fixed here instead: https://github.com/puppetlabs/puppetlabs-stdlib/blob/master/lib/puppet/parser/functions/abs.rb#L7

vStone: It probably will :D. Having generated files in git always confuses me ;)

PR #1037, opened 2019-07-18
Fix typo in abs function

vStone:
Will need an update in REFERENCE.md too.


tphoney:
Hi @vStone thanks for the PR, it is now merged.

PR #1038, opened 2019-07-26
add stdlib::ensure_at_least function

kreeuwijk:
Return the package version for a given package that is at least the given minimum version
The function will compare the currently installed package with the given minimum version and:
 - return the minimum version if the installed package is less than the minimum version, or not present
 - return the installed version if the installed package is the same or newer than the minimum version

### DIFF
--- a/lib/puppet/functions/ensure_at_least.rb
+++ b/lib/puppet/functions/ensure_at_least.rb
@@ -46,7 +46,12 @@ Puppet::Functions.create_function(:ensure_at_least) do
     ].compact.join(' ')
 
     queryresult = call_function('puppetdb_query', query)
-    if queryresult.length > 1
+    case queryresult.length
+    when 0
+      result = { 'version' => '0.0.0' }
+    when 1
+      result = queryresult.first
+    when 2..Float::INFINITY
       if provider.nil?
         # Assume first PDB query result is for the default provider
         provider = queryresult.map { |elem| elem['provider'] }.first
@@ -54,17 +59,16 @@ Puppet::Functions.create_function(:ensure_at_least) do
       # Assume last PDB query result is always the highest version
       result = queryresult.select { |elem| elem['provider'] == provider }.last
     else
-      result = queryresult.first
+      result = { 'version' => '0.0.0' }
     end
+
     installed = result['version']
 
-    if installed
-      case call_function('versioncmp', installed, minversion)
-      when 0, 1
-        installed
-      when -1
-        minversion
-      end
+    case call_function('versioncmp', installed, minversion)
+    when 0, 1
+      installed
+    when -1
+      minversion
     else
       minversion
     end
--- a/lib/puppet/functions/ensure_at_least.rb
+++ b/lib/puppet/functions/ensure_at_least.rb
@@ -6,11 +6,11 @@
 #
 # @example Example usage:
 #   # Ensure at least version 2.7.5-77.el7_6 of python is installed
-#     package { 'python':
-#       ensure => ensure_at_least('python', '2.7.5-77.el7_6')
-#     }
+#   package { 'python':
+#     ensure => ensure_at_least('python', '2.7.5-77.el7_6')
+#   }
 #
-Puppet::Functions.create_function(:ensure_at_least, Puppet::Functions::InternalFunction) do
+Puppet::Functions.create_function(:ensure_at_least) do
   # @param package
   #   The package to ensure
   #
@@ -19,7 +19,6 @@ Puppet::Functions.create_function(:ensure_at_least, Puppet::Functions::InternalF
   #
   # @return [String] package version value.
   dispatch :ensure_at_least do
-    scope_param
     required_param 'String', :package
     required_param 'String', :minversion
     return_type 'String'
--- a/lib/puppet/functions/ensure_at_least.rb
+++ b/lib/puppet/functions/ensure_at_least.rb
@@ -6,9 +6,16 @@
 #
 # @example Example usage:
 #   # Ensure at least version 2.7.5-77.el7_6 of python is installed
-#   package { 'python':
-#     ensure => ensure_at_least('python', '2.7.5-77.el7_6')
-#   }
+#     package { 'python':
+#       ensure => ensure_at_least('python', '2.7.5-77.el7_6')
+#     }
+#
+# @example To specify a specific provider:
+#   # Ensure at least version 2.7.5-77.el7_6 of python is installed
+#     package { 'openssl':
+#       ensure   => ensure_at_least('openssl', '2.1.2', 'gem'),
+#       provider => gem
+#     }
 #
 Puppet::Functions.create_function(:ensure_at_least) do
   # @param package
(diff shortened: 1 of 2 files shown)


kreeuwijk:
Some more use cases that need writing additional logic:
- multiple versions of the package resource are already present
- multiple package providers available for the package and user wants to use a specific one

lib/puppet/functions/ensure_at_least.rb
```diff
@@ -0,0 +1,72 @@
...
+#     }
+#
+# @example To specify a specific provider:
+#   # Ensure at least version 2.7.5-77.el7_6 of python is installed
+#     package { 'openssl':
+#       ensure   => ensure_at_least('openssl', '2.1.2', 'gem'),
+#       provider => gem
+#     }
+#
+Puppet::Functions.create_function(:ensure_at_least) do
```
reidmv: We should namespace this function to `stdlib::ensure_at_least`.

Most of the functions in stdlib aren't namespaced because they predate the capability; namespacing functions wasn't possible until Puppet 4.

Some functions arguably made sense without a stdlib namespace. Basic string manipulations, comparisons... things that should probably just be part of core Puppet really. E.g. `join()`.

This function is a niche use case that's effectively working around a deficiency in Puppet's ability to express desired state of a Package. It wouldn't make sense to have this in Puppet core, because if we were going to put something in core it would be more integrated into the Package type instead. As such, while this is a powerful tool and useful perhaps to have in Stdlib, I think it makes sense to namespace it so that it is clear where it comes from and that it is an add-on functionality.

lib/puppet/functions/ensure_at_least.rb
```diff
@@ -0,0 +1,72 @@
...
+      "certname = '#{trusted['certname']}'",
+      "and package_name = '#{package}'",
+      ("and provider = '#{provider}'" unless provider.nil?),
+      '}',
+    ].compact.join(' ')
+
+    queryresult = call_function('puppetdb_query', query)
+    if queryresult.length > 1
+      if provider.nil?
+        # Assume first PDB query result is for the default provider
```
reidmv: We should not assume. If there are multiple results returned and no provider specified, we should raise an error.

Better to fail catalog compilation than to possibly mis-configure a system.

lib/puppet/functions/ensure_at_least.rb
```diff
@@ -0,0 +1,72 @@
...
+        provider = queryresult.map { |elem| elem['provider'] }.first
+      end
+      # Assume last PDB query result is always the highest version
```
reidmv: Let's also not assume here. For something in stdlib, I want the logic to be certain, or to fail gracefully when necessary.

lib/puppet/functions/ensure_at_least.rb
```diff
@@ -0,0 +1,72 @@
...
+      result = queryresult.select { |elem| elem['provider'] == provider }.last
+    else
+      result = queryresult.first
+    end
```
reidmv: What about when `queryresult == nil`?

That'll cause a problem later when we try to `result['version']`. Need to account for the query coming back with nothing.

lib/puppet/functions/ensure_at_least.rb
kreeuwijk: Following that line of reasoning, it could also be delivered as a separate module on the Forge and not as part of stdlib. It remains debatable if it should still be namespaced yes or no in that case, as a separate module would ensure that it is seen as add-on functionality and not core. I personally would like to see as short a name as possible for the function, which namespacing wouldn't help with. I don't we can use `init.rb` for a function to give it the name of the module, right?

lib/puppet/functions/ensure_at_least.rb
kreeuwijk: While I could replicate the logic from the package provider to determine the standard provider when none is specified (based on the OS of the node), it would yield worse results compared to letting PuppetDB handle it. For example, PuppetDB will return packages from the chocolatey provider without issue, even though this is not the default provider on Windows. If we used traditional OS-based logic to determine the default provider when none is specified, the wrong provider would be selected in more cases. Selecting the first record from the PuppetDB results will ensure a valid provider is always chosen.

Multiple results can also come from multiple package versions being installed side-by-side, all from the same provider. That should not cause a compilation error.

lib/puppet/functions/ensure_at_least.rb
kreeuwijk: This could be replaced by recursively calling `versioncmp` on all the filtered results, which should also yield the highest version output, at the expensive of more cpu cycles. I looked through my own Package Inventory and found that PuppetDB already performs this sorting on its own, which is why selecting the last record is a cheaper method of getting the highest version. Hence one can argue wether re-comparing the records against each other is necessary, as the records are already provided by PuppetDB in the correct sorting.

lib/puppet/functions/ensure_at_least.rb
kreeuwijk: Good point, this was previously handled by `if installed` but that logic no longer catches that case as we always set `installed = result['version']` now. I tested and indeed the code fails when running the function on a non-existent package. I will edit the code to handle this.

lib/puppet/functions/ensure_at_least.rb
kreeuwijk: The PuppetDB team has confirmed that no intentional sorting is happening, it seems to be coming from Postgres' internals at the moment. The suggestion is to leverage `Gem::Version` of that is available. I will have a look at it.

lib/puppet/functions/ensure_at_least.rb
kreeuwijk: The PuppetDB team has confirmed that no intentional sorting is happening, it seems to be coming from Postgres' internals at the moment. The suggestion is to leverage `Gem::Version` of that is available. I will have a look at it.

lib/puppet/functions/ensure_at_least.rb
kreeuwijk: The PuppetDB team has confirmed that no intentional sorting is happening, it seems to be coming from Postgres' internals at the moment. No logic exists in PuppetDB on default/primary providers for a resource, this all happens outside of PuppetDB. I will have a look at pulling in the default provider logic from the package resource provider, with some additional logic to fallback to the record returned from PuppetDB in safe scenarios.

lib/puppet/functions/ensure_at_least.rb
reidmv: Not init.rb, no, but from a technical standpoint, a namesake function can be defined in a module as e.g.

```
ensure_at_least/lib/puppet/functions/ensure_at_least.rb
```

That said, I think a longer name in stdlib would have a higher likelihood of adoption and better maintenance. Because of the specificity of the use case I don't agree that a short name is particularly desirable. As such, I would favor the longer name in stdlib over a shorter name in a standalone module for this function specifically.

As I said before, I agree that short names for some functions _is_ desirable – just not for this one.

kreeuwijk: Agreed, I'll move it to `stdlib::ensure_at_least`

lib/puppet/functions/ensure_at_least.rb
kreeuwijk: Default provider logic replicated from the Puppet's package resource providers. Enriched with auto-selecting providers in known-safe scenarios.

lib/puppet/functions/ensure_at_least.rb
kreeuwijk: Implemented recursive version comparison of installed versions for the (auto-)selected provider, ensuring only the highest version is ever returned.

florindragos:
Since stdlib is a core module, we would really need to add some tests to this.
Also, another concern (as already discussed) is the location of this new feature. Are we certain that this is the right place for it? If we add it here and people start using it, it would be more difficult to move.

reidmv:
For reference: the (very old) feature request ticket this function would be a workaround for is [PUP-1519](https://tickets.puppetlabs.com/browse/PUP-1519).

michaeltlombardi:
As far as I can tell, this change effects only `package` and does so very specifically at that. This implies, to me, it belongs in core puppet with the package type and providers.

It also does not have any tests at this time.

michaeltlombardi:
Closing per comments above in favor of this change being merged into the puppet itself as it only supports the package type and provider.

PR #1042, opened 2019-08-20
FM-8411 - add support for debian10

lionce:


lionce:
https://pipelines.puppet.com/team-modules/builds/582162

PR #1048, opened 2019-09-16
(MODULES-9915) Add type aliases for cloud object store uris

hooten:
Closes https://tickets.puppetlabs.com/browse/MODULES-9915

Add a type alias for Cloud Object Store URIs, e.g., (s3://mybucket/myfile).

These URIs are commonly used as sources for files, packages, etc. in Puppet modules. Some organizations even keep all sources in a specific cloud object store like s3 in order to limit downloads from the internet when applying a Puppet catalogue across potentially thousands of nodes.

It would be useful to have a type alias for these URIs.


hooten:
I'd appreciate a review when you get a chance @lionce or @tphoney. Thanks!

tphoney:
Great work, a nice addition.

hooten:
Thanks!

PR #1063, opened 2019-11-05
puppet_settings: add additional facts

b4ldr:
This PR adds addition ssl facts from the puppet settings, specify:
  :localcacert,
  :ssldir,
  :hostpubkey,
  :hostprivkey,
  :hostcert,

any advice on hot to improve the spec tests welcome and also happy to add additional settings.

Also im not sure if removing `unit/facter/puppet_settings.rb` and the rescue fix where still needed,  they seemed redundant in my config but happy to add back. 

TODO:  im curious how to access the settings from different sections.  for instance our puppet masters have a different ssldir configurations in the master and agent sections

EDIT: config from specific sections can be fetched using the following; however i'm not sure if its general enough to add facts using this to stdlib?

```ruby
section = :master
setting = :ssldir
Puppet.settings.values(Puppet[:environment].to_sym, section).print(setting)
```


florindragos:
Hello @b4ldr 
Thanks for your contribution.
My concern with this addition is following: I'm not sure this is the right place to add these facts.
Since they are not used by this module, wouldn't it make more sense to have them inside a module that uses them?

b4ldr:
@florindragos thanks for the comments however this is true for all of the facts defined in the stdlib module.  For my own needs i have indeed put theses in a [separate module ](https://github.com/wikimedia/puppet/blob/production/modules/wmflib/lib/facter/puppet_config.rb), however they seemed like the would be useful in the general case and stdlib or puppet core seems to me to be the best place for things that are generally useful.

Of course others may not find theses useful at all and this PR can get rejected but im not sure they would make more sense going into a different module.  Of course im open to suggestions if there is a more appropriate module

DavidS:
@b4ldr the extlib module (https://github.com/voxpupuli/puppet-extlib) has been created to collect more experimental/niche features. Due to its niche nature it has a lower barrier to things like this that add overhead to all agents. If it turns out to be successful there (lots of folks start depending on it), we can consider pulling it into the stdlib.

PR #1067, opened 2019-11-21
(WIP) Add volatile_fact() function

binford2k:
This is a quick POC to add a function that resolves facts on the agent
side during catalog application. It's useful only when using cached
catalogs that depend on facts that might change over time.

This was accomplished by slightly modifying `fact()` so that it could be
deferred and then writing a quick Puppet code function to wrap it in a
`Deferred` object. If reviewers agree that this functionality is useful,
then I'll port it to a Ruby fact so the name doesn't have to be
namespaced and it can parallel the existing `fact()` function.

You would use it like so:

```
notify { 'testing':
  message  => stdlib::volatile_fact('some.volatile.fact'),
}
```

### DIFF
--- a/lib/puppet/functions/fact.rb
+++ b/lib/puppet/functions/fact.rb
@@ -31,7 +31,7 @@ Puppet::Functions.create_function(:fact) do
   end
 
   def fact(fact_name)
-    facts = closure_scope['facts']
+    facts = closure_scope['facts'] || Facter.to_hash
 
     # Transform the dot-notation string into an array of paths to walk. Make
     # sure to correctly extract double-quoted values containing dots as single
--- a/functions/volatile_fact.pp
+++ b/functions/volatile_fact.pp
@@ -1,0 +1,3 @@
+function stdlib::volatile_fact(String $fact_name) >> Deferred {
+  Deferred("fact", [$fact_name])
+}


binford2k:
@hlindberg will this do what I think it does?

DavidS:
I really do not like the hiding of the Deferred wrapping. Hiding this (rather obscure) incantation from people will only lead to all kinds of troubles down the line as people will not understand what is happening. Introducing this pattern into the ecosystem ("hiding `Deferred` in a function") means that inexperienced folks suddenly need to check every function call they encounter on when/where the function is actually executing.

If you really insist on making a wrapper function to save users a couple of keystrokes, it should be called `deferred_fact()` to make that part of it clear.

hlindberg:
@binford2k If your intent is to get the value of a fact on the agent side and that you want Facter to run to get all facts on each access - then yes.

Agree with @DavidS about naming, `volatile_fact` seems like a new concept - that is just confusing. IMO - since you would have to learn what to do in order to get the value of a fact on the agent you can just as well learn how to use `Deferred` directly since that is a general piece of knowledge that you can apply to many different things. I have a general dislike for the flea market that is created when having small "helpful" wrappers.

BTW: aren't facts already available on the agent side (I don't remember)? If not, and you are using the modified version of `fact` in this PR it would run Facter for every call since it is not remembering that it ran Facter - would be very slow.

lib/puppet/functions/fact.rb
```diff
@@ -31,7 +31,7 @@ def to_dot_syntax(array_path)
   end
 
   def fact(fact_name)
-    facts = closure_scope['facts']
+    facts = closure_scope['facts'] || Facter.to_hash
```
joshcooper: If facts aren't in scope (which are supposed to be the facts submitted by the agent), then aren't we're going to use the puppetserver's facts (?) which is bad.

binford2k:
I don't think I explained the use case very well. This is only useful in the case of _cached_ catalogs that need _current_ facts. Imagine that you had an application running on an ec2 instance and that the application needs to be configured with its current external address, even if you stop/start the machine and that address changes. You're running _cached catalogs_.

In current Puppet, that's impossible to do afaik, without resorting to exceedingly gross hacks like this. Which is fragile and probably wouldn't work anyway :trollface: 

``` puppet
variables = {
   'password' => Deferred('generate',
                   ['facter ec2_metadata | jgrep public-hostname']),
}

file { '/etc/myapp/network.conf':
  ensure  => file,
  content => Deferred('inline_epp',
               [find_template('myapp/network.conf').file, $variables]), # see https://github.com/puppetlabs/puppet/pull/7840
}
```

What's needed is a late resolved fact, that's only resolved at runtime and at every runtime.

I don't know of any other way to force a Facter run at catalog runtime. In standard Puppet runs, all the facts are gathered a single time prior to compilation. And as far as I know, they're not regenerated prior to each Puppet run.

Writing a wrapper function isn't about "saving a few keystrokes," it's about expressing _intent_. We do this _all the time_ and for good reason. The `upcase()` function just wraps the ruby function and is utterly unnecessary because you could always invoke arbitrary ruby via `inline_template()`. We don't do that because `inline_template()` doesn't express the intent we mean.

The `fact()` function itself is effectively already a wrapper function around `dig()` (really it's a reimplementation, but details). What it does is provide context so that people understand what it does and how to use it. (There are 12 users on the forge who've used `dig()` in a module.)

Calling the wrapper function `deferred_fact()` doesn't express intent. It expects the user to know the "obscure incantation" that is a deferred function. We don't document them at all. What we document is a brief page that says that you can now retrieve secrets on the agent. We don't say anywhere publicly that deferred functions are run each time the catalog is enforced.

Calling it `volatile_fact()` doesn't require the user to know all the weird details of how a deferred function works, it literally just says "this is a fact that changes often, make sure to use the current value." The average user doesn't care whether a function is deferred or not (if they even know). What they care about is that the catalog does what they need it to do.

lib/puppet/functions/fact.rb
binford2k: That's correct. Are there cases in which facts would not be in scope during a standard catalog compilation?

This is intended to use the in-scope facts during standard catalog compilation and fall back to running facter when deferred to the agent. Is there a better way?

hlindberg:
All fair points @binford2k - still don't think `volatile` is the right word here - fine if named after intent, but then `volatile` isn't it. Things like "most current fact",  "freshly obtained fact", "apply time fact", comes to mind but using some better word for that perhaps?

Think there should be some kind of guard rails against calling Facter when compiling (thus getting master facts) - would you ever want that?

binford2k:
> All fair points @binford2k - still don't think volatile is the right word here - fine if named after intent, but then volatile isn't it. Things like "most current fact", "freshly obtained fact", "apply time fact", comes to mind but using some better word for that perhaps?

Hmm;  `runtime_fact()` could maybe work, but  I wonder if a better solution is to improve the deferred docs to clarify more about what they are for and what they can or cannot do. If that were so, then `deferred_fact()` like @DavidS suggests is probably the best solution. I have the start of that filed as DOC-4384.

> Think there should be some kind of guard rails against calling Facter when compiling (thus getting master facts) - would you ever want that?

Absolutely, I just don't know the best way to do that. This *seemed* to only run Facter during application, but I don't know if it's actually correct or not.

```
facts = closure_scope['facts'] || Facter.to_hash 
```

hlindberg:
So, there are issues with getting the facts if not set in scope as `$facts`, a) risk of getting server side facts when called on master and b) if there are many "deferred facts" they would all result in running Facter and this would be slow c) since each call will run Facter, the values may differ for the same fact
 
The best would be if the facts were already present in scope when deferred calls are resolved. That requires a change to Puppet however as the setup and use of PAL on the agent side for resolution of Deferred must be given the facts to use. (The rest should already be supported by PAL; i.e. setting `$facts` etc.)

If not computed and set up front, then I think the result needs to be cached and kept while resolving deferred values. This would be good since if there is no call to get a deferred fact then Facter does not need to run at all. If going this route, there is a supported caching mechanism that functions can use. With this there is still the problem of getting the wrong facts (for server when agent was intended) as the place intention is known is when setting up the context (i.e. when setting up PAL for resolution, or when setting up for apply, master compile, bolt apply compile, etc.).

michaeltlombardi:
@binford2k are you planning on continuing work on this PR? It's been a few months, nor sure if you're still interested in picking this up.

sanfrancrisko:
Hi @binford2k - I'm just closing this PR as part of the IAC Team's audit of stale PRs. It can be reopened again if you wish to continue working to get it merged.

Thanks for the submission and ping us any time if you want to pick it up again.

igalic:
reading this pr about a year late, and wondering if it could be a potential solution to the all-to-common scenario where a module's internals depend on a version fact, but that can only be obtained after installing the package, so we need more than one run

it seems… upon closer reading… uh… probably not, at least not if we don't want more complexity than…simply running twice.

binford2k:
@michaeltlombardi @sanfrancrisko apologies for missing this earlier. I've gotten the docs updated and am now ok turning this into `deferred_fact()` like @DavidS suggests. I do not have the Puppet internals expertise to answer the points raised by @joshcooper or @hlindberg. If I had some guidance on how to implement the suggestions, then I would be happy to do so.

PR #1084, opened 2020-02-05
stdlib::end_with: create String.end_with function

b4ldr:
This PR wraps the ruby `String.end_with?` function and exposes it to the
puppet DSL.


b4ldr:
looks like the CI checks stalled and may need kicking

carabasdaniel:
Hi @b4ldr 
This looks good, thank you for your contribution.

PR #1086, opened 2020-02-10
Add start_with function

baurmatt:
This is helpful in places where you need to use variables in the string
which needs to be compare e.g. something like this:

```puppet
$test.start_with("part${string}01")
```

### DIFF
--- a/lib/puppet/functions/start_with.rb
+++ b/lib/puppet/functions/start_with.rb
@@ -1,0 +1,22 @@
+# @summary
+#   Returns true if str starts with one of the prefixes given. Each of the prefixes should be a String.
+#
+# @example
+#   'foobar'.start_with('foo') => true
+#   'foobar'.start_with('bar') => false
+#   'foObar'.start_with(['bar', 'baz']) => false
+Puppet::Functions.create_function(:start_with) do
+  # @param test_string The string to check
+  # @param prefixes The prefixes to check.
+  #
+  # @return [Boolean] True or False
+  dispatch :start_with do
+    param 'String[1]', :test_string
+    param 'Variant[String[1],Array[String[1], 1]]', :prefixes
+    return_type 'Boolean'
+  end
+
+  def start_with(test_string, prefixes)
+    test_string.start_with?(*prefixes)
+  end
+end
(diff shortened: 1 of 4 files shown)


b4ldr:
[wmflib::end_with](https://github.com/puppetlabs/puppetlabs-stdlib/blob/master/lib/puppet/functions/stdlib/end_with.rb) already exists.  and although i much prefer ends_with and starts_with i think it makes more sense to  make functions like this, a direct map to the ruby equivalent.  i.e. `startswith` should be renamed to `wmflib::start_with`

baurmatt:
Hey @b4ldr, oh missed that oO I will remove the endswith function from the PR.

Do you know when a function which be prefixed with `stdlib::`? My understanding is, that all modules should prefix their function besides stdlib.

b4ldr:
>Do you know when a function which be prefixed with stdlib::? My understanding is, that all modules should prefix their function besides stdlib.

no im not sure i thought all modules should be prefixed but ill let someone from puppetlabs give a more authoritative answer

b4ldr:
also the current `stdlib::end_with` only supports `String`, it would be nice if you could patch that to also support an `Array[String]`  like your patch

baurmatt:
Alright! :) I've updated the PR and tried to adapt the best of both variants :)

The last thing missing seems to be an authoritative answer regarding the function namespacing.

daianamezdrea:
Hi @baurmatt ,
Thank you for your contribution !

baurmatt:
Hey @daianamezdrea, I think there was a decision missing: That's the correct naming for function in stdlib? The implementation that you've just merged had both in it and will, at least, confuse people:

* stdlib::end_with
* start_with

daianamezdrea:
Hi @baurmatt , 
Based on the documentation from [here](https://puppet.com/docs/puppet/latest/functions_ruby_overview.html), the naming of the functions is correct, however I do agree with you on this naming confusion, we might need to move the "start_with" function in functions/stdlib/ folder.

Sorry I missed this in the review.

baurmatt:
Hey @daianamezdrea, which naming of the function is correct? Prefixed with `stdlib::` or without? The majority of the functions in this module don't prefix it, but some do.

I think I've read somewhere (I can't find the link anymore... -.-) that only `stdlib` is allowed to skip the module name prefix.

PR #1079, opened 2020-01-06
Puppet 4 functions

binford2k:
Recreating #1078 with a fork and rubocop fixes.

PR #1087, opened 2020-02-11
(wip) use json, not pson, steal Bens work to make pup4

justinstoller:
Needs tests and to double check the documentation / its generation.

Wanted to know if there were any concerns with the approach though.

FWIW, I first tried to overwrite the existing method but we have tests that invalid inputs cause the default given to be output instead of failing with a type error (which the newer functions would do if properly typed).

Alternatively, I could loosen the typing to allow more backwards compat but there'll still be some incompatibility between JSON <-> PSON. The content of the 4x functions was stolen from Ben's PR https://github.com/puppetlabs/puppetlabs-stdlib/pull/1079 and slightly cleaned up.

### DIFF
--- a/lib/puppet/functions/stdlib/loadjson.rb
+++ b/lib/puppet/functions/stdlib/loadjson.rb
@@ -1,0 +1,84 @@
+#
+#    @summary
+#      Load a JSON file containing an array, string, or hash, and return the data
+#      in the corresponding native data type.
+#
+#      The first parameter can be a file path or a URL.
+#      The second parameter is the default value. It will be returned if the file
+#      was not found or could not be parsed.
+#
+#    @return [Array|String|Hash]
+#      The data stored in the JSON file, the type depending on the type of data that was stored.
+#
+#    @example Example Usage:
+#      $myhash = loadjson('/etc/puppet/data/myhash.json')
+#      $myhash = loadjson('https://example.local/my_hash.json')
+#      $myhash = loadjson('https://username:password@example.local/my_hash.json')
+#      $myhash = loadjson('no-file.json', {'default' => 'value'})
+#
+#
+Puppet::Functions.create_function(:'stdlib::loadjson') do
+  require 'puppet/util/json'
+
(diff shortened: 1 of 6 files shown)


justinstoller:
fwiw, these are failing on lint checks not tests

> Wanted to know if there were any concerns with the approach though.

I've heard at least someone express a preference for making this replace the existing method rather than deprecating the old and replacing it with a "stdlib::" namespaced one. I'm unclear if we should then remove the stricter type checking? Do others have thoughts on the approach?

DavidS:
Hi @justinstoller,

Depending on how much the behaviour changes, it might be a barrier to switching over. I'm not familiar enough with the pson/json differences to be able to make a judgement call on that. I would guess that most users of those functions are actually looking for JSON, not pson semantics, so it might be really not a big deal in this case.

With the changes in #1079 we're moving all non-deprecated functions to `stdlib::` prefix. I haven't worked through all the issues there yet, but see the commit message on https://github.com/DavidS/puppetlabs-stdlib/commit/25a2af2e4fe437a769cdcf1f276489d7ebf36659 for a up-to-date snapshot of where the project is at.

lib/puppet/functions/stdlib/loadjson.rb
```diff
@@ -0,0 +1,84 @@
...
+  require 'puppet/util/json'
+
+  # If URL is in the format of https://username:password@example.local/my_info.json
+  URI_WITH_NAME_AND_PASS_PATTERN = %r{(http\://|https\://)(.*):(.*)@(.*)}
+
+  # If URL is in the format of https://username@example.local/my_info.json
+  URI_WITH_NAME_PATTERN = %r{(http\://|https\://)(.*)@(.*)}
+
+  dispatch :default_impl do
+    param 'String', :path_or_uri
```
binford2k: Could we use `Stdlib::HTTP(s)Url` here?

justinstoller: I tried something in d584063 I don't know if it's correct, still need to test it. But is it more what you were thinking?

PR #1095, opened 2020-03-04
Add correct namespace for start_with function

baurmatt:
This function was originally introduced in #1086 but never got a
decision about the correct namespace. This commit adds the stdlib::
namespace to the start_with function after a discussion I had with
Daneel on Slack about this. This will at least align the namespace with
the corresponding end_with function to not confuse people.

I'm not considering this a breaking change because the function was never released.


cmccrisken-puppet:
Thanks for the tweak @baurmatt !

PR #1097, opened 2020-03-09
[IAC-547] Remove strftime from stdlib

carabasdaniel:
Removing strftime as it was already superseded by https://github.com/puppetlabs/puppet/blob/master/lib/puppet/functions/strftime.rb


michaeltlombardi:
Marking this as a maintenance commit as the function is already overridden by the function within Puppet, meaning this is not a backwards breaking change.

DavidS:
Relabelled and retitled as bugfix so it shows up on the changelog.

PR #1106, opened 2020-05-11
Remove gemspec

ekohl:
This is not a current practice and it's very outdated. Remove it to avoid confusion.


sheenaajay:
Thank you @ekohl for submitting the PR.

PR #1108, opened 2020-05-13
Add a cron_provider fact

ekohl:
Similar to package_provider and service_provider the cron provider is useful to know. On some minimal platforms, crontab is not installed and the cron type should not be used. By having a fact for this, it becomes easy to check.

### DIFF
--- a/lib/facter/cron_provider.rb
+++ b/lib/facter/cron_provider.rb
@@ -1,0 +1,18 @@
+# Fact: cron_provider
+#
+# Purpose: Returns the default provider Puppet will choose to manage cron
+#   on this system
+#
+# Resolution: Instantiates a dummy cron resource and return the provider
+#
+# Caveats:
+#
+require 'puppet/type'
+require 'puppet/type/cron'
+
+Facter.add(:cron_provider) do
+  setcode do
+    provider = Puppet::Type.type(:cron).newcron(:name => 'dummy')[:provider]
+    provider.to_s if provider
+  end
+end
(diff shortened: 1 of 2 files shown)


sheenaajay:
Thanks @ekohl  for submitting the PR. Could you please fix the rubocop error. Thank you.

ekohl:
Updated

pmcmaw:
As there are inconsistent failures on travis and appveyor I am going to close and reopen your PR to reckick the jobs. :-)

ekohl:
This is an interesting one to see in the failure (for a different fact):
```
[2020-05-19 09:28:57.508594 ] ERROR Facter - Error loading fact /home/travis/build/puppetlabs/puppetlabs-stdlib/lib/facter/cron_provider.rb: cannot load such file -- puppet/type/cron 
```

I can't quite explain that, but should I catch that LoadError in my custom fact?

pmcmaw:
@ekohl Puppet 6 does not ship with cron as it is now self contained in its own module. For this reason I think that your PR would be better suited to the core cron module: https://github.com/puppetlabs/puppetlabs-cron_core

If you submit it there, please ping me on the PR so I can make sure it gets merged quickly.

ekohl:
Hmm, that complicates things for my use case. With stdlib I can let my module depend on some release but with the _core modules this isn't really the case. https://github.com/puppetlabs/puppetlabs-cron_core/pull/33 so maybe at some point in the future I can depend on it, but until then I'll create a custom fact in my own module.

PR #1113, opened 2020-06-18
Fix UTF-8 to_json_pretty.rb entry

Rocco83:
Ruby’s is written with UTF-8 characters, while all the other code is plain ASCII.


daianamezdrea:
Hi @Rocco83, this looks good! Thank you for your contribution!

PR #1114, opened 2020-07-08
[MODULES-3798] defined_with_params - defined types

trevor-vaughan:
This fixes an issue where trying to match against any declared defined
type with a specific parameter would never return that there was a match
in the catalog.

This is relevant when trying to ensure that you don't have port
conflicts in services that can declare multiple instances of a service
through a given defined type, such as Apache.

This, previously valid call to the function now works as expected:

``defined_with_params(My::DefinedType, { 'port' => 1337 })``

MODULES-3798 #close

### DIFF
--- a/lib/puppet/parser/functions/defined_with_params.rb
+++ b/lib/puppet/parser/functions/defined_with_params.rb
@@ -48,8 +48,13 @@ DOC
     title = nil
   end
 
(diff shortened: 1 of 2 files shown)


trevor-vaughan:
Sorry, wrong ticket number. Will resubmit

PR #1122, opened 2020-08-19
[MODULES-10781] Fix defined type defined_with_params()

trevor-vaughan:
An edge case was missed previously where calling the
defined_with_params() function from inside a defined type on another
instance of the same defined type would always return a match because it
would match against itself.

This updates the code to ignore itself and also adds a debug line output
for any resource that was matched.

MODULES-10781 #close


david22swan:
@trevor-vaughan Thanks for your contributions, this looks good to me.

PR #1125, opened 2020-09-10
(PUP-10653) Remove win32/dir constant usage

GabrielNagy:
For Puppet 7 we are dropping the `win32/dir` dependency as we only used constants from it, which we replaced with environment variables (see: https://github.com/puppetlabs/puppet/pull/8314).

This would become breaking when using the stdlib module with Puppet 7, as the win32 dependencies are no longer provided in the puppet gem.

Replace the usage of the `Dir::COMMON_APPDATA` with the `ALLUSERSPROFILE` environment variable.

### DIFF
--- a/lib/facter/facter_dot_d.rb
+++ b/lib/facter/facter_dot_d.rb
@@ -207,10 +207,8 @@ if mdata
 
       # Windows has a different configuration directory that defaults to a vendor
       # specific sub directory of the %COMMON_APPDATA% directory.
-      if Dir.const_defined? 'COMMON_APPDATA' # rubocop:disable Metrics/BlockNesting : Any attempt to alter this breaks it
-        windows_facts_dot_d = File.join(Dir::COMMON_APPDATA, 'PuppetLabs', 'facter', 'facts.d')
-        Facter::Util::DotD.new(windows_facts_dot_d).create
-      end
+      windows_facts_dot_d = File.join(ENV['ALLUSERSPROFILE'], 'PuppetLabs', 'facter', 'facts.d')
+      Facter::Util::DotD.new(windows_facts_dot_d).create
     end
   end
 end


joshcooper:
Since we dropped support for 2003 years ago, then AKAIK `Dir::COMMON_APPDATA` should always equal to `ENV['ALLUSERSPROFILE']? If so, then we should merge this to the earliest supported stdlib version, preferably whatever version is commonly used with puppet5. In other words, it's not actually a breaking change.

If we instead merge to `main` and release a new major version of stdlib, then existing stdlib versions will fail with puppet7. Users will be forced to update to the latest stdlib in order to deploy puppet7, but that can be difficult when stdlib is a transitive dependency for so many puppet modules, as each modules' metadata needs to be updated to accept the next major. I'd like to avoid that if possible given the change produces the same result, just without the win32-dir gem.

DavidS:
From a naive reading of the code (I'm still trying to catch up with everything that happened over the last two months) this does not look like a breaking change, especially as you seem to be saying that after Windows Server 2003 this is a noop change anyways? If my understanding is correct, why would it then be a problem merging this a bug fix and asking people to use this as the minimum version to use on puppet 7?

@joshcooper core puppet tools (`puppet module install`, `r10k`, the forge, etc) are unable to negotiate installing a non-latest module version matching local puppet version requirements, AND we provide a lot of value through large compatibility ranges for folks still running legacy puppet versions along-side newer installs. Therefore `main`/the latest released version are "the earliest supported stdlib version" and likely the "version is commonly used with puppet5".

adrianiurca (approved):

joshcooper:
>  If my understanding is correct, why would it then be a problem merging this a bug fix

I was mistakenly under the impression that `main` was for the next major release while 4.x, etc branches were for older stdlib releases. But it looks like those branches don't exist anymore and there's just `main`, so 👍  for all the reasons you describe. Thanks!

PR #1126, opened 2020-09-14
Stdlib::Datasize: This CR adds a new data size type alias

b4ldr:
Many class/define parameters take a value representing a data size.
e.g. 1024Mb, 1024G, 100k

This new type provides a way to validate most common data type inputs


daianamezdrea:
Hi @b4ldr, the PR looks good! Thank you for your contribution!

PR #1127, opened 2020-09-14
Resolve puppet-lint warnings

jcpunk:
I like to run puppet-lint against my whole code tree every so often.

These changes should make stdlib fully puppet-lint clean.


sanfrancrisko (approved):
Thanks for the clean up and resolution of the `puppet-lint` issues @jcpunk !

PR #1130, opened 2020-09-26
Add parsehocon() function

reidmv:
This function aligns with the existing parsejson and parseyaml functions, providing similar functionality for HOCON strings. This is useful for dealing with Puppet configuration files, many of which are HOCON format.


sanfrancrisko (approved):
Thanks for the enhancement @reidmv !

PR #1133, opened 2020-10-06
Fix latest release version in CHANGELOG.md

smortex:
There is a dupe 6.4.0 entry.  The first one is in fact the 6.5.0 release.


pmcmaw:
Thanks for this PR @smortex 👍

PR #1135, opened 2020-10-22
Allow start/end checks on empty strings

jvrsantacruz:
When using this library function on strings that come from a parameter,
they might be empty sometimes, rendering it difficult to use, having to
force hacks like.

    "_${source}".stdlib::start_with('_puppet:///modules/')

Where `_` is added to make sure that the string is never empty.

In Ruby the `start_with?` function allows being used on empty strings,
and it returns `false` for any assertion to be done:

    p ''.start_with?('test')
    false

Also in Python the `startswith()` method returns `false` for empty
strings:

    In [1]: print(''.startswith('test'))
    False

As this logic seems the most extended and allows for a wider use of the
library, removing these corner cases when the string is empty seems like a good idea.


DavidS (approved):

DavidS:
Thanks for your contribution!

PR #1141, opened 2020-11-23
(FEAT) Add simple standardized header template

jcpunk:
It seems every module has a different way of putting the `This is a puppet managed file don't edit it` in their templates.  They are all a bit different and each one says slightly different things.  This is my attempt to get something a bit more unified in place so folks can have greater consistency in their modules.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,7 @@ Puppet modules make heavy use of this standard library. The stdlib module adds t
  * Defined types
  * Data types
  * Providers
+ * Templates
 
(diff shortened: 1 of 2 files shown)


michaeltlombardi (commented):
Thanks for this feature submission! On a quick review, this looks like it might be better to land in [`puppet-extlib`](https://github.com/voxpupuli/puppet-extlib) rather than `stdlib` - the UX here is a _little_ complicated, and while the idea is getting at resolving a lack of standardization in modules, it seems like a better fit for the extended library instead of this one. Thank you though!

PR #1154, opened 2021-01-12
Drop facter.d facts needed for Facter < 1.7

ekohl:
This fact is only needed for compatibility with ancient versions of Facter. Facter 1.7 has facts.d support built in an was released in April 2013. While this file is mostly a noop on recent Facter versions, it's better not to load it at all.


pmcmaw:
Thank you for your PR @ekohl 
We appreciate your contribution as always :-) 
Merging.

PR #1160, opened 2021-02-04
Stdlib::Email type

b4ldr:
Add new type to validate emails.  The regex has been taken from the
[html5 spec][1]

[1]: https://html.spec.whatwg.org/multipage/input.html#valid-e-mail-address

### DIFF
--- a/spec/type_aliases/email_spec.rb
+++ b/spec/type_aliases/email_spec.rb
@@ -1,0 +1,67 @@
+require 'spec_helper'
+
+# Test cases are a combination of the test cases used in MediaWiki[1] and a
+# Reference found on line[2].  Some of the test cases in the later list have
+# been dropped as the regex used in the HTML5 specification[3] (and in this type)
+# allows for wilful violation of the RFC's
+#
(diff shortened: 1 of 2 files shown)


pmcmaw:
Thanks for your PR @b4ldr :-)

pegasd:
I just noticed this PR (apparently a little late) and see that newline handling is not done properly here.
As I mention here https://github.com/puppetlabs/puppetlabs-stdlib/pull/1066, it's better to use `\A`, `\z` instead of `^`, `$` in all type aliases because of the way multiline strings are handled in Puppet.

Currently, things like
```
random stuff
valid@email.com
more random stuff $^*!
```

will match, which is, obviously, not correct.

pmcmaw:
@pegasd thanks for bringing this to my attention. 

Please note nothing is ever too late :-) Our code bases are always evolving and changes/PRs can be merged at anytime. 
I have opened the following PR to resolve the issue you have highlighted and I would be super grateful if you had time for a review when the tests run. 

PR to resolve: #1163

PR #1173, opened 2021-03-04
pw_hash: add support for bcrypt variants

kjetilho:

### DIFF
--- a/spec/functions/pw_hash_spec.rb
+++ b/spec/functions/pw_hash_spec.rb
@@ -52,6 +52,13 @@ describe 'pw_hash' do
 
   context 'when the third argument contains invalid characters' do
     it { is_expected.to run.with_params('password', 'sha-512', 'one%').and_raise_error(ArgumentError, %r{characters in salt must be in the set}) }
+    it { is_expected.to run.with_params('password', 'bcrypt', '1234').and_raise_error(ArgumentError, %r{characters in salt must match}) }
+  end
+
+  context 'when run' do
+    it { is_expected.to run.with_params('password', 'sha-512', '1234').and_return(%r{^\$6\$1234\$}) }
+    it { is_expected.to run.with_params('password', 'bcrypt', '05$abcdefghijklmnopqrstuv').and_return(%r{^\$2b\$05\$abcdefghijklmnopqrstu}) }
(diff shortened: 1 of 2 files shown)


kjetilho:
(This is a good candidate for squashing commits when merging :-)

spec/functions/pw_hash_spec.rb
```diff
@@ -52,6 +52,7 @@
 
   context 'when the third argument contains invalid characters' do
     it { is_expected.to run.with_params('password', 'sha-512', 'one%').and_raise_error(ArgumentError, %r{characters in salt must be in the set}) }
+    it { is_expected.to run.with_params('password', 'bcrypt', '1234').and_raise_error(ArgumentError, %r{characters in salt must match}) }
```
david22swan, line 55: While this look's good overall I would prefer it if you were to expand the added test's to include the other new algorithms.

david22swan:
@kjetilho Just one small comment, everything else look's good and I'll happily merge once it's resolved.

spec/functions/pw_hash_spec.rb
kjetilho: Sure, I could do that, but I'm not convinced it is worthwhile since the test is just for bad salt values and the other three tests would be ~identical.  I would love to test the hash methods more properly, but as mentioned in commit f3800aa, the test system's Ruby doesn't support them :/

The salient error from the spec test was:

```
expected pw_hash("password", "bcrypt", "05$abcdefghijklmnopqrstuv") to have returned /^\$2b\$05\$abcdefghijklmnopqrstu/ instead of raising Errno::EINVAL(Invalid argument - crypt)
```

kjetilho:
So with the last changes, Travis here reports
> Only testing bcrypt results on glibc 2.28 and later (PENDING: Not yet implemented)

The tests run nicely in my CentOS 8 and Ubuntu Focal test environments, however.

kjetilho:
I can squash commits into one feature patch commit and one commit for the documentation update if desired.

daianamezdrea:
Hi @kjetilho, sounds perfect! Thank you for your contribution !

kjetilho:
Rebased and squashed down, ready for merge.

david22swan:
@kjetilho A few transient failures but rekicking sorted them out.
Everything looks good to me so I've gone ahead and merged, thank you for putting in the work and I look forward to seeing more PRs from you in the future. :)

PR #1184, opened 2021-04-26
(MODULES-11053) Add a note about the contains method

DavidS:


sanfrancrisko (approved):

sanfrancrisko (changes requested):
Rubocop doesn't seem happy with that change: https://github.com/puppetlabs/puppetlabs-stdlib/pull/1184/checks?check_run_id=2436852428#step:9:27

sanfrancrisko (approved):

PR #1189, opened 2021-05-31
(MODULES-11018) Rewrite pick_default in native Puppet

ekohl:
This rewrites pick_default in native Puppet which makes it easier to maintain.

In doing so, it should also fix a bug where calling pick_default(undef, undef) should work according to docs and return undef but it returns a string.


ekohl:
Perhaps pure Puppet functions can't be not-namespaced?

PR #1192, opened 2021-07-06
Stdlib::Http::Method: Add new type for http methods

b4ldr:
This PR creates new new resources:
 * Stdlib::Http::Method for validating http methods
 * Stdlib::Http::Status This is just a copy of Stdlib::Httpstatus
 * make Stdlib::Httpstatus and alias to Stdlib::Http::Status

Ideally we would deprecate Stdlib::Httpstatus in favour of
Stdlib::Http::Status


spec/type_aliases/http__method_spec.rb
```diff
@@ -0,0 +1,42 @@
+require 'spec_helper'
+
+if Puppet::Util::Package.versioncmp(Puppet.version, '4.5.0') >= 0
```
ekohl, line 3: This shouldn't be supported according to metadata.json:
https://github.com/puppetlabs/puppetlabs-stdlib/blob/b4048fd3f5cfe365a03e8dc3e11eb2d42cefe803/metadata.json#L101

(same for the other type)

ekohl (commented):
Makes sense to me but I'd drop the Puppet >= 4.5 part in the tests since it's not supported anyway.

types/httpstatus.pp
```diff
@@ -1 +1 @@
-type Stdlib::HttpStatus = Integer[100, 599]
+type Stdlib::HttpStatus = Stdlib::Http::Status
```
ekohl, line 1: I wonder if this type can be deprecated.

ekohl (commented):
I also just realized this may generate additional things in `REFERENCE.md`. Would you mind regenerating that and making sure the output makes sense?

spec/type_aliases/http__method_spec.rb
b4ldr: done, i also created #1193 for the rest of them

types/httpstatus.pp
b4ldr: i would also like to deprecate this in favour of `Stdlib::Http::Status` but not sure of the process for that.  i regret not namespacing this when i first introduced it :(

b4ldr:
> I also just realized this may generate additional things in `REFERENCE.md`. Would you mind regenerating that and making sure the output makes sense?

We end up with two sections for http status 

```
 
+### <a name="stdlibhttpmethod"></a>`Stdlib::Http::Method`
+
+https://www.iana.org/assignments/http-methods/http-methods.xhtml
+
+Alias of
+
+```puppet
+Enum['ACL', 'BASELINE-CONTROL', 'BIND', 'CHECKIN', 'CHECKOUT', 'CONNECT', 'COPY', 'DELETE', 'GET', 'HEAD', 'LABEL', 'LINK', 'LOCK', 'MERGE', 'MKACTIVITY', 'MKCALENDAR', 'MKCOL', 'MKREDIRECTREF', 'MKWORKSPACE', 'MOVE', 'OPTIONS', 'ORDERPATCH', 'PATCH', 'POST', 'PRI', 'PROPFIND', 'PROPPATCH', 'PUT', 'REBIND', 'REPORT', 'SEARCH', 'TRACE', 'UNBIND', 'UNCHECKOUT', 'UNLINK', 'UNLOCK', 'UPDATE', 'UPDATEREDIRECTREF', 'VERSION-CONTROL']
+```
+
+### <a name="stdlibhttpstatus"></a>`Stdlib::Http::Status`
+
+The Stdlib::Http::Status data type.
+
+Alias of
+
+```puppet
+Integer[100, 599]
+```
+
 ### <a name="stdlibhttpstatus"></a>`Stdlib::HttpStatus`
 
 The Stdlib::HttpStatus data type.
@@ -6777,7 +6865,7 @@ The Stdlib::HttpStatus data type.
 Alias of
 
 ```puppet
-Integer[100, 599]
+Stdlib::Http::Status
 ```
```

types/httpstatus.pp
b4ldr: can always do this bit in a separate pr if that makes things easier?

michaeltlombardi:
Test Failure unrelated; this was based on the #1193, which passed.

b4ldr:
Im not sure what happened to this but i don't see it in the history https://github.com/puppetlabs/puppetlabs-stdlib/commits/main/types :/

PR #1193, opened 2021-07-06
(MAINT) Remove testing logic for Puppet 4.5

b4ldr:
Support for puppet 4.5 was dropped from stdlib some time ago however the 
spec tests all include an if guard to prevent them running on puppet
4.5.  As this is no longer required this PR cleans up theses guards


ekohl (changes requested):
Looks like it includes https://github.com/puppetlabs/puppetlabs-stdlib/pull/1192 now. I'd prefer to keep that separate. Also looks like Rubocop is unhappy.

b4ldr:
> Looks like it includes #1192 now. I'd prefer to keep that separate. Also looks like Rubocop is unhappy.

good catch that was over-site should be removed now

spec/type_aliases/ip_address_v6_nosubnet_full_spec.rb
```diff
@@ -1,28 +1,24 @@
 require 'spec_helper'
 
-if Puppet::Util::Package.versioncmp(Puppet.version, '4.5.0') >= 0
-  describe 'Stdlib::IP::Address::V6::Nosubnet::Full' do
```
ekohl, line 4: I think this line should not have been removed

types/httpstatus.pp
```diff
@@ -1 +1 @@
-type Stdlib::HttpStatus = Integer[100, 599]
+type Stdlib::HttpStatus = Stdlib::Http::Status
```
ekohl, line 1: This is also unrelated

ekohl (changes requested):

spec/type_aliases/ip_address_v6_nosubnet_full_spec.rb
b4ldr: added back

ekohl (approved):
I'll let others take a look as well.

PR #1196, opened 2021-07-27
Fix ensure_packages

cocker-cc:
Flip `installed` and `present` in Function `ensure_packages`.
Puppet-Type `package` defaults to `installed` and not to `present` (compare [here](https://github.com/puppetlabs/puppet/blob/main/lib/puppet/type/package.rb#L148) ).
This Misconception lies in the Code since the most early Days.

With flipping these two Values, Rspec-Tests do not have to be adapted, when migrating from `package {}` to `ensure_packages()`.

### DIFF
--- a/lib/puppet/parser/functions/ensure_packages.rb
+++ b/lib/puppet/parser/functions/ensure_packages.rb
@@ -20,12 +20,12 @@ module Puppet::Parser::Functions
 
     if arguments[0].is_a?(Hash)
       if arguments[1]
-        defaults = { 'ensure' => 'present' }.merge(arguments[1])
-        if defaults['ensure'] == 'installed'
-          defaults['ensure'] = 'present'
+        defaults = { 'ensure' => 'installed' }.merge(arguments[1])
+        if defaults['ensure'] == 'present'
+          defaults['ensure'] = 'installed'
         end
       else
-        defaults = { 'ensure' => 'present' }
+        defaults = { 'ensure' => 'installed' }
       end
 
       Puppet::Parser::Functions.function(:ensure_resources)
@@ -34,12 +34,12 @@ module Puppet::Parser::Functions
       packages = Array(arguments[0])
 
       if arguments[1]
-        defaults = { 'ensure' => 'present' }.merge(arguments[1])
-        if defaults['ensure'] == 'installed'
-          defaults['ensure'] = 'present'
+        defaults = { 'ensure' => 'installed' }.merge(arguments[1])
+        if defaults['ensure'] == 'present'
+          defaults['ensure'] = 'installed'
         end
       else
-        defaults = { 'ensure' => 'present' }
+        defaults = { 'ensure' => 'installed' }
       end
 
       Puppet::Parser::Functions.function(:ensure_resource)
--- a/spec/functions/ensure_packages_spec.rb
+++ b/spec/functions/ensure_packages_spec.rb
@@ -24,7 +24,7 @@ describe 'ensure_packages' do
 
       # this lambda is required due to strangeness within rspec-puppet's expectation handling
       it { expect(-> { catalogue }).to contain_package('puppet').with_ensure('absent') }
-      it { expect(-> { catalogue }).to contain_package('facter').with_ensure('present') }
+      it { expect(-> { catalogue }).to contain_package('facter').with_ensure('installed') }
     end
 
     describe 'after running ensure_package("facter", { "provider" => "gem" })' do
@@ -32,7 +32,7 @@ describe 'ensure_packages' do
 
       # this lambda is required due to strangeness within rspec-puppet's expectation handling
       it { expect(-> { catalogue }).to contain_package('puppet').with_ensure('absent').without_provider }
-      it { expect(-> { catalogue }).to contain_package('facter').with_ensure('present').with_provider('gem') }
+      it { expect(-> { catalogue }).to contain_package('facter').with_ensure('installed').with_provider('gem') }
     end
   end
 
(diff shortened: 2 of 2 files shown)


adrianiurca (approved):

adrianiurca:
LGTM

bastelfreak:
Hi Puppet,
can you please rethink if this is a backwards incompatible change? This change breaks all unit tests in modules that verify ensure state: https://github.com/voxpupuli/puppet-openvpn/pull/405

cocker-cc:
> Hi Puppet,
> can you please rethink if this is a backwards incompatible change? This change breaks all unit tests in modules that verify ensure state: [voxpupuli/puppet-openvpn#405](https://github.com/voxpupuli/puppet-openvpn/pull/405)

This may be true, but as `ensure_packages()` did it wrong all the Time, the now breaking Tests relied on this wrong behaviour.  At this Point the Cat bites its own Tail.  I prefer to correct `ensure_packages()` and let the Tests break.  All there is to apply in the breaking Tests, is to convert from `present` to `installed`.  

Exactly this is my Point:  When I have `package { 'blubb': }` (with default `ensure => installed`) and Test for `contain_package('blubb').with_ensure('installed')` and then I decide to migrate to `ensure_packages('blubb')` then my Test breaks, because of `ensure_packages`'s wrong Default `ensure => present`.

bastelfreak:
I don't mind having this merged, I just think it's backwards incompatible and the next stdlib release should be 8.0.0 and not 7.x.

ekohl:
> I don't mind having this merged, I just think it's backwards incompatible and the next stdlib release should be 8.0.0 and not 7.x.

I'm not entirely sure whether I agree. While functionally it's a change, I think it doesn't really warrant a major version bump because for end users it doesn't change anything.

You can argue that the testing method that was linked was checking internal behavior of `ensure_packages()` because there was no other way to test it.

Another thing that we should consider is that many (most?) Puppet modules use git checkouts so the version number is really irrelevant. Weighing this against the very painful costs for stdlib (so many modules need updating), I don't think we should.

That said, I have changed the PR title because it was really not covering the intent of this PR. That makes for an unusable changelog entry.

alexjfisher:
> > I don't mind having this merged, I just think it's backwards incompatible and the next stdlib release should be 8.0.0 and not 7.x.
> 
> I'm not entirely sure whether I agree. While functionally it's a change, I think it doesn't really warrant a major version bump because for end users it doesn't change anything.

@ekohl It looks like it breaks all of my actual catalogs.  I've just run a catalog-diff against a branch with stdlib 8.2.0 vs 5.2.0 and I've got ~1600 compilation failures.  Duplicate declaration of package resources. :(

alexjfisher:
The following compiled before, and now it doesn't.
```puppet
package { 'tzdata':
  ensure => present,
}

ensure_packages(['tzdata'])
```

```
Error: Evaluation Error: Error while evaluating a Function Call, Duplicate declaration: Package[tzdata] is already declared at (file: /home/afisher/test.pp, line: 1); cannot redeclare (file: /home/afisher/test.pp, line: 5) (file: /home/afisher/test.pp, line: 5, column: 1) on node host.example.com

alexjfisher:
Whilst the default for `ensure` in https://github.com/puppetlabs/puppet/blob/b3ccdff405c7dc1328649badf15ceb24a65bdc11/lib/puppet/type/package.rb#L148 might be `installed`, I think it's *far* more common to see people specify `ensure => present` than leave it to the default.  (Maybe someone could run some stats on this for some common puppet forge modules?? @binford2k ?? ;) )

IMO, this PR should **not** have been merged.  Ironically, whilst the purpose of this function is a (hacky) way to avoid duplicate resource declaration, this PR has created them for people who were using it.

Maybe it's still less harmful to revert it even though it's been in a release since last August??

cocker-cc:
> The following compiled before, and now it doesn't.
> […]

There are these Options for you, @alexjfisher:
* in the first Statement also use `ensure_packages('tzdata')` (most prefered)
* in the first Statement just use `package { 'tzdata': }` to use `package`'s Default-Value `ensure => installed`
* in the first Statement use `package { 'tzdata': ensure => installed }` to duplicate `package`'s Default-Value `ensure => installed`
* in the second Statement use `ensure_packages('tzdata', { ensure => present })` (least prefered)

Yes, `ensure_packages()` is a Hack. But `package` defaulted to `installed` since ever.  And the Hack `ensure_packages()` did not respect this.  According to the "Principle of least Surprise" both Default-Values should not differ.  Exactly this was the Goal of this PR.

alexjfisher:
That was a simple example to show the problem.  Would you like to create PRs for all of the following??
https://github.com/puppetlabs/puppetlabs-ntp/blob/aae5001a20352bf815c8f86b5606a6ed06168def/data/common.yaml#L36
https://github.com/puppetlabs/puppetlabs-docker/blob/5ab357f86d9f381f2dadfb318a8cf2dea82643bd/manifests/params.pp#L5
https://github.com/puppetlabs/puppetlabs-java/blob/5ebefbf1a9b2c52310be65106ef60c9fd7d33e13/manifests/init.pp#L40
https://github.com/puppetlabs/puppetlabs-java/blob/5ebefbf1a9b2c52310be65106ef60c9fd7d33e13/manifests/init.pp#L40
https://github.com/puppetlabs/puppetlabs-postgresql/blob/7992559aaa219d0f736a6b661d48d2e721c322a1/manifests/params.pp#L26
https://github.com/puppetlabs/puppetlabs-haproxy/blob/24049ec73f935119370289263774d5d4e995eff3/manifests/init.pp#L123
https://github.com/puppetlabs/puppetlabs-mysql/blob/a67f0ebcfd8e8c666159df7c4bb75fbdc20cf117/manifests/params.pp#L13

etc.

alexjfisher:
That was the whole point of the function.  If you'd included some module outside of your direct control, you would generally get away with using `ensure_packages` later.  Now you don't.

ekohl:
I'm wondering if it somehow could become flexible enough to accept either alias. Looks like it's using `defined_with_params`. Perhaps it needs some smarter logic to say "it can be either x or y".

alexjfisher:
Rightly or wrongly, this did go in, so I'll try and be more constructive and work on a fix that adds more intelligence and works for everyone.

> I'm wondering if it somehow could become flexible enough to accept either alias. Looks like it's using defined_with_params. Perhaps it needs some smarter logic to say "it can be either x or y".

`ensure_resource` does which is eventually called by `ensure_packages`. So, I suspect the enhancement will need to go directly in `ensure_packages`.  I'm not sure there's a way that I can generically find out what parameter values are aliases for others.

PR #1209, opened 2021-09-09
Added to_toml function

nmaludy:
This adds a function to stdlib for converting Puppet hashes to TOML. This is useful for Go services like Telegraf and Gitlab CI Runner than use TOML as their config format. Having this as a common function alleviates the problem of trying to load the `toml-rb` gem before running catalogs (chicken before the egg problem).


carabasdaniel:
Hi @nmaludy, thank you for your contribution as always. This looks great 👍

carabasdaniel (approved):

PR #1210, opened 2021-09-16
pdksync - (IAC-1598) - Remove Support for Debian 8

david22swan:
(IAC-1598) - Remove Support for Debian 8
pdk version: `2.2.0`

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -51,7 +51,6 @@
     {
       "operatingsystem": "Debian",
       "operatingsystemrelease": [
-        "8",
         "9",
         "10",
         "11"


ekohl:
The current bugfix label is very usual for dropping an OS.

It's also been extremely annoying to have so many major version bumps of stdlib for basically nothing. Is this really needed? The last major version bump hasn't even propagated through the community (I haven't even started on my modules). Please reconsider the policy on this.

I feel like stdlib should stop listing individual OS releases. It's pretty much only Ruby code which is OS-independent.

adrianiurca (approved):

PR #1211, opened 2021-09-22
[MODULES-11082] Fix Ruby lang typo in root_home.rb

riton:
Fixes `/var/lib/puppet/lib/facter/root_home.rb:43: syntax error unexpected '.'`

* Adapt changes introduced by 03d4b82630b911b6e574fae58f47520669096133
* Fixes [MODULES-11082](https://tickets.puppetlabs.com/browse/MODULES-11082)

### DIFF
--- a/lib/facter/root_home.rb
+++ b/lib/facter/root_home.rb
@@ -40,7 +40,7 @@ Facter.add(:root_home) do
   root_home = nil
   setcode do
     str = Facter::Util::Resolution.exec('lsuser -c -a home root')
-    str&.split("\n")&.each do |line|
+    str.split("\n").each do |line|
       next if %r{^#}.match?(line)
       root_home = line.split(%r{:})[1]
     end


ekohl (commented):
This is not a typo but rather a Ruby 2.5 language feature. By removing them you also remove handling of `nil` result from `Facter::Util::Resolution.exec`. I don't know if that can actually be returned, but that's important to realize.

Out of interest, why do you have a Ruby 2.4 (or older) installation? Is it Puppet 5?

riton:
Damn, pardon my ignorance, I didn't know about that Ruby 2.5 feature.

The infrastructure I was testing on is still running a Puppet 4 agent against a Puppet 7 server. This bug raises during our migration process.
I've just tested again with a v7 agent and this works without any problem.

I'm closing this P.R since there is nothing to fix. Sorry for the noise

PR #1230, opened 2022-04-12
add Stdlib::Dnssrv type

jhoblitt:


jhoblitt:
I am having trouble figuring out how to make this work within the constraints of a type alias (and ruby's regex engine).  I would like to require that a `_foo.` segment has to appear somewhere within the string but not require it to be the first segment as real world SRV records end up with constructions like `dev._locations.example.com`.

types/dnssrv.pp
```diff
@@ -0,0 +1,2 @@
+# match DNS RR (DNS SRV) similar to what is described by RFC2782 but is a superset to match real world usage by Freeipa. See https://www.rfc-editor.org/rfc/rfc2782
+type Stdlib::Dnssrv = Pattern[/\A(([_a-zA-Z0-9][a-zA-Z0-9\-]*[a-zA-Z0-9])\.)*(([_a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]*[a-zA-Z0-9])\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\-]*[A-Za-z0-9])\z/]
```
b4ldr, line 2: thinking about extensability i think it would be better to name space this as Stdlib::Dns::Srv

spec/type_aliases/dnssrv_spec.rb
```diff
@@ -0,0 +1,43 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe 'Stdlib::Dnssrv' do
+  describe 'valid handling' do
+    [
+      '_kerberos.example.com',
```
b4ldr, line 8: AFAIK proto is required

chelnak:
Hey @jhoblitt, are you able to move forward with this PR? It looks like there are a couple of comments to address.

Thank you 😄

jhoblitt:
@chelnak I'm not sure what to do here. What @b4ldr is saying is technically correct w.r.t. to the RFC but I need this for SRV records as are required by Red Hat Identity Management / FreeIPA and are supported by route53.  The spec examples are taking from one of my live production route53 zones.

b4ldr:
> @chelnak I'm not sure what to do here. 

Perhaps we could have something like the following
```puppet
Stdlib::Dns::Srv = ... # RFC compliment form
Stdlib::Dns::Srv::Route53  = ...  # vendor specific form
Stdlib::Dns::Srv::FreeIPA  = ... 
# suggest using only one of theses
Stdlib::Dns::Srv::Any = Varient[Stdlib::Dns::Srv::Route53, Stdlib::Dns::Srv::FreeIPA, Stdlib::Dns::Srv] 
```


> i need this for SRV records as are required by Red Hat Identity Management / FreeIPA and are supported by route53. The spec examples are taking from one of my live production route53 zones.

regardless of the outcome of the above comment, can you provide some documentation on the specification used by route53/FreeIPA so we can ensure this implementation matches the intention?  thanks

b4ldr:
Another option could be to have

```puppet
Stdlib::DNS::Srv::Strict = # rfc
Stdlib::DNS::Srv::Loose = # some looser definition which is common among multiple provider implementations 
Stdlib::Dns::Srv = # an argument on what should be the default to obviously vote strict ;)
```

jhoblitt:
Ha. Having a `::loose` type was just what I was thinking too.  Let me see if I can find documentation on why freeipa is doing what it is doing.  I've been working off the output of a command to dump the list of dns records which are required.

pmcmaw:
Hey @jhoblitt just wanting to check if you have any bandwidth or plans to pick this up again? :-)

jordanbreen28:
Hi @jhoblitt - We are hoping to progress some of our ageing PRs - Just checking if you still hope to proceed with this? If so, there seems to be some comments that need addressed and the PR will need rebasing with the current main. Thanks :)

jhoblitt:
Hi, yes, thanks for the nudge.

jordanbreen28:
Hi @jhoblitt! We are going through some of our ageing PRs and this one seems to have limited activity. We understand that you may have other commitments keeping you busy, but for now we will close this PR. 

Once you're ready to pick it up again, please do re-open this PR and we will pick it up again from here. Thanks for your understanding!

PR #1232, opened 2022-04-13
(maint) Update str2saltedpbkdf2.rb to use the correct salt length

AriaXLi:
The documentation in str2saltedpbkdf2.rb did not have the right length salt. 

OS X versions 10.15 and higher require the salt to be 32-bytes. Since Puppet's user  resource requires the value to be hex encoded, the length of the salt's string must be 64. str2saltedpbkdf2.rb was updated to reflect this.


chelnak (approved):
This seems like a sensible change! Thank you.

PR #1233, opened 2022-04-14
MODULES-11309 : convert a string to a resource

jcpunk:
I'll need a bit of help loading the resources into the catalog for the tests, but in theory that should be all that is required.

### DIFF
--- a/lib/puppet/functions/stdlib/str2resource.rb
+++ b/lib/puppet/functions/stdlib/str2resource.rb
@@ -19,7 +19,7 @@ Puppet::Functions.create_function(:'stdlib::str2resource') do
   # @return Puppet::Resource
   dispatch :str2resource do
     param 'String', :res_string
-    #return_type 'Puppet::Resource'
+    # return_type 'Puppet::Resource'
     return_type 'Any'
   end
 
@@ -29,9 +29,9 @@ Puppet::Functions.create_function(:'stdlib::str2resource') do
     resource = closure_scope.findresource(type_name, title)
 
     if resource.nil?
-      raise(Puppet::ParseError, "stdlib::str2resource(): could not find #{type_name}[#{title}], this is parse order dependant and values should not be quoted")
+      raise(Puppet::ParseError, "stdlib::str2resource(): could not find #{type_name}[#{title}], this is parse order dependent and values should not be quoted")
     end
 
-    return resource
+    resource
   end
 end
--- a/lib/puppet/functions/stdlib/str2resource.rb
+++ b/lib/puppet/functions/stdlib/str2resource.rb
@@ -1,0 +1,37 @@
+# frozen_string_literal: true
+
+# @summary
+#   This converts a string to a puppet resource.
+#
+#   This attempts to convert a string like 'File[/foo]' into the
+#   puppet resource `File['/foo']` as detected by the catalog.
+#
+#   Things like 'File[/foo, /bar]' are not supported as a
+#   title might contain things like ',' or ' '.  There is
+#   no clear value seperator to use.
+#
+#   This function can depend on the parse order of your
+#   manifests/modules as it inspects the catalog thus far.
+Puppet::Functions.create_function(:'stdlib::str2resource') do
+  # @param res_string The string to lookup as a resource
+  # @example
+  #   stdlib::str2resource('File[/foo]') => File[/foo]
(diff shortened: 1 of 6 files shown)


lib/puppet/parser/functions/str2resource.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+#
+# str2resource.rb
+#
+module Puppet::Parser::Functions
+  newfunction(:str2resource, type: :rvalue, doc: <<-DOC
```
ekohl, line 7: Isn't this the legacy API? Any thoughts on using the new Ruby v4 API?

jcpunk: I'd be happy to do so, is there some code I can clone from?

alexjfisher: This module has got [loads of examples](https://github.com/puppetlabs/puppetlabs-stdlib/tree/main/lib/puppet/functions/stdlib).  Only a [few of which](https://github.com/puppetlabs/puppetlabs-stdlib/tree/main/lib/puppet/functions/stdlib) are namespaced, but I think this is preferred for new function.

@binford2k Wrote a [tool for converting functions](https://binford2k.com/2019/11/19/refactoring-legacy-functions/) written in the legacy API.

The API is documented in detail [here](https://github.com/puppetlabs/puppet-specifications/blob/master/language/func-api.md).

What's your use-case for this BTW?

jcpunk: I've tried to convert to the new style.  I'd swear my code matches the other working ones, but I'm getting a few errors I don't understand....

As for my use case, I'm working in both Ansible and Puppet these days.  I've grown fairly attached to a simple module that lets me define trivial resources in hiera in a similar way to what I'm doing in Ansible.

psudo code
```puppet
class makeit ( Hash $resources = {} ) {
  $resources.each | $type, $value | {
    create_resources($type, $value)
  }
}
```

When I'm pushing out things like the ssl certificate, it would be handy to be able to set `notify => Class['apache']` and `require => Package[ca-certificates]` so that the service gets information about the state change and things happen in the order I want.  Building an entire module infrastructure for each of my system secrets so they can notify the right classes/services/etc feels like overkill.  I could probably trick the ordering, but this would let me be more explicit in my definitions.

alexjfisher: @jcpunk Is it as simple as the converted function now not being in the correct directory?

jcpunk: :) Indeed, that made a world of difference.

jcpunk:
I'm not sure what to do about the windows test failure....

jcpunk:
I was trying to get the test timeout to go away and added an example of how I'd use this.  There are dozens of modules on the forge that have various versions of this pattern.  In theory having this in stdlib would help unify the community and reduce the maintenance burden on various folks.

lib/puppet/functions/stdlib/str2resource.rb
```diff
@@ -0,0 +1,37 @@
...
+    return_type 'Any'
+  end
+
+  def str2resource(res_string)
+    type_name, title = Puppet::Resource.type_and_title(res_string, nil)
+
+    resource = closure_scope.findresource(type_name, title)
+
+    if resource.nil?
+      raise(Puppet::ParseError, "stdlib::str2resource(): could not find #{type_name}[#{title}], this is parse order dependant and values should not be quoted")
```
ekohl, line 32:
```suggestion
      raise(Puppet::ParseError, "stdlib::str2resource(): could not find #{type_name}[#{title}], this is parse order dependent and values should not be quoted")
```

ekohl (commented):
I'm questioning the usefulness of this. Do you have examples where it makes sense?

Also for ordering you can use a collector. I think this should work, whether the file exists or not.
```puppet
Package['mypackage'] -> File <| title == '/foo' |>
```

jcpunk:
> Package['mypackage'] -> File <| title == '/foo' |>

I'm not sure how I'd deference the variables to get the `Package/File` resource for use with the collectors....

david22swan (approved):
@jcpunk This look's good to me so I'm gonna go ahead and merge.
Thank you for putting the work in to get it up and over the line :)

PR #1234, opened 2022-04-20
pdksync - (Maint) PDK Update

david22swan:
(Maint) PDK Update
pdk version: `2.3.0`

### DIFF
--- a/.github/workflows/spec.yml
+++ b/.github/workflows/spec.yml
@@ -54,10 +54,10 @@ jobs:
           buildevents step $TRACE_ID $STEP_ID $STEP_START 'Setup Environment'
           echo STEP_ID=Setup-Acceptance-Test-Matrix >> $GITHUB_ENV
           echo STEP_START=$(date +%s) >> $GITHUB_ENV
-#       - name: Run Static & Syntax Tests
-#         if: ${{ github.repository_owner == 'puppetlabs' }}
-#         run: |
-#           buildevents cmd $TRACE_ID $STEP_ID 'static_syntax_checks' -- bundle exec rake syntax lint metadata_lint check:symlinks check:git_ignore check:dot_underscore check:test_file rubocop
+      - name: Run Static & Syntax Tests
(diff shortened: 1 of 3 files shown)


.github/workflows/spec.yml
```diff
@@ -54,10 +54,10 @@ jobs:
...
           echo STEP_ID=Setup-Acceptance-Test-Matrix >> $GITHUB_ENV
           echo STEP_START=$(date +%s) >> $GITHUB_ENV
-#       - name: Run Static & Syntax Tests
-#         if: ${{ github.repository_owner == 'puppetlabs' }}
-#         run: |
-#           buildevents cmd $TRACE_ID $STEP_ID 'static_syntax_checks' -- bundle exec rake syntax lint metadata_lint check:symlinks check:git_ignore check:dot_underscore check:test_file rubocop
+      - name: Run Static & Syntax Tests
+        if: ${{ github.repository_owner == 'puppetlabs' }}
+        run: |
+          buildevents cmd $TRACE_ID $STEP_ID 'static_syntax_checks' -- bundle exec rake syntax lint metadata_lint check:symlinks check:git_ignore check:dot_underscore check:test_file rubocop
```
ekohl, line 60: I don't really understand why this is so complex. It can be this easy:
```suggestion
          buildevents cmd $TRACE_ID $STEP_ID 'static_syntax_checks' -- bundle exec rake validate lint check rubocop
```

I also submitted a PR for this in https://github.com/puppetlabs/pdk-templates/pull/441 but it looks like this is lost somehow?

metadata.json
```diff
@@ -115,5 +115,5 @@
   "description": "Standard Library for Puppet Modules",
   "pdk-version": "2.3.0",
```
LukasAud, line 116: isn't pdk-version supposed to be 2.4.0?

david22swan: ah, the version updated in between the original commit and when I added the fixes and exclusions. Since this one required no exclusions a second pdksync wasn't run

LukasAud (approved):

PR #1235, opened 2022-04-20
(MODULES-4976) Add windows escaping functions

smortex:
The `shell_escape` function is not usable on windows where batch files and
powershell scripts use different escape rules.

Add two functions to escape strings on windots for batch (`batch_escape`)
and powershell (`powershell_escape`).

### DIFF
--- a/lib/puppet/parser/functions/batch_escape.rb
+++ b/lib/puppet/parser/functions/batch_escape.rb
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+
+#
+# batch_escape.rb
+#
+module Puppet::Parser::Functions
+  newfunction(:batch_escape, type: :rvalue, doc: <<-DOC
+    @summary
+      Escapes a string so that it can be safely used in a batch shell command line.
+
+    @return
+      A string of characters with special characters converted to their escaped form.
+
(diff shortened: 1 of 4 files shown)


bastelfreak (approved):

bastelfreak:
thanks!

smortex:
@alexjfisher probably nothing, I though it makes more sense to have these functions beside `shell_escape()`.

Is there something preventing from moving all those legacy functions to newer ones?

alexjfisher:
> Is there something preventing from moving all those legacy functions to newer ones?

TLDR: Pick some functions you care about and go for it!

Other than time, in most cases no.  There was https://github.com/puppetlabs/puppetlabs-stdlib/pull/1079  Not too sure what happened.  Perhaps it autoclosed when the branch it was targeting got deleted? @binford2k ??  Any thoughts?  That PR was also going to namespace all functions which would have been a breaking change.  If namespacing functions, maybe providing deprecated non-namespaced shims like I did in https://github.com/puppetlabs/puppetlabs-postgresql/blob/7992559aaa219d0f736a6b661d48d2e721c322a1/lib/puppet/functions/postgresql_password.rb would be worth doing.

PR #1236, opened 2022-04-26
Add `xml_encode` function

alexjfisher:
Example for unit test taken from
https://stackoverflow.com/questions/1443816/convert-an-arbitrary-string-to-xml-in-ruby


bastelfreak (approved):

david22swan (approved):
@alexjfisher his change look's good to me, can't see anything wrong with it.
So gonna go ahead and merge, thanks for putting the work in :)

PR #1238, opened 2022-04-27
Modernize escape functions

smortex:
Quoting https://github.com/puppetlabs/puppetlabs-stdlib/pull/1235#issuecomment-1109923594

> What's with these functions using the old API?

Let's fix this!


ekohl (approved):

smortex:
Cc @alexjfisher

This is a quick test to check what needs to be done when modernizing a function. I am about to be AFK for a few weeks but will be happy to help migrate more functions when I am back :wink:

lib/puppet/functions/batch_escape.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+# @summary
+#   Escapes a string so that it can be safely used in a batch shell command line.
+#
+# >* Note:* that the resulting string should be used unquoted and is not intended for use in double quotes nor in single
+# quotes.
+Puppet::Functions.create_function(:batch_escape) do
+  # @return
```
alexjfisher, line 9: Should probably document the parameter here too, and regenerate the REFERENCE.MD to check everything looks good.

lib/puppet/functions/shell_escape.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+require 'shellwords'
```
alexjfisher, line 3: Not sure if this should be moved into the `shell_escape` method.  With newer versions of Ruby it might otherwise trigger https://github.com/puppetlabs/puppet-strings/issues/296

lib/puppet/functions/batch_escape.rb
smortex: Well, I copy-pasted the original comments and REFERENCE.md was unchanged so I was happy with it, but it looks like `bundle exec rake strings:generate:reference` in fact fails and the REFERENCE.md is not updated…  I am digging into this…

lib/puppet/functions/shell_escape.rb
smortex: Ah, I was hesitant to do this in order to avoid including code at load time but only when used.  Another good reason to do so, i will update it!

lib/puppet/functions/batch_escape.rb
smortex: I will track this issue in another PR.

smortex: I updated #1239 to improve this and avoid it in the future :wink:

david22swan (approved):
Changes LGTM
Thanks for the work :)

PR #1240, opened 2022-05-02
Switch parsejson() from PSON to JSON parsing

seanmil:
The use of PSON is deprecated and mostly removed from Puppet.
Switch this function from using PSON to a real JSON parsing
library via the MultiJson helper in Puppet::Util::Json.

### DIFF
--- a/lib/puppet/parser/functions/parsejson.rb
+++ b/lib/puppet/parser/functions/parsejson.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require 'puppet/util/json'
 #
 # parsejson.rb
 #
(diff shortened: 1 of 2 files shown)


chelnak:
Hey @seanmil, are you still interested in progressing with this Pull Request?

It looks like there is a question from @david22swan to address.

seanmil:
> Apologies but could you provide some evidence regarding the said deprecation?
> From what I can tell PSON is still included within the newest versions of puppet.

@david22swan Sorry, I could have phrased that a lot better. What I meant was that unless I'm misremembering, back in the distant past everything in Puppet which used JSON used PSON instead, even if it was supposed to be real JSON and that using PSON instead of JSON for things which should be JSON has been not in favor for quite a while.

Regardless though, that is not really the point. The point is that the function name is "parsejson", but actually parses PSON, which doesn't parse all the JSON which the JSON functions will parse (see the updated spec examples).

If there is a desire/need for parsepson then that should be introduced as a separate function.

Thanks!

david22swan (approved):
LGTM

Logic behind this makes sense to me so I'm happy to merge. Have put up a PR to create a specific `parsepson` function:
https://github.com/puppetlabs/puppetlabs-stdlib/pull/1259

PR #1241, opened 2022-05-09
Update load_module_metadata.rb to correct capitalisation in strings documentartion

davidsandilands:
Just a simple capitalisation error on usage


chelnak:
Good spot!

chelnak (approved):

PR #1244, opened 2022-05-20
Convert `ensure_packages` to new API and refactor

alexjfisher:
As a first step before fixing https://github.com/puppetlabs/puppetlabs-stdlib/pull/1196 in a separate PR, this PR converts to the new API and gets rid of some code duplication etc. in the process.

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -60,6 +60,7 @@ or the default value if nothing was found.
 * [`empty`](#empty): **Deprecated:** Returns true if the variable is empty.
 * [`enclose_ipv6`](#enclose_ipv6): Takes an array of ip addresses and encloses the ipv6 addresses with square brackets.
 * [`ensure_packages`](#ensure_packages): Takes a list of packages and only installs them if they don't already exist.
+* [`ensure_packages`](#ensure_packages): Deprecated 3x version of the `ensure_packages` function
 * [`ensure_resource`](#ensure_resource): Takes a resource type, title, and a list of attributes that describe a
 resource.
 * [`ensure_resources`](#ensure_resources): Takes a resource type, title (only hash), and a list of attributes that describe a
@@ -1966,17 +1967,41 @@ Returns: `Any` encloses the ipv6 addresses with square brackets.
 
 ### <a name="ensure_packages"></a>`ensure_packages`
 
-Type: Ruby 3.x API
+Type: Ruby 4.x API
 
 It optionally takes a hash as a second parameter that will be passed as the
 third argument to the ensure_resource() function.
+
+#### `ensure_packages(Variant[String[1], Array[String[1]], Hash[String[1], Any]] $packages, Optional[Hash] $default_attributes)`
+
+It optionally takes a hash as a second parameter that will be passed as the
+third argument to the ensure_resource() function.
+
+Returns: `Undef` Returns nothing.
+
+##### `packages`
+
+Data type: `Variant[String[1], Array[String[1]], Hash[String[1], Any]]`
+
+The packages to ensure are installed. If it's a Hash it will be passed to `ensure_resource`
+
+##### `default_attributes`
+
+Data type: `Optional[Hash]`
+
+Default attributes to be passed to the `ensure_resource()` function
+
+### <a name="ensure_packages"></a>`ensure_packages`
+
+Type: Ruby 3.x API
+
+Deprecated 3x version of the `ensure_packages` function
 
 #### `ensure_packages()`
 
-It optionally takes a hash as a second parameter that will be passed as the
-third argument to the ensure_resource() function.
+The ensure_packages function.
 
-Returns: `Any` install the passed packages
+Returns: `Any`
 
 ### <a name="ensure_resource"></a>`ensure_resource`
 
--- a/lib/puppet/functions/ensure_packages.rb
+++ b/lib/puppet/functions/ensure_packages.rb
@@ -4,19 +4,20 @@
 #
 # It optionally takes a hash as a second parameter that will be passed as the
 # third argument to the ensure_resource() function.
-Puppet::Functions.create_function(:ensure_packages) do
+Puppet::Functions.create_function(:ensure_packages, Puppet::Functions::InternalFunction) do
   # @param packages
   #   The packages to ensure are installed. If it's a Hash it will be passed to `ensure_resource`
   # @param default_attributes
   #   Default attributes to be passed to the `ensure_resource()` function
   # @return [Undef] Returns nothing.
   dispatch :ensure_packages do
+    scope_param
     param 'Variant[String[1], Array[String[1]], Hash[String[1], Any]]', :packages
     optional_param 'Hash', :default_attributes
     return_type 'Undef'
   end
 
-  def ensure_packages(packages, default_attributes = nil)
+  def ensure_packages(scope, packages, default_attributes = nil)
     if default_attributes
       defaults = { 'ensure' => 'installed' }.merge(default_attributes)
       if defaults['ensure'] == 'present'
@@ -27,10 +28,10 @@ Puppet::Functions.create_function(:ensure_packages) do
     end
 
     if packages.is_a?(Hash)
-      call_function('ensure_resources', 'package', packages.dup, defaults)
+      scope.call_function('ensure_resources', ['package', packages.dup, defaults])
     else
       Array(packages).each do |package_name|
-        call_function('ensure_resource', 'package', package_name, defaults)
+        scope.call_function('ensure_resource', ['package', package_name, defaults])
       end
     end
     nil
--- a/lib/puppet/functions/ensure_packages.rb
+++ b/lib/puppet/functions/ensure_packages.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+
+# @summary Takes a list of packages and only installs them if they don't already exist.
+#
+# It optionally takes a hash as a second parameter that will be passed as the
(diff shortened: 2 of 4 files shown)


REFERENCE.md
```diff
@@ -1891,17 +1891,29 @@ Returns: `Any` encloses the ipv6 addresses with square brackets.
...
 third argument to the ensure_resource() function.
 
-#### `ensure_packages()`
+#### `ensure_packages(Variant[Hash,String[1],Array[String[1]]] $packages, Optional[Hash] $default_attributes)`
 
 It optionally takes a hash as a second parameter that will be passed as the
 third argument to the ensure_resource() function.
 
-Returns: `Any` install the passed packages
+Returns: `Any`
```
ekohl, line 1904: Is there are return type for this?

lib/puppet/functions/ensure_packages.rb
```diff
@@ -0,0 +1,37 @@
...
+    else
+      defaults = { 'ensure' => 'installed' }
+    end
+
+    if packages.is_a?(Hash)
+      call_function('ensure_resources', 'package', packages.dup, defaults)
+    else
+      packages = Array(packages)
+
+      packages.each do |package_name|
```
ekohl, line 32: Shouldn't this be `map()` if you want a return type?

REFERENCE.md
alexjfisher: Is there a way to specify that the function doesn't return anything useful?

lib/puppet/parser/functions/ensure_packages.rb
```diff
@@ -1,54 +0,0 @@
-# frozen_string_literal: true
-
-#
-# ensure_packages.rb
-#
-module Puppet::Parser::Functions
-  newfunction(:ensure_packages, type: :statement, doc: <<-DOC
```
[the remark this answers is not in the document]
ekohl: Interesting, I never knew that.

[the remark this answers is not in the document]
ekohl: Would that actually make it a possibly breaking change? Would people out there have used `ensure_packages ['mypackage']` and now need to write `ensure_packages(['mypackage'])`?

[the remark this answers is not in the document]
alexjfisher: It was a 'statement' function in so much that it didn't return anything.  It can't be called without parenthesis though as it doesn't appear in this list. https://github.com/puppetlabs/puppet/blob/ac45fe2b783708f64cb3a28f21ee92673e1f3216/lib/puppet/pops/model/factory.rb#L969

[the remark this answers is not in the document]
alexjfisher: Done.  I've also added a 3x shim.  I doubt it'll get any use for this function, so could drop it again if people prefer.  There are other functions (like ensure_resource) where it would possibly be more beneficial should we ever convert those.

ekohl (approved):

lib/puppet/functions/ensure_packages.rb
```diff
@@ -0,0 +1,40 @@
...
```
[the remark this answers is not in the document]
alexjfisher: Updated.  Thanks.

lib/puppet/parser/functions/ensure_packages.rb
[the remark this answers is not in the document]
ekohl: It was confirmed by @hlindberg that there is always a return value and being explicit in `nil` / `undef` is the correct way to go. In API v3 `type: :statement` did essentially the same.

lib/puppet/functions/ensure_packages.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+# @summary Takes a list of packages and only installs them if they don't already exist.
```
ekohl, line 3: Technically you can also ensure they're absent via `default_attributes`, right?

lib/puppet/functions/ensure_packages.rb
```diff
@@ -0,0 +1,38 @@
...
+#
+# It optionally takes a hash as a second parameter that will be passed as the
+# third argument to the ensure_resource() function.
+Puppet::Functions.create_function(:ensure_packages) do
+  # @param packages
+  #   The packages to ensure are installed
```
ekohl, line 9:
```suggestion
  #   The packages to ensure are installed. If it's a Hash it will be passed to ensure_resource
```

ekohl (approved):
Code wise :+1: just some documentation suggestions. I also wonder if `@see ensure_resource` could be added, but I'm not 100% sure puppet-strings properly supports that.

lib/puppet/functions/ensure_packages.rb
alexjfisher: yes, or by specifying a hash for the first parameter.  But... I really don't know why people do that.  IMO it becomes a fairly  useless function the moment you start specifying attributes as they likely won't match with any other definition you already have in the catalog for that package.

lib/puppet/functions/ensure_packages.rb
alexjfisher: Updated.

lib/puppet/functions/ensure_packages.rb
ekohl: Perhaps `manages` instead of `installs` is better? Also, due to the behavior of `ensure_resource(s)` the state must match. So perhaps `ensures` is even better.
```suggestion
# @summary Takes a list of packages and ensures they exist
```

hlindberg:
I suspect this function needs to be an `InternalFunction` (see function `contain`, `require` in puppet source how that is done. It also needs to capture the calling scope (like those functions do), and then make the call to `ensure_resource` using that scope. The risk is otherwise that a resource won't show up with the correct containment (where it came from) in case it was created by the call to `ensure_packages`. I think you need to add a test case for that as well, and I suspect it will show that scope must be passed on.

ekohl:
Containment is an interesting point. I think today it only does something if the resource was not previously defined. Otherwise it **may** be contained (I don't know). So it's already ambiguous where the package resource ends up. I've always written my code by assuming it was **not** contained added explicit `require` properties if needed.

You could add in containment, but I wonder if that could cause risks with dependency cycles. I'm thinking about this because the package may not be contained within two classes.

alexjfisher:
Good call.  I think it was contained, by the calling class and now it isn't.

alexjfisher:
See https://gist.github.com/alexjfisher/b0287dc84a5ce451093d43f2de16f4ed

I'll look into making this an Internal function to preserve the existing behaviour.  Not sure how to write a test case for this.  @hlindberg any pointers? Do I just do `that_comes_before('Whit[Completed_class[Foo]]')` ???

hlindberg:
@alexjfisher sorry, don't know how to write that test.

alexjfisher:
> @alexjfisher sorry, don't know how to write that test.

No luck so far.  The meta `whit` resources don't even appear to be in the catalogue object rspec-puppet generates.

hlindberg:
I also think the function to be called (with the scope passed) needs to be called a special way - IIRC, there should be something like `call_function_with_scope` available to the internal-function. Unfortunately, the `contain` and `require` functions in puppe where optimized to instead of calling `include` they now do what include did as it was faster - so you cannot immediately see how it can be done. You may be able to look at the history of those functions to see how they used to make "call `include` with given scope".

alexjfisher:
Not just `scope.call_function('ensure_resource', ['package', package_name, defaults])` instead of `call_function('ensure_resource', 'package', package_name, defaults)` ?

I'll need to check on Monday morning, but I think that's what my uncommitted change I was working on looks like.  _Think_ it works.  Need to double check.

binford2k:
I'm going to vote pretty hard that we do not try to maintain containment artifacts. At best, the existing behavior is unspecified and depending on its implementation details already causes surprising failures.

Given the [link](https://gist.github.com/alexjfisher/b0287dc84a5ce451093d43f2de16f4ed) that @alexjfisher posted earlier, the graph will already be different if someone includes the `tzdata` package somewhere else, either explicitly or with `ensure_packages()` and it _happens to be_ declared first.

Imho, that's the whole point of this function. It explicitly says "I don't care how this package gets declared, just make sure it's in the catalog somewhere."

ekohl:
I do agree with @binford2k, with the note that changing it in a minor/patch version might be surprising. Breaking implicit behavior still is painful for users.

alexjfisher:
Thanks for the feedback.

To summarise...
* I'll drop https://github.com/puppetlabs/puppetlabs-stdlib/pull/1244/commits/b565f7f097c05a62ec44e4a729c05da43ba9092c * Mark this a breaking change
* Update the first comment with a more user friendly summary of why it might be breaking in some cases

Once merged, I can then get on with coming up with a fix for making it work no matter if any existing package resource has `ensure => present`, or `ensure => installed`.

ekohl:
I'd be happy with merging it as is, releasing it in a minor version and then doing the breaking change later. Major version bumps in stdlib are painful so I'd rather combine a few things.

For example, perhaps we should deprecate the compat data types and drop them in the next major version (looking at you https://github.com/puppetlabs/puppetlabs-stdlib/tree/main/types/compat).

chelnak:
@ekohl are you still good for this change to be merged?

ekohl (approved):
@chelnak yes, let's get this in.

chelnak:
@ekohl Great thanks!

Issue #1245, opened 2022-05-23
Stdlib::Compat::Ipv4 and Stdlib::Compat::Ipv6 are not consistent about validation of IP networks


frank-f:
## Use Case
- Stdlib::Compat::Ipv4 considers 192.168.0.1 and 192.168.0.0/24 to be valid
- Stdlib::Compat::Ipv6 however only accepts single addresses like 2001:db8:1:2::3, but not a network like 2001:db8:1:2::/64

The Apache module uses Stdlib::Compat::Ipv4 and Stdlib::Compat::Ipv6 to validate IP addresses before including them into the mod_remoteip trusted proxy list. Therefore adding an entire IPv4 network works fine, but an IPv6 network does not. Both are valid settings for mod_remoteip itself.

## Describe the Solution You Would Like
The current behavior does not make sense to me, because the IPv6 space is much larger and therefore one would benefit much more from the possibility to use entire IPv6 networks as opposed to IPv4.
I'll submit a PR later that will add support for /xxx at the end of an IPv6 address and would be really delighted to see it included.

## Describe Alternatives You've Considered
1. Add all possible addresses individually - this would work if it's just a few, but even adding an entire /112 network is obviously out of the question, let alone larger networks
2. Ask/modify the Apache module to use a different validation method. Doesn't seem sensible to me - why write an own validation method if stdlib provides one that just needs to be tweaked?

b4ldr:
the Stdlib::Compat resources where added, AFAIK, to be compatible with the validate_ipaddress functions and as such i don't think there implementation will change.   Any new code should use Stdlib::IP::Address and the types in there e.g.
 * Stdlib::IP::Address::V6
 * Stdlib::IP::Address::V6::Cidr
 * Stdlib::IP::Address::V4
 * Stdlib::IP::Address::V4::Nosubnet
 
As such i would say this is a bug in the apache modules which should switch to theses newer types

frank-f:
Alright, thank you for the heads up. I'll submit a PR to the apache module repository then.

PR #1246, opened 2022-05-23
#1245 Accept IPv6 networks in CIDR notation

frank-f:
Make the behavior of Stdlib::Compat::Ipv6 consistent with Stdlib::Compat::Ipv4 by adding the possibility to specify IPv6 networks in CIDR notation, for example 2001:db8:1:2::/64


frank-f:
Is it an absolute must to commit with an email address? I purposely leave it blank on public commits.

frank-f:
Alright, thank you for the heads up. I'll submit a PR to the apache module repository. :)

PR #1250, opened 2022-06-06
Simplify stdlib::manage

jcpunk:
This should make the resources much less sensitive to parse order,
since pupppet is already "doing the right thing" to Strings in this
parameter location.

### DIFF
--- a/spec/classes/manage_spec.rb
+++ b/spec/classes/manage_spec.rb
@@ -39,6 +39,6 @@ describe 'stdlib::manage' do
 
     it { is_expected.to compile }
     it { is_expected.to contain_file('/etc/motd.d/hello').with_content('I say Hi').with_notify('Service[sshd]') }
-    it { is_expected.to contain_package('example').with_ensure('installed').with_subscribe(['Service[sshd]', 'File[/etc/motd.d]']) }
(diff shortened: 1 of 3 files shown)


petergmurphy (approved):

spec/classes/manage_spec.rb
```diff
@@ -38,6 +39,6 @@
 
     it { is_expected.to compile }
     it { is_expected.to contain_file('/etc/motd.d/hello').with_content('I say Hi').with_notify('Service[sshd]') }
-    it { is_expected.to contain_package('example').with_ensure('installed') }
+    it { is_expected.to contain_package('example').with_ensure('installed').with_subscribe(['Service[sshd]', 'File[/etc/motd.d]']) }
```
ekohl, line 42: It is better to use `that_subscribes_to` instead of `with_subscribe`. The former checks the relation exists in the catalog, the latter checks whether the attribute is present.
```suggestion
    it { is_expected.to contain_package('example').with_ensure('installed').that_subscribes_to(['Service[sshd]', 'File[/etc/motd.d]']) }
```

ekohl (changes requested):

ekohl (approved):
Code wise :+1: but I'm not sure what's up with the acceptance tests, but I also saw it in a different puppetlabs repo. Infrastructure issue?

PR #1251, opened 2022-06-15
(MODULES-2892) Handle missing file in file_line

silug:
In noop mode, file_line should not fail if the file does not exist.

MODULES-2892 #close

### DIFF
--- a/lib/puppet/provider/file_line/ruby.rb
+++ b/lib/puppet/provider/file_line/ruby.rb
@@ -83,6 +83,9 @@ Puppet::Type.type(:file_line).provide(:ruby) do
   rescue TypeError => _e
     # Ruby 1.8 doesn't support open_args
     @lines ||= File.readlines(resource[:path])
+  rescue Errno::ENOENT => _e
+    raise _e unless resource.noop?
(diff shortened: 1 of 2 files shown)


lib/puppet/provider/file_line/ruby.rb
```diff
@@ -83,6 +83,9 @@ def lines
   rescue TypeError => _e
     # Ruby 1.8 doesn't support open_args
     @lines ||= File.readlines(resource[:path])
+  rescue Errno::ENOENT => _e
+    raise _e unless resource.noop?
```
ekohl, line 87: I think this is the best way to reraise the exception:
```suggestion
  rescue Errno::ENOENT
    raise unless resource.noop?
```

ekohl (commented):
Slight nit on coding style, otherwise :+1:

lib/puppet/provider/file_line/ruby.rb
silug: Change made.  Thanks for the review.

ekohl (approved):

PR #1253, opened 2022-07-01
deferrable epp function simplifying deferred templates

binford2k:


reidmv:
I think this is the right UX for the purpose; basically an arg-for-arg wrapper for `epp()`. 👍 

-------------------

Wild future improvement idea:

If we like this idea, there's a possibility we could modify the internals a bit to sub out the deferred arguments for `%<deferred_1>s`, `%<deferred_2>s`, and so forth, run the `epp()` with those subs, then return a `Deferred['sprintf', [$deferred_args_hash]]` – do this instead of returning a `Deferred['inline_epp' ...]`. If we did that, we might eliminate the template and variable scope constraints this implementation warns against.

chelnak (approved):

chelnak:
@binford2k Looks like you have a failing lint check

```
functions/deferrable_epp.pp - ERROR: there should be a single newline at the end of a manifest on line 19 (check: manifest_whitespace_double_newline_end_of_file)
```

Also I wonder if we should have a spec test for this?

spec/functions/stdlib_deferrable_epp_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+  context 'defers rendering with deferred input' do
+    let(:pre_condition) do (<<~END)
+        function epp($str, $data) { fail("should not have invoked epp()") }
+        function find_template($str) { return "path" }
+        function file($path) { return "foo: <%= foo %>" }
+      END
+    end
+    it {
+      foo = Puppet::Pops::Types::TypeFactory.deferred.create('join', [1,2,3])
+      is_expected.to run.with_params('mymod/template.epp', {'foo' => foo }) #.and_return(a_kind_of Puppet::Pops::Evaluator::DeferredValue)
```
binford2k, line 22: @joshcooper I couldn't quite get the return type here. Suggestions?

binford2k: Never mind. I figured out the return type and submitted an rspec-puppet PR to actually allow this matcher to work.

https://github.com/puppetlabs/rspec-puppet/pull/24

chelnak (approved):

PR #1259, opened 2022-07-25
(FEAT) Add function parsepson

david22swan:
This commit adds a new function to parse PSON formatted JSON.
This functionality was formally covered by `parsejson` which, despite its name, defaulted to PSON.
However this has been changed so that the function now accurately uses JSON, with this function covering the potential need for PSON too specifically be used.

### DIFF
--- a/lib/puppet/functions/parsepson.rb
+++ b/lib/puppet/functions/parsepson.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+
+# @summary
+#   This function accepts PSON, a Puppet variant of JSON, as a string and converts
+#   it into the correct Puppet structure
+#
+# @example How to parse pson
+#   $data = parsepson('{"a":"1","b":"2"}')
+#
+# For more information on PSON please see the following link:
+# https://puppet.com/docs/puppet/7/http_api/pson.html
+#
+Puppet::Functions.create_function(:parsepson) do
+  # @param pson_string A valid PSON string
+  # @param default An optional default to return if parsing the pson_string fails
+  # @return [Data]
+  dispatch :parsepson do
+    param          'String[1]', :pson_string
+    optional_param 'Any',       :default
+  end
+
+  def parsepson(pson_string, default = :no_default_provided)
+    PSON.load(pson_string)
+  rescue StandardError => err
+    Puppet.debug("Parsing PSON failed with error: #{err.message}")
+    raise err if default == :no_default_provided
+    default
+  end
+end
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -140,6 +140,8 @@ true boolean.
 Puppet structure
 * [`parsejson`](#parsejson): This function accepts JSON as a string and converts it into the correct
 Puppet structure.
+* [`parsepson`](#parsepson): This function accepts PSON, a Puppet variant of JSON, as a string and 
+converts it into the correct Puppet structure.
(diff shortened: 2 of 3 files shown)


ekohl (changes requested):
Could this be written as a modern function rather than the legacy one? You also have `parsejson` in the PR title and commit message, which should be `parsepson`.

david22swan:
Had just copied the original function pre changes, but modernizing it would probably be good.
As for the PR and Commit................ yeh
It was a long day

lib/puppet/functions/parsepson.rb
```diff
@@ -0,0 +1,32 @@
...
+
+#
+# parsepson.rb
+#
+# @summary
+#   This function accepts PSON, a Puppet variant of JSON, as a string and converts
+#   it into the correct Puppet structure
+#
+# @example How to parse pson
+#   $data = parsepson("{any valid pson: string}")
```
ekohl, line 11: Is this valid pson? Looks more like hocon.

lib/puppet/functions/parsepson.rb
```diff
@@ -0,0 +1,32 @@
...
+#
+# For more information on PSON please see the following link:
+# https://puppet.com/docs/puppet/7/http_api/pson.html
+#
+Puppet::Functions.create_function(:parsepson) do
+  # @param pson_string A valid PSON string
+  # @param default An optional default to return if parsing the pson_string fails
+  # @return [Data]
+  dispatch :parsepson do
+    param          'String', :pson_string
```
ekohl, line 21: Since an empty string is invalid, should this be at least non-empty?
```suggestion
    param          'String[1]', :pson_string
```

lib/puppet/functions/parsepson.rb
david22swan: My bad, copied the wrong code.
Took from the hocon test

lib/puppet/functions/parsepson.rb
david22swan: Updated

ekohl (dismissed):

lib/puppet/functions/parsepson.rb
david22swan: Have walked myself back on this a bit.
Setting it to a `String[1]` overrides the ability to set a default return value.
Since I'm not entirely sure of all the circumstances that the setting of this default would be valuable I think it's best to leave it as a standard `String` value.

pmcmaw (approved):
Approving as failing spec tests are currently present on main.
Following ekohls suggestion of using the newer function style this is something on the teams radar and will be addressed as soon as we have bandwidth.

ekohl:
> Following ekohls suggestion of using the newer function style this is something on the teams radar and will be addressed as soon as we have bandwidth.

That's good, but IMHO we should at least only add new style functions (as we did here). Conversion is lower priority.

alexjfisher:
> > Following ekohls suggestion of using the newer function style this is something on the teams radar and will be addressed as soon as we have bandwidth.
> 
> That's good, but IMHO we should at least only add new style functions (as we did here). Conversion is lower priority.

And possibly also only allow new functions if they are namespaced too?

ekohl:
I feel stdlib is always a bit different. All other modules should certainly only have namespaced functions but stdlib is in this weird place. Perhaps best to discuss it in its own issue?

Issue #1262, opened 2022-08-08
to_toml function is not working


skoef:
## Describe the Bug
With stdlib 8.2.0, when using the `to_toml` function, introduced in 8.1.0, I get the following error message:

```
Internal Server Error: org.jruby.exceptions.LoadError: (LoadError) no such file to load -- puppet_x/stdlib
```

looking up the trace in puppetserver, I ended up on [this line](https://github.com/puppetlabs/puppetlabs-stdlib/blob/main/lib/puppet_x/stdlib/toml_dumper.rb#L21) where the problem would be. Googling a potential solution, I found that @mrbanzai [solved the issue](https://github.com/puppetlabs/puppetlabs-stdlib/commit/7dde528bfc9317474c21653cc58ddfd38ece7710) on a fork of stdlib and it works for me as well. 

If the proposed fix is considered okay, could we upstream that fix perhaps?

## Expected Behavior

With the fix applied, the catalog just applies and TOML is properly formatted.

## Steps to Reproduce
1. Use `to_toml` while on stdlib 8.2.0

## Environment
 - Version: 8.2.0
 - Platform: Ubuntu 20.04
 - Puppetserver: 7.4.2

david22swan:
PR with this fix has been created: https://github.com/puppetlabs/puppetlabs-stdlib/pull/1264

PR #1264, opened 2022-08-15
(GH-1262) Use 'require_relative' to load stdlib due to lookup errors

david22swan:


ekohl (dismissed):
This is a common problem. I wonder if we can somehow find a CI check for such a thing. No idea what the right way is and which tool to use, but users should be able to find these kind of problems early.

LukasAud (approved):

PR #1267, opened 2022-09-27
Add a Stdlib::CreateResources type

ekohl:
It is very common to have a parameter which creates resources, either via `create_resources()` or via iteration.


bastelfreak (approved):

david22swan:
@ekohl This look's good to me.
It need's a rebase before I can merge it though.

chelnak:
I'm a fan of types being singular nouns. I wonder if calling this type 'resource' would be too ambiguous?

ekohl:
> I'm a fan of types being singular nouns. I wonder if calling this type 'resource' would be too ambiguous?

This describes multiple resources. You can't get away from it IMHO. For example, you also have [ensure_resource](https://github.com/puppetlabs/puppetlabs-stdlib/blob/main/lib/puppet/parser/functions/ensure_resource.rb) and [ensure_resources](https://github.com/puppetlabs/puppetlabs-stdlib/blob/main/lib/puppet/parser/functions/ensure_resources.rb). If you were to define types for that, you'd also have one singular and one plural.

Technically you have this:
```puppet
type Stdlib::ResourceAttribute = Any
type Stdlib::ResourceAttributes = Hash[String[1], Stdlib::ResourceAttribute]
type Stdlib::CreateResources = Hash[String[1], Stdlib::ResourceAttributes]
```

But that's really verbose.

Also, you could also narrow down the title to a `Pattern` that's much smaller, but I'm not sure how much you'd win.

chelnak:
Yeah fair enough. I can see your point here. 👍

david22swan (approved):
LGTM

david22swan:
Thank's for getting the rebase done so fast :)

PR #1268, opened 2022-10-04
facter: do not use dummy package instantiation on windows

rismoney:


bastelfreak:
can you explain why you propose this change?

rismoney:
I believe (still reviewing) that this dummy package instantiation is causing excessive delays when run over a winrm connection.
The answer is always 'windows' as I don't believe you can change the package provider on windows before facter runs.
Here is what I see using facter -p --timing:
fact 'package_provider', took: (16.278307) seconds
when every other fact is sub-second.

rismoney:
not sure if I put this in the right spot.
https://tickets.puppetlabs.com/browse/PUP-11647?filter=-2

bastelfreak:
maybe it is worth investigating why ot takes so long (if that is always the cas). I know that many people use chololatey on windows, but I am not sure if people set that as default provider.

rismoney:
you can't set the default provider to chocolatey before puppet runs.  If you set Package {provider=chocolatey} that occurs after facter.  so I am not sure what this fact accomplishes on windows. Personally i think the fact is a bit 'dumb', as it only confirms what puppet already knew.

rismoney:
>  custom provider, like the chocolety one, could overwrite the default provider that puppet core sets.

Can you substantiate that?  If that is the case, I will abandon this effort.

jordanbreen28:
Hi @rismoney - as you have noted, a custom provider cannot override the default. It can claim to be default and then Puppet will "randomly" choose one and warn you about it. 

Unfortunately, there seems to be test failures for this PR - could you look to resolve them and then rebase with the current main so we can proceed? :-)

lib/facter/package_provider.rb
```diff
@@ -14,6 +14,11 @@
 
 # These will be nil if Puppet is not available.
 Facter.add(:package_provider) do
+
+  confine do
+    Facter.value(:kernel) != 'windows'
+  end
```
ekohl, line 20: You can even use weights. It still works on Windows so just give this a weight of 50:
```suggestion
  has_weight 50
```

lib/facter/package_provider.rb
```diff
@@ -23,3 +28,16 @@
     end
   end
 end
+
+Facter.add(:package_provider) do
+
+  confine do
+    Facter.value(:kernel) == 'windows'
+  end
```
ekohl, line 36: And then give this a higher weight (like 100) and confine it to windows.
```suggestion
  has_weight 100
  confine kernel: 'windows'
```

lib/facter/package_provider.rb
```diff
@@ -14,6 +14,11 @@
 
 # These will be nil if Puppet is not available.
 Facter.add(:package_provider) do
+
```
ekohl, line 17: In our coding style we don't have this empty line after a `do` block starts.
```suggestion
```

lib/facter/package_provider.rb
```diff
@@ -23,3 +28,16 @@
     end
   end
 end
+
+Facter.add(:package_provider) do
+
```
ekohl, line 33:
```suggestion
```

spec/unit/facter/package_provider_spec.rb
```diff
@@ -40,6 +40,15 @@
...
       end
+
+      context 'when windows' do
+        it 'returns windows' do
+          provider = Puppet::Type.type(:package).provider(:windows)
+          allow(Puppet::Type.type(:package)).to receive(:defaultprovider).and_return(provider)
+          expect(Facter.fact(:package_provider).value).to eq('windows')
+        end
+      end
+
```
ekohl, line 51:
```suggestion
```

ekohl (changes requested):

jordanbreen28:
Hi @rismoney,
As there has been no movement on this PR since October 4th, I will proceed to close this request.
Once you have the bandwidth to pick this back up again, and review the comments above, please do re-open this request and we can think about proceeding!
Thanks for your continued hard work on this!

PR #1275, opened 2022-10-12
(CONT-200) Fix require relative paths

chelnak:
Prior to this PR using stdlib/to_toml function in a separate environment outside of production will result in the following error if the latest version doesn't also live in production:

```
Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Internal Server Error: org.jruby.exceptions.LoadError: (LoadError) no such file to load -- puppet_x/stdlib
```

This PR fixes the issue by updating code to use the correct paths for require_relative.

### DIFF
--- a/lib/puppet_x/stdlib/toml_dumper.rb
+++ b/lib/puppet_x/stdlib/toml_dumper.rb
@@ -18,7 +18,7 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
 
-require_relative '../stdlib'
+require_relative '../../puppet_x/stdlib'
 require 'date'
 
 module PuppetX::Stdlib


lib/puppet_x/stdlib.rb
```diff
@@ -1,4 +1,4 @@
-require 'puppet_x'
+require_relative '../puppet_x'
```
bastelfreak, line 1: is there maybe a rubocop rule we can use to enforce this?

bastelfreak (approved):

lib/puppet_x/stdlib.rb
chelnak: I've not seen one.. it would be cool if there was though. Even something to warn and say "maybe you should use require_relative here" would be nice.

pmcmaw (approved):
Failing tests are not in relation to this change.

PR #1277, opened 2022-10-14
(CONT-173) - Updating deprecated facter instances

jordanbreen28:
Prior to this PR, this module contained instances of Facter::Util::Resolution.exec and Facter::Util::Resolution.which, which are deprecated. This PR aims to replace these exec helpers with their supported Facter::Core::Execution counterparts.

This PR:

Replaces all Facter::Util::Resolution instances with corresponding Facter::Core::Execution exec helpers

### DIFF
--- a/lib/facter/root_home.rb
+++ b/lib/facter/root_home.rb
@@ -9,7 +9,7 @@ module Facter::Util::RootHome
   class << self
     # determines the root home directory
     def returnt_root_home
-      root_ent = Facter::Util::Resolution.exec('getent passwd root')
+      root_ent = Facter::Core::Execution.execute('getent passwd root')
       # The home directory is the sixth element in the passwd entry
       # If the platform doesn't have getent, root_ent will be nil and we should
       # return it straight away.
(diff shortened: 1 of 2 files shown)


lib/facter/root_home.rb
```diff
@@ -39,7 +39,7 @@ def returnt_root_home
   confine kernel: :aix
   root_home = nil
```
ekohl, line 40: This line is weird and should really be inside `setcode`.

ekohl (commented):
Argh, so many unrelated issues. I wonder why this isn't using `Etc.getpwuid(0).dir`, at least on Linux or even POSIX.

Also, I wonder why this isn't restricted to not run on Windows.

chelnak:
@ekohl Nice spot with those finds. Would you mind creating an issue so we can track them?

chelnak (approved):
Changes are sound.

@ekohl raises some good points too. We can address them in another PR.

ekohl:
I wonder why this was merged since https://github.com/puppetlabs/puppetlabs-stdlib/pull/1278 was pretty trivial to write.

PR #1278, opened 2022-10-14
Determine root_home without shelling out

ekohl:
This uses Ruby's Etc library to determine the root home. This avoids shelling out and writing multiple implementations.

### DIFF
--- a/lib/facter/root_home.rb
+++ b/lib/facter/root_home.rb
@@ -1,49 +1,11 @@
 # frozen_string_literal: true
 
-# root_home.rb
-module Facter::Util::RootHome
-  # @summary
-  #   A facter fact to determine the root home directory.
-  #   This varies on PE supported platforms and may be
-  #   reconfigured by the end user.
-  class << self
-    # determines the root home directory
-    def returnt_root_home
-      root_ent = Facter::Core::Execution.execute('getent passwd root')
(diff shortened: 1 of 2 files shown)


lib/facter/root_home.rb
```diff
@@ -1,49 +1,13 @@
...
-    hash = {}
-    str.split("\n").each do |pair|
-      key, value = pair.split(%r{:})
-      hash[key] = value
-    end
-    hash['dir'].strip
-  end
-end
-
 Facter.add(:root_home) do
```
chelnak, line 3: Nice removal of code @ekohl.

What do you think about confining the fact to osx and linux? I noticed the test just checks for `nil` when `:kernel` is Windows.

ekohl: Today there's also fact for AIX and I'd expect this code to work on any unix (like Solaris). However, I can't test that since I don't have access to those systems. Any you could confine it to non-Windows but I think this is effectively the same.

bastelfreak (approved):

lib/facter/root_home.rb
chelnak: Fair point!

jordanbreen28 (approved):
Looks good! Thanks @ekohl.

PR #1280, opened 2022-10-16
Drop Puppet < 3.6 support in package_provider fact

ekohl:
The metadata states it's only compatible with Puppet 6+ so maintaining compatibility is useless and can only slow things down.


bastelfreak (approved):

chelnak (approved):

chelnak:
Perfect!

PR #1281, opened 2022-10-19
Add Stdlib::Ensure::Package type

arjenz:
Like the ones for file, service.
Valid values taken from https://puppet.com/docs/puppet/7/types/package.html#package-attribute-ensure

### DIFF
--- a/spec/type_aliases/ensure_package_spec.rb
+++ b/spec/type_aliases/ensure_package_spec.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe 'Stdlib::Ensure::Package' do
+  describe 'valid handling' do
+    [
+      'present',
+      'absent',
+      'purged',
+      'disabled',
+      'installed',
+      'latest',
+      '1',
+      '1.1',
+      '>=6.0',
+    ].each do |value|
+      describe value.inspect do
+        it { is_expected.to allow_value(value) }
+      end
+    end
+  end
+  describe 'No complex types can match' do
+    context 'Garbage inputs, no complex or non string types can match' do
+      [
+        1,
+        1.1,
+        [1.1],
+        '',
+        { 'foo' => 'bar' },
+        {},
+      ].each do |value|
+        describe value.inspect do
+          it { is_expected.not_to allow_value(value) }
+        end
+      end
+    end
+  end
+end
--- a/types/ensure/package.pp
+++ b/types/ensure/package.pp
@@ -1,0 +1,2 @@
+# @summary Validate the value of the ensure parameter for a package
+type Stdlib::Ensure::Package = Variant[Enum['present', 'absent', 'purged', 'disabled', 'installed', 'latest'], String[1]]


types/ensure/package.pp
```diff
@@ -0,0 +1,2 @@
+# @summary Validate the value of the ensure parameter for a package
+type Stdlib::Ensure::Package = Variant[Enum['present', 'absent', 'purged', 'disabled', 'installed', 'latest'], Pattern[/\./]]
```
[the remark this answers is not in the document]
arjenz: When I was working on this I was also thinking "this would not allow non decimal numbers or bare git commit hash as version", but the Puppet docs don't allow it, so I'm also not allowing it here :)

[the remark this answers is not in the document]
arjenz: Or, am I misreading https://puppet.com/docs/puppet/7/types/package.html#package-attribute-ensure, and is that actually a regex wildcard . in there, `/./`, in that case, my version is too strict (and `String[1]` would be better).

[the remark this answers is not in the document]
kenyon: That's indeed a regex equivalent to `String[1]`.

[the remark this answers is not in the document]
arjenz: I changed it to `String[1]`

GSPatton (changes requested):
this looks good. Could you add a unit test for this change? Thanks

jordanbreen28:
Hi @arjenz,
Did you get the chance to implement @GSPatton's requested changes above?
Thanks

arjenz:
@jordanbreen28 Not yet, I wanted to copy the tests for the similar types in there, but there weren't any yet, I will try and get to that this week and push accordingly, thanks!

jordanbreen28:
@arjenz Thanks! Appreciate your work on this one.

types/ensure/package.pp
```diff
@@ -0,0 +1,2 @@
+# @summary Validate the value of the ensure parameter for a package
+type Stdlib::Ensure::Package = Variant[Enum['present', 'absent', 'purged', 'disabled', 'installed', 'latest'], String[1]]
```
b4ldr, line 2: This may as well be:
```suggestion
type Stdlib::Ensure::Package = String[1]
```
The Enum is completely superfluous and im not sure there is a whole lot of use in having a type definitions which is just an alias to String[1].  no objections either though

david22swan:
@arjenz Hey, just checking in on whether you've made any progress on this?

types/ensure/package.pp
LukasAud: @arjenz @b4ldr The Enum is, indeed, pretty much an alias for String[1]. However, we prefer to have it defined that way for a couple of reasons. One is to make the parameter easy to understand, by displaying valid values right away, and the other is to keep the documentation (see REFERENCE.md) as accurate as possible regarding $ensure values.

ekohl (approved):
I can see the benefits of explicit for documentation purposes. You could write a small test (https://rspec-puppet.com/documentation/type_aliases/) to make sure it's valid, but :+1: to this.

types/ensure/package.pp
b4ldr: ack no  blocker from me

GSPatton:
as @ekohl suggested, could you add a couple of test cases to ensure this new type works consistently in the future? After which we would be happy to merge. Thanks

arjenz:
Having some issues on M1 laptop getting the tests to run (and unfortunately the other types don't have similar tests yet), I'll try and create a test.

ekohl:
> unfortunately the other types don't have similar tests yet

There are a bunch of them: https://github.com/puppetlabs/puppetlabs-stdlib/tree/main/spec/type_aliases

arjenz:
> I can see the benefits of explicit for documentation purposes. You could write a small test (https://rspec-puppet.com/documentation/type_aliases/) to make sure it's valid, but 👍 to this.

@ekohl 
Done!

spec/type_aliases/ensure_package_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+require 'spec_helper'
+
+describe 'Stdlib::Ensure::Package' do
+  describe 'valid handling' do
+    [
+      'present',
+      'absent',
+      'purged',
+      'disabled',
+      'installed',
```
jcpunk, line 12: `latest`?

arjenz: Fell off when I copy/pasted from https://www.puppet.com/docs/puppet/7/types/package.html#package-attribute-ensure, re-added, thanks!

arjenz:
I don't think any of the failing checks are related to my change

spec/type_aliases/ensure_package_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+      'latest',
+      '1',
+      '1.1',
```
ekohl, line 15: Could you also add a version range, which is supported by at least some package providers (at least DNF):
```suggestion
      '>=6.0',
```

It'll be more for documentation purposes/regression testing than anything else.

ekohl (approved):
Thanks!

LukasAud (approved):
LGTM

LukasAud:
Thanks for your contribution!

PR #1286, opened 2022-12-09
stdlib::ensure: update function to support the generic case

b4ldr:
Often custom resources only support present/absent for the ensure parameter and often the inclusion of theses resources are enabled via some boolean value as such i often find my self using the following pattern

```puppet
$ensure_feature = $enable_feature.bool2str('ensure', 'present')
```

This patch updates the stdlib::ensure function so we can simplify this to

```puppet
$ensure_feature = $enable_feature.stdlib::ensure   # or ...
$ensure_feature = stdlib::ensure($enable_feature) 
```

### DIFF
--- a/functions/ensure.pp
+++ b/functions/ensure.pp
@@ -3,7 +3,7 @@
 # @return [String]
 function stdlib::ensure(
   Variant[Boolean, Enum['present', 'absent']] $ensure,
-  Enum['directory', 'link', 'mounted', 'service', 'file', 'package'] $resource,
+  Optional[Enum['directory', 'link', 'mounted', 'service', 'file', 'package']] $resource = undef,
 ) >> String {
   $_ensure = $ensure ? {
     Boolean => $ensure.bool2str('present', 'absent'),
@@ -22,6 +22,7 @@ function stdlib::ensure(
         default   => 'stopped',
       }
     }
+    undef: { $_ensure }
     default: {
       $_ensure ? {
         'present' => $resource,
--- a/spec/functions/stdlib_ensure_spec.rb
+++ b/spec/functions/stdlib_ensure_spec.rb
@@ -3,6 +3,10 @@
 require 'spec_helper'
 
 describe 'stdlib::ensure' do
+  context 'work without resource' do
+    it { is_expected.to run.with_params(true).and_return('present') }
+    it { is_expected.to run.with_params(false).and_return('absent') }
+  end
   context 'work with service resource' do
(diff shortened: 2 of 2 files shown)


spec/functions/stdlib_ensure_spec.rb
```diff
@@ -3,6 +3,12 @@
 require 'spec_helper'
 
 describe 'stdlib::ensure' do
+  context 'work with generi resource' do
```
ekohl, line 6:
```suggestion
  context 'work with generic resource' do
```

functions/ensure.pp
```diff
@@ -3,7 +3,7 @@
 # @return [String]
 function stdlib::ensure(
   Variant[Boolean, Enum['present', 'absent']] $ensure,
-  Enum['directory', 'link', 'mounted', 'service', 'file', 'package'] $resource,
+  Enum['present', 'directory', 'link', 'mounted', 'service', 'file', 'package'] $resource = 'present',
```
ekohl, line 6: I don't think there is a resource type named `present`. If you want to make it optional, then perhaps it should be made `undef` and caught as an `undef` in the case statement.
```suggestion
  Optional[Enum['present', 'directory', 'link', 'mounted', 'service', 'file', 'package']] $resource = undef,
```

ekohl (changes requested):

spec/functions/stdlib_ensure_spec.rb
```diff
@@ -3,6 +3,10 @@
 require 'spec_helper'
 
 describe 'stdlib::ensure' do
+  context 'work with generic resource' do
```
ekohl, line 6:
```suggestion
  context 'work without resource' do
```

ekohl (approved):
Would you mind squashing the commits?

I'd also appreciate if someone else took a look since I suggested the implementation.

b4ldr:
> Would you mind squashing the commits?

done

> I'd also appreciate if someone else took a look since I suggested the implementation.

sure no rush my end

kenyon (changes requested):
Would be good to fix the typo in the PR title and commit message, `spport` :)

b4ldr:
> Would be good to fix the typo in the PR title and commit message, `spport` :)

done

kenyon (approved):

ekohl (approved):
In this repository Puppet's module team needs to approve before it can be merged.

LukasAud (approved):
LGTM

PR #1288, opened 2022-12-28
Add `stdlib::crc32`

jcpunk:
CRC32 is a handy way to run trivial checks against values.  Unlike the SHA hashing series generating collisions is fairly trivial which makes it useful for describing secret data in a public manner.  For example, "Is you app using secret `18fbd270` or `7d5469f0`" cannot possibly be used to reconstruct the secret but does permit concrete conversation about the secret itself.  This can be further useful when the developers do not have direct access to the secret in question.

It is also handy for establishing an application specific shared seed across multiple hosts, for establishing trivial user sharding, or other applications where uniqueness is not actually an asset.

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -173,6 +173,7 @@ the provided regular expression.
 * [`sort`](#sort): Sorts strings and arrays lexically.
 * [`sprintf_hash`](#sprintf_hash): Uses sprintf with named references.
 * [`squeeze`](#squeeze): Returns a new string where runs of the same character that occur in this set are replaced by a single character.
+* [`stdlib::crc32`](#stdlibcrc32): Run a CRC32 calculation against a given value.
 * [`stdlib::deferrable_epp`](#stdlibdeferrable_epp): This function returns either a rendered template or a deferred function to render at runtime. If any of the values in the variables hash are 
 * [`stdlib::end_with`](#stdlibend_with): Returns true if str ends with one of the prefixes given. Each of the prefixes should be a String.
 * [`stdlib::ensure`](#stdlibensure): function to cast ensure parameter to resource specific value
@@ -4190,7 +4191,9 @@ hash types are:
 |bcrypt-x |2x    |bug compatible       |
 |bcrypt-y |2y    |historic alias for 2b|
 
-The third argument to this function is the salt to use.
+The third argument to this function is the salt to use. For bcrypt-type hashes,
+the first two characters of the salt represent a strength parameter, with a value
+between 4 and 31 inclusive.
 
 > *Note:*: this uses the Puppet Server's implementation of crypt(3). If your
   environment contains several different operating systems, ensure that they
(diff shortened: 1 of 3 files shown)


spec/functions/crc32_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  context 'when testing an integer' do
+    it { is_expected.to run.with_params(0).and_return('f4dbdf21') }
+    it { is_expected.to run.with_params(100).and_return('237750ea') }
+    it { is_expected.to run.with_params(sensitive(100)).and_return('237750ea') }
+  end
+
+  context 'when testing a float' do
+    it { is_expected.to run.with_params(200.300).and_return('7d5469f0') }
```
ekohl, line 27: `200.300` and `200.3` describe the same source so I think Rubocop is correct here.

lib/puppet/functions/stdlib/crc32.rb
```diff
@@ -0,0 +1,26 @@
...
+# @note
+#   The CRC32 algorithm can easily generate collisions,
+#   but may be useful for generating sharding, describing
+#   secrets, or seeding nonce values.
+#
+# @summary
+#   Run a CRC32 calculation against a given value.
+Puppet::Functions.create_function(:'stdlib::crc32') do
+  # @param my_data The ScalarData to evaluate
+  # @example
```
ekohl, line 13: It's best to name your examples. You can see there's an empty header in the reference. Ideally speaking something would at least raise a warning, possibly an error.

spec/functions/crc32_spec.rb
jcpunk: I was intending to test that first the float rules of drop the trailing `00` are applied before the resource is converted to a string, but that may not be worth testing?

lib/puppet/functions/stdlib/crc32.rb
jcpunk: Added and squashed.

spec/functions/crc32_spec.rb
ekohl: I think it's changed during parsing:
```console
$ pry
[1] pry(main)> 200.300
=> 200.3
```

REFERENCE.md
```diff
@@ -4662,6 +4667,40 @@ The squeeze function.
...
+
+#### `stdlib::crc32(Variant[ScalarData, Sensitive[ScalarData]] $my_data)`
+
+Run a CRC32 calculation against a given value.
+
+Returns: `String` String
+
+##### Examples
+
+######
```
ekohl, line 4692: I think you need to regenerate the reference

jcpunk: :) should be done now.

lib/puppet/functions/stdlib/crc32.rb
```diff
@@ -0,0 +1,31 @@
...
+#   The CRC32 algorithm can easily generate collisions,
+#   but may be useful for generating sharding, describing
+#   secrets, or seeding nonce values.
+#
+# @summary
+#   Run a CRC32 calculation against a given value.
+Puppet::Functions.create_function(:'stdlib::crc32') do
+  # @param my_data The ScalarData to evaluate
+  # @example Check a simple string value
+  #   stdlib::crc32('my string') => 18fbd270
```
ekohl, line 14: I'd keep it valid Puppet syntax
```suggestion
  #   stdlib::crc32('my string') == '18fbd270'
```

ekohl (approved):

bastelfreak (approved):

chelnak (approved):

PR #1294, opened 2023-02-07
(PUP-11752) fqdn_rand_string_spec.rb relies on legacy fqdn fact

AriaXLi:


spec/functions/fqdn_rand_string_spec.rb
```diff
@@ -57,7 +57,7 @@ def fqdn_rand_string(max, args = {})
 
     # workaround not being able to use let(:facts) because some tests need
     # multiple different hostnames in one context
-    allow(scope).to receive(:lookupvar).with('::fqdn', {}).and_return(host)
+    allow(scope).to receive(:lookupvar).with('facts', {}).and_return( { 'networking' => { 'fqdn' => host } })
```
smortex, line 60:
```suggestion
    allow(scope).to receive(:lookupvar).with('facts', {}).and_return({ 'networking' => { 'fqdn' => host } })
```

smortex (commented):
Good catch, this will be needed with the next version of Puppet.

Minor style issue, also the commit message needs need adjusting :smile:

spec/functions/fqdn_rand_string_spec.rb
AriaXLi: Yeah my git is being difficult rn 😭  but I'll fix that once I get that figured out, thank you!

spec/functions/fqdn_rand_string_spec.rb
```diff
@@ -57,7 +57,7 @@ def fqdn_rand_string(max, args = {})
 
     # workaround not being able to use let(:facts) because some tests need
     # multiple different hostnames in one context
-    allow(scope).to receive(:lookupvar).with('::fqdn', {}).and_return(host)
+    allow(scope).to receive(:lookupvar).with('facts', {}).and_return({ 'networking' => { 'fqdn' => host } })
```
joshcooper, line 60: Ah the expectation needs to be different based on the version of puppet (before and after this change https://github.com/puppetlabs/puppet/commit/c0d25605ba32aca64149459d561b2880f5c99deb). So something like

```ruby
if Puppet::PUPPETVERSION.to_f < 7.23
  allow(scope).to receive(:lookupvar).with('::fqdn', {}).and_return(host)
else
  allow(scope).to receive(:lookupvar).with('facts', {}).and_return({ 'networking' => { 'fqdn' => host } })
end
```

Or there may be a better way to stub out the facts using `RspecPuppetFacts`?

ekohl: Perhaps this function should just mock `fqdn_rand`? It really is `fqdn_rand` which gets the fact, so we don't care about the internal implementation:
https://github.com/puppetlabs/puppetlabs-stdlib/blob/dfd9b30f1a2c04278b11088bae738da0a58d6aae/lib/puppet/parser/functions/fqdn_rand_string.rb#L37

PR #1298, opened 2023-03-10
Create Stdlib::File::* types

ekohl:
This creats a Stdlib::File namespace for file attributes. It introduces Stdlib::File::Content, but I'm not sure if this is complete. https://www.puppet.com/docs/puppet/7/type.html#file-attribute-content doesn't describe valid or invalid input. Please discuss what should be accepted.


types/file/content.pp
```diff
@@ -0,0 +1,2 @@
+# @summary Validate a file content attribute
+type Stdlib::File::Content = Variant[String, Sensitive[String], Deferred[String]]
```
b4ldr, line 2: TIL: this can also be [Binary](https://github.com/puppetlabs/puppet/blob/main/lib/puppet/type/file/content.rb#L66)

spec/type_aliases/file_source_spec.rb
```diff
@@ -0,0 +1,64 @@
+# frozen_string_literal: true
```
b4ldr, line 1: we should also remove spec/type_aliases/filesource_spec.rb and spec/type_aliases/filemode_spec.rb

ekohl: Since they are still aliases, I kept them around. Kind of not sure what to do with it though.

bastelfreak (approved):

spec/type_aliases/file_content_spec.rb
```diff
@@ -0,0 +1,30 @@
+# coding: utf-8
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe 'Stdlib::Filemode' do
```
b4ldr, line 6:
```suggestion
describe 'Stdlib::File::Content' do
```

PR #1299, opened 2023-03-10
Stdlib::Http::Method: Add new type for http methods

### LINKED ISSUE #1297: missing merge 1192
i just noticed that #1192 seems to be missing.  I don't see anything removing it and also don't see the commit in the [history](https://github.com/puppetlabs/puppetlabs-stdlib/commits/main/types).  Wondering if I'm missing something, if i need to submit again or if it was removed because of some other issues thanks

b4ldr:
This PR creates new new resources:
 * Stdlib::Http::Method for validating http methods
 * Stdlib::Http::Status This is just a copy of Stdlib::Httpstatus
 * make Stdlib::Httpstatus and alias to Stdlib::Http::Status

This PR re-intorduces #1192 which i apparently nuked in #1193

Fixes #1297


types/httpstatus.pp
```diff
@@ -1,2 +1,2 @@
 # @summary Validate a HTTP status code
```
ekohl, line 1: I think you can deprecate this type. The `@see` may not be needed if `@deprecated` creates a link, but I don't think it does.
```suggestion
# @summary Validate a HTTP status code
# @deprecated Use Stdlib::Http::Status
# @see Stdlib::Http::Status
```

Please check by regenerating `REFERENCE.md`.

types/http/status.pp
```diff
@@ -0,0 +1 @@
+type Stdlib::Http::Status = Integer[100, 599]
```
ekohl, line 1: # @summary A valid HTTP status code per RFC9110
# @see https://httpwg.org/specs/rfc9110.html#overview.of.status.codes
```suggestion
type Stdlib::Http::Status = Integer[100, 599]
```

types/http/method.pp
```diff
@@ -0,0 +1,42 @@
+# https://www.iana.org/assignments/http-methods/http-methods.xhtml
```
ekohl, line 1:
```suggestion
# @summary Valid HTTP method verbs
# @see https://www.iana.org/assignments/http-methods/http-methods.xhtml
```

types/httpstatus.pp
b4ldr: i have added both but i don't think deprecated dose anything yet https://github.com/puppetlabs/puppet-strings/issues/339

b4ldr:
i'm guessing the [ci issue](https://github.com/puppetlabs/puppetlabs-stdlib/actions/runs/4385438700/jobs/7678160829) is unrelated?

REFERENCE.md
```diff
@@ -7893,14 +7906,43 @@ Alias of
...
+* **See also**
+  * https://www.iana.org/assignments/http-methods/http-methods.xhtml
+
+Alias of
+
+```puppet
+Enum['ACL', 'BASELINE-CONTROL', 'BIND', 'CHECKIN', 'CHECKOUT', 'CONNECT', 'COPY', 'DELETE', 'GET', 'HEAD', 'LABEL', 'LINK', 'LOCK', 'MERGE', 'MKACTIVITY', 'MKCALENDAR', 'MKCOL', 'MKREDIRECTREF', 'MKWORKSPACE', 'MOVE', 'OPTIONS', 'ORDERPATCH', 'PATCH', 'POST', 'PRI', 'PROPFIND', 'PROPPATCH', 'PUT', 'REBIND', 'REPORT', 'SEARCH', 'TRACE', 'UNBIND', 'UNCHECKOUT', 'UNLINK', 'UNLOCK', 'UPDATE', 'UPDATEREDIRECTREF', 'VERSION-CONTROL']
+```
+
+### <a name="stdlibhttpstatus"></a>`Stdlib::Http::Status`
```
ekohl, line 7922: This is "funny": strings generates the same anchor for `Stdlib::Http::Status` as `Stdlib::HttpStatus`.

ekohl:
You can ignore mend, but I wonder why the unit test is failing.

REFERENCE.md
b4ldr: hmm not sure we can do much about it here, i have create an [issue](https://github.com/puppetlabs/puppet-strings/issues/341) in puppet strings

b4ldr:
thanks, the failed spec test relate to `fqdn_rand_string` so i don't think anything in this PR could have broken them. i have kicked the testes of again to see if it was just something transient

ekohl:
Oh no, that's https://github.com/puppetlabs/puppetlabs-stdlib/pull/1294.

REFERENCE.md
ekohl: I don't think there's much we can do here other than what you did. I think the good news is that it generates the one we do want first so it's likely browsers will use that anchor.

I wonder why they do this since they could use `-` (which is invalid in a Puppet type, so should work). According to https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id that should be fine.

TIL: the `name` attribute is actually deprecated according to https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a and they recommend `id` instead.

b4ldr: look is like this has already been [reported](https://github.com/puppetlabs/puppet-strings/issues/300) and [fixed](https://github.com/puppetlabs/puppet-strings/pull/303).  seems that bolt is locking puppet-strings to ~> 2.3

b4ldr: ok so i see the fix to bolt is included in 3.27 which was released 17hours ago, however its not yet on rubygems.  As such i think we can probably wait a day or two for bolt to get released then `bundle update` and regenerate

bastelfreak (approved):

types/httpstatus.pp
```diff
@@ -1,2 +1,4 @@
...
-type Stdlib::HttpStatus = Integer[100, 599]
+# @deprecated Use Stdlib::Http::Status
+# @see Stdlib::Http::Status
+type Stdlib::HttpStatus = Stdlib::Http::Status
```
bastelfreak, line 4: after this is merged, can you raise a new PR that removes this old type? Then we can keep it open until the next major release and won't forget it.

REFERENCE.md
b4ldr: i have regenerated references.md in #1301 once that's merged ill rebase

types/httpstatus.pp
b4ldr: sure will do, ill leave this open for now so i dont forget

REFERENCE.md
b4ldr: ok updated

ekohl (approved):
Thanks!

PR #1301, opened 2023-03-14
REFERENCE.md: apply fix for unique anchors from puppet strings

b4ldr:
Regenerate references with the following fix
 https://github.com/puppetlabs/puppet-strings/pull/303


alexjfisher (approved):
+1 for this being its own PR and keeping the noise out of all the other PRs with code changes.

PR #1307, opened 2023-04-21
(CONT-801) Puppet 8 support / Drop Puppet 6 support

LukasAud:
This PR aims to implement support for Puppet 8. Some of the changes performed might result in loss of compatibility for Puppet 6, which we no longer support. Current changes implemented in this PR:

- Pinned agent to latest version (This brings Puppet 8 into the formula, alongside Ruby 3.2)
- Updated version requirements in metadata.json
- Bumped Rubocop gems
- Adjusted Rubocop configuration
- Resolved new Rubocop offenses
- Deprecated uriescape.rb and parsepson.rb

Issues addressed:

- .scope is no longer a PuppetInternals functions. Now there is a new version living in RSpec. Addressed related testing.
- srand() returns a different sort after Ruby 3 while inputting the same seed.
- uriescape.rb no longer works past Ruby 3. Limiting testing to Puppet 7.
- parsepson.rb is no longer supported as of Puppet 8 due to removal of native PSON support. Limiting testing to Puppet 7.
- Found and addressed an issue related to positional vs keyword arguments in loadjson.rb

Note: Various issues found during testing were addressed by external PRs. Check PR history nearing this Request by backwards incompatible labels.


alexjfisher:
Any chance we can we get a final 8.x release in with https://github.com/puppetlabs/puppetlabs-stdlib/pull/1300 included first please?
cc @binford2k

LukasAud:
Hey @alexjfisher. Sure, we always aim to get a pre-release sorted before releasing any major update. Will take a deeper look into that PR once this update is finished and before we start the releasing new versions. It might, however, be hard to merge given the current state of the unit tests (got a few of them broken, trying to troubleshoot at the moment). Will let you know if we need a rebase on the other PRs end with our fix.

alexjfisher:
@LukasAud https://github.com/puppetlabs/puppetlabs-stdlib/pull/1308 should help with some of the current unit test failures.

.rubocop_todo.yml
```diff
@@ -1,26 +1,718 @@
...
+# AllowedIdentifiers: capture3, iso8601, rfc1123_date, rfc822, rfc2822, rfc3339, x86_64
+Naming/VariableNumber:
+  Exclude:
+    - 'spec/functions/delete_undef_values_spec.rb'
+
+# Offense count: 1
+# This cop supports unsafe autocorrection (--autocorrect-all).
+Performance/MapCompact:
+  Exclude:
+    - 'lib/puppet/parser/functions/values_at.rb'
```
alexjfisher, line 212: Fixing this one would be more than just 'unsafe'.  This is a ruby 2.7 specific example and Puppet Server 7 is only ruby 2.6 compatible.

LukasAud: Yes, there is certain rubocop rules that we have been avoiding during our updates. Mainly those that break the behaviour of the module right away. We will be amending code as soon as possible.

alexjfisher:
Is this also a good opportunity to dump some of the deprecated functions? (Some of these could probably go now even _without_ a major version bump)

.rubocop_todo.yml
ekohl: That would be interesting, because puppetlabs-concat already uses Ruby 2.7 features: https://github.com/puppetlabs/puppetlabs-concat/blob/7d9fd72f22165b41328fdb29a6a15ea56d275e75/lib/puppet/type/concat_file.rb#L193 uses filter_map.

I'm really wondering about the Ruby 2.6 part. If that's true, then it's going to be very messy.

alexjfisher: Maybe that bit only ever gets executed on the agent, and in this case, it's still loadable by ruby 2.6 even if execution would fail??

LukasAud: Its a very tricky situation as this revelation has come when over 80% of the modules had already undergone the TargetRubyVersion 2.7. Nevertheless, probably only the very first few will modules to be worked on will present problems as we started to reduce the workload each module represented by moving onto a "only deal with autocorrections, ignore problematic cop rules" policy when we realised how limited on time we were.

LukasAud:
@alexjfisher Yes, already mentioned but we will take a look around and invite everyone else too to help us find and remove any deprecated functions along with their test cases. A release has been cut very recently so these PRs should be mergeable right away.

LukasAud:
Note: Regarding #1354, Stdlib seems to contain its own version of `.exists?` defined [here](https://github.com/puppetlabs/puppetlabs-stdlib/blob/main/lib/puppet/provider/file_line/ruby.rb#L13). This seems to be the logic the module resolves to when the method is called. Furthermore, attempting to replace `.exists?` instances with `.exist?` breaks the code logic

spec/functions/grep_spec.rb
```diff
@@ -3,17 +3,20 @@
...
   it {
     pending('grep does not actually check this, and raises NoMethodError instead')
-    is_expected.to run.with_params('one', 'two').and_raise_error(Puppet::ParseError, %r{first argument not an array})
+    expect(subject).to run.with_params('one', 'two').and_raise_error(Puppet::ParseError, %r{first argument not an array})
   }
+
   it {
     pending('grep does not actually check this, and raises NoMethodError instead')
-    is_expected.to run.with_params(1, 'two').and_raise_error(Puppet::ParseError, %r{first argument not an array})
+    expect(subject).to run.with_params(1, 'two').and_raise_error(Puppet::ParseError, %r{first argument not an array})
```
jordanbreen28, line 17: Same as above

lib/puppet/parser/functions/load_module_metadata.rb
```diff
@@ -17,14 +17,19 @@ module Puppet::Parser::Functions
...
+
     mod = args[0]
     allow_empty_metadata = args[1]
     module_path = function_get_module_path([mod])
     metadata_json = File.join(module_path, 'metadata.json')
 
     metadata_exists = File.exists?(metadata_json) # rubocop:disable Lint/DeprecatedClassMethods : Changing to .exist? breaks the code
     if metadata_exists
-      metadata = PSON.load(File.read(metadata_json))
+      metadata = if Puppet::Util::Package.versioncmp(Puppet.version, '8.0.0').negative?
```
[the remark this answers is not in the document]
LukasAud: Rubocop prefers the usage of `.negative?` instead of `< 0`

Gemfile
```diff
@@ -14,31 +14,31 @@ def location_for(place_or_version, fake_version = nil)
...
+  gem "rubocop", '~> 1.48.1',                      require: false
+  gem "rubocop-performance", '~> 1.16',            require: false
+  gem "rubocop-rspec", '~> 2.19',                  require: false
+  gem "rb-readline", '= 0.5.5',                    require: false, platforms: [:mswin, :mingw, :x64_mingw]
+  gem "github_changelog_generator", '= 1.15.2',    require: false
 end
 group :system_tests do
-  gem "puppet_litmus", '~> 1.0', require: false, platforms: [:ruby, :x64_mingw]
-  gem "serverspec", '~> 2.41',      require: false
+  gem "puppet_litmus", '< 1.0.0', require: false, platforms: [:ruby, :x64_mingw]
```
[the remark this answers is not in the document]
LukasAud: Sure

spec/functions/grep_spec.rb
```diff
@@ -3,17 +3,20 @@
...
 describe 'grep' do
-  it { is_expected.not_to eq(nil) }
+  it { is_expected.not_to be_nil }
   it { is_expected.to run.with_params.and_raise_error(Puppet::ParseError, %r{wrong number of arguments}i) }
   it { is_expected.to run.with_params('one').and_raise_error(Puppet::ParseError, %r{wrong number of arguments}i) }
+
   it {
     pending('grep does not actually check this, and raises NoMethodError instead')
-    is_expected.to run.with_params('one', 'two').and_raise_error(Puppet::ParseError, %r{first argument not an array})
+    expect(subject).to run.with_params('one', 'two').and_raise_error(Puppet::ParseError, %r{first argument not an array})
```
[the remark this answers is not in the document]
LukasAud: This is a pending exception, you can see the message at Line 11.

spec/functions/grep_spec.rb
LukasAud: Same as above.

spec/functions/ensure_resource_spec.rb
```diff
@@ -14,7 +15,7 @@
 
   it {
     pending('should not accept numbers as arguments')
-    is_expected.to run.with_params(1, 2, 3).and_raise_error(Puppet::ParseError)
+    expect(subject).to run.with_params(1, 2, 3).and_raise_error(Puppet::ParseError)
```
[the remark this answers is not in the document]
LukasAud: As stated in the pending message, this function should not be accepting numbers and, upon trying, should return a ParseError. However, it looks like this logic has not been implemented yet (or perhaps removed for some reason at some point) so this is another pending exception that the module carries over from previous versions.

spec/functions/loadjson_spec.rb
```diff
@@ -49,8 +54,13 @@
         allow(File).to receive(:exists?).with(filename).and_return(true).once
         allow(File).to receive(:read).with(filename).and_return(json).once
         allow(File).to receive(:read).with(filename).and_return(json).once
-        allow(PSON).to receive(:load).with(json).and_return(data).once
+        if Puppet::PUPPETVERSION[0].to_i < 8
```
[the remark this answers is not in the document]
jordanbreen28: can `Puppet::PUPPETVERSION[0].to_i < 8` not be used instead of `Puppet::Util::Package.versioncmp(Puppet.version, '8.0.0').negative?`

lib/puppet/parser/functions/load_module_metadata.rb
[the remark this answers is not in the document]
LukasAud: I believe the way versioncmp works is by returning `-1` if the value is lower, `0` is equal, and `1` if higher. Hence, the `.negative?` / `.positive?` methods. So on Puppet 8.0.0, it will return a `0`. I am unsure, however, if it returns a `0` or a `1` on Puppet 8.0.1+.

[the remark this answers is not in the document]
jordanbreen28: Cool -  well we are running puppet 8.0.1 at the moment ([here](https://github.com/puppetlabs/puppetlabs-stdlib/actions/runs/5059183004/jobs/9080360953?pr=1307#step:4:328)) and looks to be all good so 👍 . I'll resolve this.

david22swan (approved):
LGTM

jordanbreen28 (approved):

lib/puppet/functions/to_json_pretty.rb
```diff
@@ -61,22 +61,14 @@ def to_json_pretty(data, skip_undef = false, opts = nil)
...
-        data = data.reject { |value| value.nil? }
-      elsif data.is_a? Hash
-        data = data.reject { |_, value| value.nil? }
-      end
-    end
+    data = data.compact if skip_undef && (data.is_a?(Array) || Hash)
     # Call ::JSON to ensure it references the JSON library from Ruby's standard library
     # instead of a random JSON namespace that might be in scope due to user code.
-    ::JSON.pretty_generate(data, opts) << "\n"
+    JSON.pretty_generate(data, opts) << "\n"
```
alexjfisher, line 72: I wonder why these leading colons were removed? It doesn't seem to be a rubocop violation, and the comment above said they were there for a reason.

LukasAud: Looks like it got 'safely' autocorrected by Rubocop during the Puppet 8 update. This might have been one of those cases where we should have added an inline exception instead of allowing it, but it probably sneaked past us.

I'll raise a PR to revert this and let the Modules team know about it.

LukasAud: Just for context, the Puppet 8 update period was a fairly chaotic point where we were handling a lot of module changes simultaneously. At the same time, work leading to Puppet 8 also required us to clean up previous Lint exceptions living in the code for most modules, and one of those exceptions had to do with the usage of top-scope variable calling, so it was not weird for us to see the :: symbol being removed left and right.

LukasAud: oh nvm, looks like the entire function was deprecated some time ago

alexjfisher: Only the non-namespaced version is deprecated.  But don't worry about a PR right now.  I'm about to throw in a PoC/RFC style PR that touches this function.  Would be interested to hear your thoughts.

PR #1308, opened 2023-04-21
(PUP-11752) Fix fqdn_rand_string_spec.rb test

alexjfisher:
Since Puppet `7.23.0` `fqdn_rand` uses the modern structured `networking` fact instead of the legacy `fqdn` fact.

In this commit we mock the new fact if using `7.23.0` or later.

Rebase of https://github.com/puppetlabs/puppetlabs-stdlib/pull/1294 incorporating @joshcooper 's suggestion.

### DIFF
--- a/spec/functions/fqdn_rand_string_spec.rb
+++ b/spec/functions/fqdn_rand_string_spec.rb
@@ -57,7 +57,11 @@ describe 'fqdn_rand_string' do
 
     # workaround not being able to use let(:facts) because some tests need
     # multiple different hostnames in one context
-    allow(scope).to receive(:lookupvar).with('::fqdn', {}).and_return(host)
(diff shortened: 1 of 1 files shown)


bastelfreak (dismissed):

ekohl (commented):
Would it make sense to just mock `fqdn_rand`, rather than the implementation detail of how it's determined?

alexjfisher:
> Would it make sense to just mock `fqdn_rand`, rather than the implementation detail of how it's determined?

If you want, but this issue has been open for a couple of months with _no_ fix being merged.

https://en.wikipedia.org/wiki/Perfect_is_the_enemy_of_good

Same argument could be made for my `ensure_packages` fix.  That's been sat there for a long time now.  IMO, nothing stops either of these fixes being merged and then someone circling back later to replace with a better fix.

bastelfreak (approved):

PR #1310, opened 2023-04-24
Remove deprecated Private function

MartyEwings:
Private ceased to work in version of puppet greater than 3, and should be removed


alexjfisher (approved):
I can't see how this could possibly be relevant anymore so LGTM.

LukasAud (approved):
Happy enough to merge. Thanks for your contribution!

PR #1312, opened 2023-04-24
(MAINT) Adjust .sync.yml config

LukasAud:
Prior to this commit, changelog generator would not work as expected during our actions auto-release. Issue was deducted to be a change in tags at some point in the past.

This commit aims to establish v5.0.0 as the starting point of changelog generation to avoid crashes in our actions.


alexjfisher:
This should be automatic based on the header in HISTORY.md??

PR #1314, opened 2023-04-24
Remove Puppet 5.5 deprecations

MartyEwings:
as noted in this PR https://github.com/puppetlabs/puppetlabs-stdlib/pull/896

as of Puppet 5.5 these functions have been available in core puppet and should now be removed


LukasAud (approved):
Yep, these also look like they can be removed right away. LGTM

david22swan (approved):

PR #1316, opened 2023-04-25
Remove dig and dig44 functions

MartyEwings:
dig became part of puppet core in Puppet 4, and even the compatibility function dig44 does not work in puppet 6 or later

this should be removed.

As part of this PR try_get_value was also removed as it was deprecated in favour of dig 6 years ago


MartyEwings:
Looks like some of the unit tests for Other functions used Dig? ill have to figure out how to unhook all of that

bastelfreak (approved):
I'm surprised that the REFERENCE.md wasn't updated. The CI probably doesn't check for it (we recently implemented. I don't think you need to regenerate it for each PR, but during the release PR we need to ensure it's up2date (puppetlabs_spec_helper has a rake task for that I think). CC: @chelnak 

I think vox pupuli calls this: release_checks -> validate -> https://github.com/puppetlabs/puppetlabs_spec_helper/blob/main/lib/puppetlabs_spec_helper/rake_tasks.rb#L213

MartyEwings:
> I'm surprised that the REFERENCE.md wasn't updated. The CI probably doesn't check for it (we recently implemented. I don't think you need to regenerate it for each PR, but during the release PR we need to ensure it's up2date (puppetlabs_spec_helper has a rake task for that I think). CC: @chelnak
> 
> I think vox pupuli calls this: release_checks -> validate -> https://github.com/puppetlabs/puppetlabs_spec_helper/blob/main/lib/puppetlabs_spec_helper/rake_tasks.rb#L213

I think if this is the same as my workflows on my team, REFERENCE.MD is updated as part of the release prep workflow

chelnak:
Yeah I believe that is correct

PR #1318, opened 2023-04-25
Remove deprecated downcase function

MartyEwings:
*Note:* **Deprecated** from Puppet 6.0.0, this function has been replaced with a
    built-in [`downcase`](https://puppet.com/docs/puppet/latest/function.html#downcase) function.


LukasAud (approved):
LGTM

smortex (approved):
There also is an `upcase` function we can remove for the same reason.

* #1334

alexjfisher:
Do you think this (and lots of the similar PRs) actually needs be be marked as breaking?  If the function has been built-in since Puppet 6 and we dropped 5 a long time ago??  Unlike say, `has_key` that is deprecated because there's a puppet native way of achieving the same.

MartyEwings:
> 

Im not sure, i can go back and get them changed, this was more so the resulting version change would be a Major version, but i guess dropping puppet 6 will also do that...

PR #1319, opened 2023-04-25
Remove deprecated has_key function

MartyEwings:
**Note:** **Deprecated** since Puppet 4.0.0, this can now be achieved in the Puppet
    language with the following equivalent expression:

```
    $my_hash = {'key_one' => 'value_one'}
    if 'key_one' in $my_hash {
      notice('this will be printed')
    }
```


LukasAud (approved):
LGTM but what is the equivalent expression here?

PR #1324, opened 2023-04-25
Remove deprecated is_absolute_path function

MartyEwings:
deprecated 4 years ago

use match expressions with Stdlib::Compat::Absolute_Path instead. They are described at https://docs.puppet.com/puppet/latest/reference/lang_data_type.html#match-expressions.


smortex:
Same as I say in another PR, while I agree with the removal, I do think we could go further and drop the compat data types too. These are basically the standard data types with bug-to-bug compatibility with the old functions if I recall correctly.

bastelfreak (approved):

PR #1325, opened 2023-04-25
Remove deprecated is_array function

MartyEwings:
use match expressions with Stdlib::Compat::Array instead. They are described at https://docs.puppet.com/puppet/latest/reference/lang_data_type.html#match-expressions.


smortex (commented):
While I agree with the removal, I do think we could go further and drop the compat data types too.  These are basically the standard data types with bug-to-bug compatibility with the old functions if I recall correctly.

bastelfreak (approved):

bastelfreak:
@smortex I agree to 100%

davidcomeyne:
Match expressions on the docs.puppet.com has been moved to https://www.puppet.com/docs/puppet/8/lang_data_type.html#match-expressions

PR #1326, opened 2023-04-25
Modernise `has_interface_with` function

alexjfisher:
Convert the function to the modern function API as a namespaced function and use the `networking` fact instead of legacy facts.

A non-namespaced shim is also created (but marked deprecated), to preserve compatibility.

### DIFF
--- a/lib/puppet/functions/stdlib/has_interface_with.rb
+++ b/lib/puppet/functions/stdlib/has_interface_with.rb
@@ -1,0 +1,47 @@
+# frozen_string_literal: true
+
+# @summary Returns boolean based on network interfaces present and their attribute values.
+#
+# Can be called with one, or two arguments.
+Puppet::Functions.create_function(:'stdlib::has_interface_with') do
+  # @param interface
+  #   The name of an interface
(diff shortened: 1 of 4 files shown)


lib/puppet/functions/stdlib/has_interface_with.rb
```diff
@@ -0,0 +1,48 @@
...
+    interfaces.key? interface
+  end
+
+  def has_interface_with(kind, value) # rubocop:disable Naming/PredicateName
+    # For compatibility with older version of function that used the legacy facts, alias `ip` with `ipaddress` and `mac` with `macaddress`
+    kind = 'ip' if kind == 'ipaddress'
+    kind = 'mac' if kind == 'macaddress'
+
+    interface = interfaces.find { |_interface, params| params[kind] == value }
+    !interface.nil?
```
ekohl, line 42:
```suggestion
    interfaces.any? { |_interface, params| params[kind] == value }
```

alexjfisher: How did I not think of this?  Doh! Fixed.

bastelfreak (approved):

ekohl (approved):

PR #1321, opened 2023-04-25
Remove deprecated lstrip function

MartyEwings:
**Note:** **Deprecated** from Puppet 6.0.0, this function has been replaced with a
    built-in [`lstrip`](https://puppet.com/docs/puppet/latest/function.html#lstrip) function.
    DOC

PR #1332, opened 2023-04-25
Remove deprecated lstrip function

smortex:
From Puppet 6.0.0, this function has been replaced with a built-in
[`lstrip`](https://puppet.com/docs/puppet/latest/function.html#lstrip)
function.


smortex:
Crap, dupe of #1321.

Issue #1315, opened 2023-04-25
Parser Functions using legacy facts


smoeding:
## Describe the Bug
The two functions `has_interface_with` and `fqdn_rotate` do not work correctly when running the Puppet Agent with the `--no-include_legacy_facts` option that seems to be the default in Puppet 8. 

* `has_interface_with` uses `lookupvar(interfaces)` which is no longer available 
* `fqdn_rotate` calls `lookupvar(::fqdn)` which is also no longer available

## Expected Behavior
Both functions should work as before.

## Environment
 - Version 7.24.0 with option `--no_include_legacy_facts`
 - Platform Debian-11

alexjfisher:
Half fixed in https://github.com/puppetlabs/puppetlabs-stdlib/pull/1326

alexjfisher:
Remaining half of the issue fixed in #1341

PR #1341, opened 2023-04-26
Modernise `fqdn_rotate` function

alexjfisher:
Convert the function to the modern function API as a namespaced function and use the `networking` fact instead of legacy facts.

A non-namespaced shim is also created (but marked deprecated), to preserve compatibility.

Along with #1326 this is the other half of the fix for #1315


spec/functions/fqdn_rotate_spec.rb
```diff
@@ -68,9 +66,9 @@ def fqdn_rotate(value, args = {})
 
     # workaround not being able to use let(:facts) because some tests need
     # multiple different hostnames in one context
-    allow(scope).to receive(:lookupvar).with('::fqdn').and_return(host)
+    allow(subject.func.closure_scope).to receive(:[]).with('facts').and_return({ 'networking' => { 'fqdn' => host } })
```
alexjfisher, line 69: What I really wanted to do was mock the function's private method `fqdn_fact`, but can't figure out how to do this.

bastelfreak (dismissed):

lib/puppet/functions/fqdn_rotate.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+# @summary DEPRECATED.  Use the namespaced function [`stdlib::fqdn_rotate`](#stdlibfqdn_rotate) instead.
```
b4ldr, line 3:
```suggestion
# @summary Rotates an array or string a random number of times, combining the `$fqdn` fact
#    and an optional seed for repeatable randomness.
# @deprecated Use the namespaced function [`stdlib::fqdn_rotate`](#stdlibfqdn_rotate) instead.
```

deprecated should be supported in puppet-strings 4.0.0 which we should be able to use once the bolt dependency is update (https://github.com/puppetlabs/bolt/pull/3204)

lib/puppet/functions/stdlib/fqdn_rotate.rb
```diff
@@ -0,0 +1,66 @@
...
+  end
+
+  def fqdn_rotate_array(input, *seeds)
+    # Check whether it makes sense to rotate ...
+    return input if input.size <= 1
+
+    result = input.clone
+
+    require 'digest/md5'
+    seed = Digest::MD5.hexdigest([fqdn_fact, seeds].join(':')).hex
```
b4ldr, line 46: sa  #1342 ][
```suggestion
    require 'digest/sha2'
    seed = Digest::SHA256.hexdigest([fqdn_fact, seeds].join(':')).hex
```

alexjfisher: I did look at this, but it would be a breaking change...

I don't think anybody using this wants their _deterministic_ rotated things changed just because we think SHA256 would create a better _seed_ for https://github.com/puppetlabs/puppet/blob/f431150d36545626af9ad8602cdd6a7461ca025b/lib/puppet/util.rb#L725

Unwanted changes can mean services restarting, refresh_only execs being trigger etc.

The only point in the `Digest::MD5.hexdigest('seed').hex` was to turn `seed` into a repeatable integer (as `Random.new` only accepts integers for seeds)

And then there's no real point that this seed is then fed to `Random.new`.  All we are trying to do is generate a repeatable number of times, (`offset.times`), to rotate the input characters (or array) by.

Honestly I think `(seed % result.size).times` would have been good enough.

b4ldr: ack makes senses and agree this is minor and doesn't warrant the distruption

b4ldr (dismissed):

LukasAud:
Hey @alexjfisher, looks like there are some conflicts in the branch that need to be addressed before this can be merged. If you can get this sorted, the PR should be mergeable right away.

ic248:
Any progress with this PR, with the latest version we're currently still seeing lots of changes on each Puppet run as mentioned in #1381

spec/functions/fqdn_rotate_spec.rb
```diff
@@ -4,9 +4,9 @@
 
 describe 'fqdn_rotate' do
```
smortex, line 5:
```suggestion
describe 'stdlib::fqdn_rotate' do
```

spec/functions/fqdn_rotate_spec.rb
```diff
@@ -68,9 +66,9 @@ def fqdn_rotate(value, args = {})
 
     # workaround not being able to use let(:facts) because some tests need
     # multiple different hostnames in one context
-    allow(scope).to receive(:lookupvar).with('::fqdn').and_return(host)
+    allow(subject.func.closure_scope).to receive(:[]).with('facts').and_return({ 'networking' => { 'fqdn' => host } })
 
     function_args = [value] + extra
-    scope.function_fqdn_rotate(function_args)
+    scope.call_function('fqdn_rotate', function_args)
```
smortex, line 72:
```suggestion
    scope.call_function('stdlib::fqdn_rotate', function_args)
```

lib/puppet/functions/fqdn_rotate.rb
```diff
@@ -0,0 +1,12 @@
...
+Puppet::Functions.create_function(:fqdn_rotate) do
+  dispatch :deprecation_gen do
+    repeated_param 'Any', :args
+  end
+  def deprecation_gen(*args)
+    call_function('deprecation', 'fqdn_rotate', 'This method is deprecated, please use stdlib::fqdn_rotate instead.')
```
smortex, line 9: With your work on #1377, I think it makes sense to:

```suggestion
    call_function('deprecation', 'fqdn_rotate', 'This method is deprecated, please use stdlib::fqdn_rotate instead.', false)
```

We may also open another PR to be merged when the next major version is released and that restore the failure?

smortex (commented):
That looks good, see some minor in-line comments.  I don't see merge conflicts but a rebase may help @LukasAud.

lib/puppet/functions/fqdn_rotate.rb
alexjfisher: Fixed whilst rebasing by rerunning script.

spec/functions/fqdn_rotate_spec.rb
alexjfisher: This was a deliberate choice to show that the shim does actually correctly invoke the namespaced function.

alexjfisher:
Rebased to fix #1381

lib/puppet/functions/stdlib/fqdn_rotate.rb
ekohl: It may matter on FIPS systems where md5 may not be available, but that's not the point of this PR.

ekohl (approved):

PR #1343, opened 2023-04-27
Rewrite fqdn_rand_string() as a Puppet 4.x function

smortex:
The 3.x function rely on `is_integer()` which is deprecated.  Rewrite it using the more modern puppet 4.x function to rely on data types for better parameters validation.

* Rework the tests to check actual values;
* Rewrite `fqdn_rand_string()` as a Puppet 4.x function.


alexjfisher:
IMO all non core puppet functions should be namespaced. My approach has been to create a non-namespace deprecated shim that we can remove at a later date.

smortex:
@alexjfisher hum, yeah, maybe we can do this. For now, I want to remove the deprecated functions so need a backward compatible way to use `fqdn_rand_string()` in order to open another PR I am preparing ATM.

When these are merged, we can namespace the function and add a deprecation warning in the non-namespaced version in another PR if you are okay with this?

spec/functions/fqdn_rand_string_spec.rb
```diff
@@ -6,68 +6,48 @@
...
+  it { is_expected.to run.with_params('-10').and_raise_error(ArgumentError, %r{parameter 'length' expects an Integer\ value, got String}) }
+  it { is_expected.to run.with_params('string').and_raise_error(ArgumentError, %r{parameter 'length' expects an Integer\ value, got String}) }
+  it { is_expected.to run.with_params([]).and_raise_error(ArgumentError, %r{parameter 'length' expects an Integer value, got Array}) }
+  it { is_expected.to run.with_params({}).and_raise_error(ArgumentError, %r{parameter 'length' expects an Integer value, got Hash}) }
+  it { is_expected.to run.with_params(1, 1).and_raise_error(ArgumentError, %r{parameter 'charset' expects a String value, got Integer}) }
+  it { is_expected.to run.with_params(1, []).and_raise_error(ArgumentError, %r{parameter 'charset' expects a String value, got Array}) }
+  it { is_expected.to run.with_params(1, {}).and_raise_error(ArgumentError, %r{parameter 'charset' expects a String value, got Hash}) }
+  it { is_expected.to run.with_params('100').and_raise_error(ArgumentError, %r{parameter 'length' expects an Integer value, got String}) }
```
smortex, line 20: `'100'` was allowed by `is_integer()` and was removed bellow.

spec/functions/fqdn_rand_string_spec.rb
```diff
@@ -6,68 +6,48 @@
...
+  it { is_expected.to run.with_params(100, nil).and_raise_error(ArgumentError, %r{parameter 'charset' expects a String value, got Undef}) }
```
smortex, line 21: `nil` was accepted with 3.x function and is rejected by 4.x function. It was remove bellow.

lib/puppet/functions/fqdn_rand_string.rb
```diff
@@ -0,0 +1,37 @@
...
+  # @return [String]
+  #
+  # @example Example Usage:
+  #   fqdn_rand_string(10)
+  #   fqdn_rand_string(10, 'ABCDEF!@$%^')
+  #   fqdn_rand_string(10, '', 'custom seed')
+  dispatch :fqdn_rand_string do
+    param 'Integer[1]', :length
+    optional_param 'String', :charset
+    optional_param 'String', :seed
```
alexjfisher, line 22: It looks like this function use to handle seeds the same way as `fqdn_rotate` in that it accepted zero to many.
_It_ even had [tests](https://github.com/puppetlabs/puppetlabs-stdlib/pull/1341/files#diff-f7e08598996c6cc9c8923e4a4524c628def79565a199d188c000e6e66f4e0dcaR28) for this.

To fully maintain compatibility, you'd have to do something like...
```suggestion
    optional_repeated_param 'String', :seeds
```

lib/puppet/functions/fqdn_rand_string.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  def fqdn_rand_string(length, charset = '', seed = nil)
```
alexjfisher, line 25: ... and change the method signature to...
```suggestion
  def fqdn_rand_string(length, charset = '', *seeds)
```

lib/puppet/functions/fqdn_rand_string.rb
```diff
@@ -0,0 +1,37 @@
...
+    charset = charset.chars.to_a
+
+    charset = (0..9).map { |i| i.to_s } + ('A'..'Z').to_a + ('a'..'z').to_a if charset.empty?
+
+    rand_string = ''
+    length.times do |current|
+      rand_string += charset[call_function('fqdn_rand', charset.size, [seed, (current + 1).to_s].compact.join(':')).to_i]
```
alexjfisher, line 32: and finally something more like (untested)...
```suggestion
      rand_string += charset[call_function('fqdn_rand', charset.size, (seeds + [(current + 1).to_s].join(':')).to_i]
```

lib/puppet/functions/fqdn_rand_string.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+# @summary
+#   Generates a random alphanumeric string. Combining the `$fqdn` fact and an
+#   optional seed for repeatable randomness.
+#
+# Optionally, you can specify a character set for the function (defaults to alphanumeric).
+Puppet::Functions.create_function(:fqdn_rand_string) do
```
b4ldr, line 8: ditto re namespacing

lib/puppet/functions/fqdn_rand_string.rb
smortex: Good catch, I put this on the side in my mind and forgot about it.  I amended the commits to fix this and am pushing them 1 by 1 to make sure this is backwards compatible. Thanks!

lib/puppet/functions/fqdn_rand_string.rb
smortex: Yeah, it is the plan :grin: I am currently playing three-cushion billiards:

1. Adjust non-deprecated functions so that they do not rely on deprecated functions; <--- we are here
2. Remove deprecated functions (since they have been deprecated for a long time);
3. "Break things" and namespace functions. <--- what you are talking about

The above is pretty clean in my head but I guess I should dump this somewhere where we can track progress, see what is coming next and so on.  I am really not good at this, and any suggestion is welcome!

b4ldr: ack, feel free to close all comments related to this and we can deal with the general issue in #1346

smortex: Great! I subscribed to the issue. Thanks!

b4ldr (approved):

binford2k (approved):

lib/puppet/functions/fqdn_rand_string.rb
```diff
@@ -0,0 +1,37 @@
...
+  #
+  # @return [String]
+  #
+  # @example Example Usage:
+  #   fqdn_rand_string(10)
+  #   fqdn_rand_string(10, 'ABCDEF!@$%^')
+  #   fqdn_rand_string(10, '', 'custom seed')
+  dispatch :fqdn_rand_string do
+    param 'Integer[1]', :length
+    optional_param 'String', :charset
```
alexjfisher, line 21: Was it necessary to make this a breaking change?  I have code that called the function like this `fqdn_rand_string(20, undef, 'MY_SEED')`

alexjfisher: Fixed in https://github.com/puppetlabs/puppetlabs-stdlib/pull/1367

PR #1344, opened 2023-04-28
Rewrite seeded_rand() as a Puppet 4.x function

smortex:
The 3.x function rely on `is_integer()` which is deprecated. Rewrite it using the more modern puppet 4.x function to rely on data types for better parameters validation.

* Rework the tests to check actual values;
* Rewrite `seeded_rand()` as a Puppet 4.x function.


lib/puppet/functions/seeded_rand.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+# @summary
+#   Generates a random whole number greater than or equal to 0 and less than max, using the value of seed for repeatable randomness.
+Puppet::Functions.create_function(:seeded_rand) do
```
b4ldr, line 5: nit: i think it would be nice to create this as a namespaced function with a legacy deprecated shim e.g.
https://github.com/puppetlabs/puppetlabs-stdlib/pull/1341/files

lib/puppet/functions/seeded_rand.rb
```diff
@@ -0,0 +1,22 @@
...
+  #   A random number greater than or equal to 0 and less than max
+  dispatch :seeded_rand do
+    param 'Integer[1]', :max
+    param 'String', :seed
+  end
+
+  def seeded_rand(max, seed)
+    require 'digest/md5'
+
+    seed = Digest::MD5.hexdigest(seed).hex
```
b4ldr, line 19: i think sha256 would be better but also see comment/discussion here
https://github.com/puppetlabs/puppetlabs-stdlib/pull/1341/files#r1178984906

tl;dr changing the algo would cause changes which may be undesirable.

smortex: Yeah, changing this is not a good idea. Maybe the hash algorithm should be a user settable thing at puppet level and modules would use this transparently, but it is another issue. Will mark this conversation as resolved since that would be an anoying breaking change.

binford2k (approved):

Issue #1282, opened 2022-11-13
Stdlib::Fqdn type accepts ipv4 addresses


canihavethisone:
## Describe the Bug
Specifying a variable as type `Stdlib::Fqdn` allows ipv4 addresses.  This issue is compounded when using `Stdlib::Host` as with the inclusion of `Fqdn` it allows truncated ipv4 addresses such as `10.10.10.10.10`.  This was experienced in stdlib v8.4.0

## Expected Behavior
IPv4 addresses should be validated when using `Stdlib::Host` and not accepted within the `Stdlib:Fqdn` type

## Steps to Reproduce
1. Create a variable with type `Stdlib::Fqdn`
2. Submit valid or truncated ipv4 address, which passes (the regex)

1. Create a variable with type `Stdlib::Host`
2. Submit a truncated ipv4 address, which passes (the regex)

## Environment
 - Version [e.g. 1.27.0]
 - Platform [e.g. Ubuntu 18.04]

## Additional Context
Unsure if this is a flaw in catering for numerical fqdn's, and how it can be addressed.  TL:DR is that using `Stdlib::Host` does not ensure ipv4 addresses are valid

ekohl:
> Unsure if this is a flaw in catering for numerical fqdn's, and how it can be addressed. TL:DR is that using `Stdlib::Host` does not ensure ipv4 addresses are valid

My bet would indeed be on the numberical FQDNs because `Stdlib::Host` has the problem already as shown in https://github.com/puppetlabs/puppetlabs-stdlib/pull/1283.

canihavethisone:
Interesting problem to solve.  I'm wondering if a possible solution should include a test for invalid ipv4, such as 10.10.10.300.

ekohl:
I'd suspect that the host definition should check that the last part (TLD) should be non-numeric. That should catch it.

b4ldr:
>  it allows truncated ipv4 addresses such as 10.10.10.10.10.
FTR this is a valid FQDN

>a host domain name is now allowed to begin with a digit and could legally be entirely numeric
re: [rfc1123 Section 2](https://datatracker.ietf.org/doc/html/rfc1123#section-2)
> I'd suspect that the host definition should check that the last part (TLD) should be non-numeric. That should catch it.

There are two issues with this
[...]

PR #1345, opened 2023-04-28
Rewrite validate_domain_name() as a Puppet 4.x function

smortex:
The 3.x function rely on `is_domain_name()` which is deprecated. Rewrite it using the more modern puppet 4.x function to rely on data types for better parameters validation.
    
While here, adjust the `Stdlib::Fqdn` / `Stdlib::Host` data types to test fully numerical addresses and add a `Stdlib::Dns::Zone`data type that is basically the same but with a trailing dot.


spec/functions/validate_domain_name_spec.rb
```diff
@@ -5,7 +5,6 @@
 describe 'validate_domain_name' do
   describe 'signature validation' do
     it { is_expected.not_to eq(nil) }
-    it { is_expected.to run.with_params.and_raise_error(Puppet::ParseError, %r{wrong number of arguments}i) }
```
smortex, line 8: This is a change: before calling the function without parameters caused an exception, but now it pass (I would have assumed `repeated_param` was one or more but it seems to allow 0 values).

While we can add some code to make match the old behavior, it seems overkill and I decided to just ignore this corner case.

lib/puppet/functions/validate_domain_name.rb
```diff
@@ -0,0 +1,28 @@
...
+  #   validate_domain_name('domain.tld', 'puppet.com', $my_domain_name)
+  #
+  # @exapmle Failing examples (causing compilation to abort)
+  #   validate_domain_name(1)
+  #   validate_domain_name(true)
+  #   validate_domain_name('invalid domain')
+  #   validate_domain_name('-foo.example.com')
+  #   validate_domain_name('www.example.2com')
+  dispatch :validate_domain_name do
+    repeated_param 'Variant[Stdlib::Fqdn, Stdlib::Dns_zone]', :values
```
alexjfisher, line 24: Should you declare that its return type is `undef`?

lib/puppet/functions/validate_domain_name.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+# @summary
+#   Validate that all values passed are syntactically correct domain names.
+#   Fail compilation if any value fails this check.
+Puppet::Functions.create_function(:validate_domain_name) do
```
alexjfisher, line 6: I'm still keen on all functions ported to the new API are available as namespaced versions.
The approach I've been taking is to add a non-namespaced deprecated shim.  See https://github.com/puppetlabs/puppetlabs-stdlib/pull/1341

(We also took this approach originally in extlib.  The deprecated versions were eventually dropped in a major release)

b4ldr: +1

types/fqdn.pp
```diff
@@ -1,2 +1,2 @@
 # @summary Validate a Fully Qualified Domain Name
-type Stdlib::Fqdn = Pattern[/\A(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]*[a-zA-Z0-9])\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\-]*[A-Za-z0-9])\z/]
+type Stdlib::Fqdn = Pattern[/\A((([[:alnum:]]|[[:alnum:]][[:alnum:]-]*[[:alnum:]])\.)*[[:alpha:]]+)\z/]
```
b4ldr, line 2: i dont think we should change to the posix expressions as they also include unicode and although DNS has IDN's many places where you would enter a domain name (including many DNS daemons) expect the name to be in the ASCII punycode format .  At the very least this should be in a separate PR so its clearly highlighted and marked as a change in behaviour  (possibly even backwards incompatible)

types/dns_zone.pp
```diff
@@ -0,0 +1,2 @@
+# @summary Validate a DNS zone name
+type Stdlib::Dns_zone = Pattern[/\A((([[:alnum:]]|[[:alnum:]][[:alnum:]-]*[[:alnum:]])\.)*[[:alpha:]]+\.|\.)\z/]
```
b4ldr, line 2: for dns zones i think its even less likley that consumers will want unicode as you are most likley configuering dns software which preferes punycode

> Knot Resolver does not provide any convenience support for IDN. Therefore everywhere (all configuration, logs, RPZ files) you need to deal with the [xn-- forms](https://en.wikipedia.org/wiki/Internationalized_domain_name#Example_of_IDNA_encoding) of domain name labels, instead of directly using unicode characters.
ref: https://knot-resolver.readthedocs.io/en/stable/modules-policy.html#filters

spec/type_aliases/fqdn_spec.rb
```diff
@@ -22,6 +22,7 @@
         "\nexample",
         "\nexample\n",
         "example\n",
+        '10.10.10.10.10',
```
b4ldr, line 25: been somewhat pedantic this is a valid DNS name but its not a valid IANA DNS name.  i.e. someone could create there own private TLD for '10.'

types/fqdn.pp
b4ldr, line 2: also changing the end of this to `*[[:alpha:]]+` is invalid and breaking.  There are two issues with this
 * A fqdn can contain just one label i.e. com, net, google are all valid FQDN's  as such this type can and will be used to validated hostnames which could contain a numbers e.g. webserver1001 
 * This only assumes the IANA DNS it doesn't account for:
   * private TLDs e.g. 7up may have there own private DNS in there compan
   * Pseudo TLD's e.g. [i2p](https://en.wikipedia.org/wiki/I2P)

Arguably the IANA DNS is the one most users would want to check however i think we should split that out to 

 Stdlib::DNS::IANA 

ill copy this response to #1282

b4ldr (changes requested):

types/fqdn.pp
smortex: Oh :hankey:! I was 100% sure that `[a-z]` included things like `é` and that I was changing nothing here!  Thank you for spotting this, I will revert to the list of allowed chars!

lib/puppet/functions/validate_domain_name.rb
smortex: :100:! Fixed!

types/fqdn.pp
smortex: Reverted!

types/dns_zone.pp
smortex: Fixed. I also included some punycode in the test suite.

types/fqdn.pp
smortex: Reverted and adjusted the test suite accordingly.

b4ldr (approved):
LGTM excluding the namespacing comments

types/dns_zone.pp
```diff
@@ -0,0 +1,2 @@
+# @summary Validate a DNS zone name
+type Stdlib::Dns_zone = Pattern[/\A((([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9-]*[a-zA-Z0-9])\.)+|\.)\z/]
```
b4ldr, line 2: minor nit but id go for Stdlib::DNS::Zone

lib/puppet/functions/validate_domain_name.rb
smortex: I would like to do this after the whole refactoring where I fix non-legacy functions that relied on legacy functions, part of #1346.  So resolving for now.

bastelfreak (approved):

b4ldr (approved):

smortex:
I just realized the PR description and the commit message where not in sync with the updated code. I force-pushed to fix the commit message. Sorry for the inconvenience!

spec/functions/validate_domain_name_spec.rb
smortex: I addressed this in #1351.

PR #1347, opened 2023-04-28
types - dns: add a new dns type with more specific fqdn definitions

### LINKED ISSUE #1282: Stdlib::Fqdn type accepts ipv4 addresses

b4ldr:
This PR introduces a new set of types for validating FQDN's.  it creates a type for the more loose rfc definitions of domain names and a stricter and likely more useful iana type.

This allows us to create a new type that more does to what most users expect i.e. that a dns name is one that works on the internt, without breaking current uses for users that may be using the Stdlib::Fqdn to validate validate DNS names that don't work with the IANA roots.

The intention of this patch would be to deprecate the currnet Stdlib::Fqdn type and encourage users to move to the appropriate Stdlib::DNS::* type which for most users will likely be the stricter Stdlib::DNS::Fqdn type

Note: this PR is intentionally a bit rough to first garner thoughts as to if this is the correct direction

Fixes #1282 (not sure it fixes but want it tagged)

### DIFF
--- a/types/dns/punycode.pp
+++ b/types/dns/punycode.pp
@@ -1,0 +1,1 @@
+type Stdlib::DNS::Punycpde = Pattern[/xn--[a-z0-9]+/]
--- a/types/fqdn.pp
+++ b/types/fqdn.pp
@@ -1,2 +1,2 @@
 # @summary Validate a Fully Qualified Domain Name
-type Stdlib::Fqdn = Pattern[/\A(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]*[a-zA-Z0-9])\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\-]*[A-Za-z0-9])\z/]
+type Stdlib::Fqdn = Stdlib::DNS::Rfc::Fqdn
--- a/types/fqdn.pp
+++ b/types/fqdn.pp
@@ -1,2 +1,2 @@
 # @summary Validate a Fully Qualified Domain Name
-type Stdlib::Fqdn = Stdlib::DNS::Rfc::Fqdn
+type Stdlib::Fqdn = Stdlib::DNS::Fqdn::Rfc
--- a/types/dns/fqdn.pp
+++ b/types/dns/fqdn.pp
@@ -1,0 +1,1 @@
+type Stdlib::DNS::Fqdn = Stdlib::DNS::Fqdn::IANA::ASCII
--- a/types/dns/fqdn/iana/ascii.pp
+++ b/types/dns/fqdn/iana/ascii.pp
@@ -1,0 +1,1 @@
+type Stdlib::DNS::Fqdn::IANA::ASCII = Pattern[/\A(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]*[a-zA-Z0-9])\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\-]*[a-z0-9]*[a-z]+[a-z0-9]*)\z/]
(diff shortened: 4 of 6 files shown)


types/dns/punycode.pp
```diff
@@ -0,0 +1 @@
+type =Stdlib::DNS::Punycpde = Pattern[/xn--[a-z0-9]+/]
```
smortex, line 1:
```
type =Stdlib::DNS::Punycpde = Pattern[/xn--[a-z0-9]+/]
     ^
     `- typo?
```

types/dns/rfc/fqdn.pp
```diff
@@ -0,0 +1 @@
+type Stdlib::DNS::Rfc::Fqdn = Stdlib::Fqdn
```
smortex, line 1: If I understand your point, the actual Pattern from `Stdlib::Fqdn` should be moved here, and `Stdlib::Fqdn` should equal to `Stdlib::DNS::Rfc::Fqdn`.

smortex (commented):
This look very complex, and makes me think that choosing the correct type will be tricky.  They would need an extended documentation to help choose the right one.

I am not sure that such a level of detail is important at the stdlib level, but understand it is required in some advanced cases.  So I have nothing against this, but I do not feel comfortable with it neither :thinking: :shrug: :grin:

types/dns/rfc/fqdn.pp
b4ldr: yes i think that makes senses, updated

b4ldr:
> This look very complex, and makes me think that choosing the correct type will be tricky. 

The intention is that Stdlib::DNS::Fqdn will point to the "correct" or at least most useful type and picky users can pick the more specific type if needed, this is sort of similar to the `Stdlib::IP` namespace

> They would need an extended documentation to help choose the right one.

If people think this is the right direction happy to clean up the pr and add some better docs

 
> I am not sure that such a level of detail is important at the stdlib level, but understand it is required in some advanced cases. 

My intention here is to provide a migration path away from Stdlib::Fqdn which i suspect is not what most people want with something more strict which likely matches what they do want, whilst at the same time provide types for the edge case that may break if we make Stdlib::Fqdn stricter.

types/fqdn.pp
```diff
@@ -1,2 +1,2 @@
 # @summary Validate a Fully Qualified Domain Name
-type Stdlib::Fqdn = Pattern[/\A(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]*[a-zA-Z0-9])\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\-]*[A-Za-z0-9])\z/]
+type Stdlib::Fqdn = Stdlib::DNS::Rfc::Fqdn
```
JonasVerhofste, line 2:
```suggestion
type Stdlib::Fqdn = Stdlib::DNS::Fqdn::Rfc
```
Seems to be a typo :)

JonasVerhofste (commented):
This probably also needs some unit tests in `spec/type_aliases` to detect breaking changes in the future. They will also help with clarifying the difference between `Stdlib::DNS::Fqdn::IANA::ASCII` and `Stdlib::DNS::Fqdn::Rfc`.

jst-cyr:
@b4ldr : I'm reviewing old open PRs and it seems this one didn't get to an approved status a while ago. Do you still have a need for this?

PR #1350, opened 2023-05-04
Rewrite validate_email_address() as a Puppet 4.x function

smortex:
The 3.x function rely on `is_email_address()` which is deprecated. Rewrite it using the more modern puppet 4.x function to rely on data types for better parameters validation.


spec/functions/validate_email_address_spec.rb
```diff
@@ -5,7 +5,6 @@
 describe 'validate_email_address' do
   describe 'signature validation' do
     it { is_expected.not_to eq(nil) }
-    it { is_expected.to run.with_params.and_raise_error(Puppet::ParseError, %r{wrong number of arguments}i) }
```
smortex, line 8: This is a change: before calling the function without parameters caused an exception, but now it pass (I would have assumed repeated_param was one or more but it seems to allow 0 values).

While we can add some code to make match the old behavior, it seems overkill and I decided to just ignore this corner case.

binford2k: did the old version fail the check intentionally or did it blow up? If the function itself failed, then I don't see a reason to maintain that behavior.

smortex: It was explicitly checking if more than 0 values where passed and raised an exception if not.

While we can adjust the code to be something like:

```
  dispatch :validate_email_address do
    param 'Stdlib::Email', :first
    repeated_param 'Stdlib::Email', :maybe_more
  end
```

that feels… meh :neutral_face: …

smortex: Ah, just sow how `min()` check this. Will do the same!

binford2k (approved):

Issue #1346, opened 2023-04-28
RFC: Convert all Puppet 4.x API functions to namespaced variants


alexjfisher:
IMO, only core puppet functions should not have namespaces, and all stdlib functions should start `stdlib::`.

I'd like to suggest that all current 'modern' (non 3.x style) functions get converted to namespaced versions with a non-namespaced deprecated shim replacing the original file in the same way as is done [here](https://github.com/puppetlabs/puppetlabs-stdlib/blob/main/lib/puppet/functions/has_interface_with.rb).

This was also the approach we took in `[extlib](https://github.com/voxpupuli/puppet-extlib/pull/120)` and `[postgresql](https://github.com/puppetlabs/puppetlabs-postgresql/commit/700d2c5bb54b7ea91d518de96e2c7a22318d0afa)`.

In a future major release we could drop the shims similar to how we did in `[extlib 5.0.0](https://github.com/voxpupuli/puppet-extlib/pull/153)`

chelnak:
I think this is a reasonable suggestion.

Lets put it out to the wider community too (i'm sure many have already seen).. and get some votes.

alexjfisher:
If this is the way we want to go, then perhaps either a rake task or some sort of helper (or both!) could reduce the tedious repetitiveness of creating the shims?? @ekohl - Any thoughts?

ekohl:
I wonder if you need to declare them in the same filename as the function. Perhaps you could generate them in a for loop.

alexjfisher:
Ooh, maybe that could work.  Or maybe you do if you want `puppet-strings` to mention them??

alexjfisher:
Actually, I think a function can only be loaded if it's in the correct file.

alexjfisher:
_and_ it has to have a call to `create_function` in it.

>The code loaded from modules/stdlib/lib/puppet/functions/foo.rb does not seem to be a Puppet 4x API function - no create_function call.

smortex:
I am having a look at it ATM.  Some functions might not be namespaced, or more precisely they could be moved to Puppet itself.  I am thinking about:

* `fact()`
* `deprecation()`

Changing `fact(...)` to `stdlib::fact(...)` does not feel right (it is supposed to be some kind of shortcut and fail if we namespace it), and the `deprecation()` feature might also be available without stdlib.

Though?

LukasAud:
[...]

PR #1352, opened 2023-05-04
Remove deprecated functions

smortex:
This PR remove a bunch of legacy functions that are now part of Puppet or which where added to stdlib to allow parameters validation prior to Puppet 4 introduction of data types.

All these functions produce deprecation warnings.

The `has_interface_with()` function deprecation was excuded because it is a "new" one due to the function being namespaced (see #1346 for details).

The `validate_legacy()` function was also kept. It is not used anymore by the module, but prior to removing it we may want to deprecate it.


smortex:
CI currently failing because of #1350

smortex:
[RFC] I started some research to determine how we should remove `validate_legacy()`.

It appear that while there are a lot of [calls to these functions into the wild](https://github.com/search?q=validate_legacy+lang%3APuppet&type=code), they are from numerous vendors and the "puppet-centric-shops" are quite good students with few usage of it :100::

* [Puppetlabs](https://github.com/search?q=validate_legacy+lang%3APuppet+org%3Apuppetlabs&type=code), 1 module affected:
  * puppetlabs/puppetlabs-wsus_client
* [Voxpupuli](https://github.com/search?q=validate_legacy+lang%3APuppet+org%3Avoxpupuli&type=code), 1 module affected:
  * voxpupuli/puppet-dropbear

#### Context

`validate_legacy()` was used with a data type and a validation function. If the value was matched by both the data type and the validation function, it produced no warning, otherwise it produced a warning if the new data type was matched but the legacy validation function was not accepting the value or an deprecation message if the data type was not matched but the legacy validation was accepted.  The validation functions used are the ones being removed by this PR, so `validate_legacy()` needs to be taken care of.

The table bellow summarize the current behavior of `validate_legacy()`:

Data Type | Legacy function | result
----------|-----------------|----------
:heavy_check_mark: pass | :heavy_check_mark: pass | :heavy_check_mark: pass
:heavy_check_mark: pass | :x: fail | :warning: notice (Accepting previously invalid value for target type)
:x: fail | :heavy_check_mark: pass | :warning: deprecation(validate_legacy)
:x: fail | :x: fail | :x: fail

#### Assumptions

The last line (fail/fail) is not expected to be seen into the wild because the compilation failure makes modules unusable.

For the 3 first line, only the first one does not produce annoying deprecation warning / notice, so we expect that such issues have been fixed.

#### Proposal

Update the function to always emit a deprecation message, and only validate the value against the provided data type.  The validation function is therefore completely ignored (but the function prototype is not changed for backwards compatibility).

The table bellow summarize the proposed behavior of `validate_legacy()`:

Data Type | Legacy function | result
----------|-----------------|----------
:heavy_check_mark: pass | :grey_question: ignored | :warning: deprecation(replace validate_legacy() with data type)
:x: fail | :grey_question: ignored | :x: fail

bastelfreak (approved):

lib/puppet/parser/functions/validate_bool.rb
```diff
@@ -1,41 +0,0 @@
...
-          validate_bool(true, true, false, $iamtrue)
-
-      The following values will fail, causing compilation to abort:
-
-          $some_array = [ true ]
-          validate_bool("false")
-          validate_bool("true")
-          validate_bool($some_array)
-      DOC
-  ) do |args|
```
LukasAud, line 30: I dont see any mention of deprecation in this function. Could this Parser Function deletion be a mistake?

smortex: The `validate_bool()` function indeed does not produce deprecation message, but it call `is_bool()` on line 36 which does. I assumed no deprecation message was put here in order to avoid some kind of "duplicate deprecation warning".

LukasAud: Ah I see. Yeah, that looks to be correct.

LukasAud (approved):
LGTM

traylenator:
Amazing `is_integer('1234')` with a string used to be true , `Pattten[/\d+/]` it is.

PR #1353, opened 2023-05-11
Deprecate the `validate_legacy()` function

smortex:
This is a follow-up to #1352 and in particular https://github.com/puppetlabs/puppetlabs-stdlib/pull/1352#issuecomment-1538844946.

I reproduce the content bellow so that we can discuss this, and provide an implementation.

#### Context

`validate_legacy()` was used with a data type and a validation function. If the value was matched by both the data type and the validation function, it produced no warning, otherwise it produced a warning if the new data type was matched but the legacy validation function was not accepting the value or an deprecation message if the data type was not matched but the legacy validation was accepted.  The validation functions used are the ones being removed by this PR, so `validate_legacy()` needs to be taken care of.

The table bellow summarize the current behavior of `validate_legacy()`:

Data Type | Legacy function | result
----------|-----------------|----------
:heavy_check_mark: pass | :heavy_check_mark: pass | :heavy_check_mark: pass
:heavy_check_mark: pass | :x: fail | :warning: notice (Accepting previously invalid value for target type)
:x: fail | :heavy_check_mark: pass | :warning: deprecation(validate_legacy)
:x: fail | :x: fail | :x: fail

#### Assumptions

The last line (fail/fail) is not expected to be seen into the wild because the compilation failure makes modules unusable.

For the 3 first line, only the first one does not produce annoying deprecation warning / notice, so we expect that such issues have been fixed.

#### Proposal

Update the function to always emit a deprecation message, and only validate the value against the provided data type.  The validation function is therefore completely ignored (but the function prototype is not changed for backwards compatibility).

The table bellow summarize the proposed behavior of `validate_legacy()`:

Data Type | Legacy function | result
----------|-----------------|----------
:heavy_check_mark: pass | :grey_question: ignored | :warning: deprecation(replace validate_legacy() with data type)
:x: fail | :grey_question: ignored | :x: fail

### DIFF
--- a/lib/puppet/functions/validate_legacy.rb
+++ b/lib/puppet/functions/validate_legacy.rb
@@ -1,13 +1,14 @@
 # frozen_string_literal: true
 
 # @summary
-#   Validate a value against both the target_type (new) and the previous_validation function (old).
+#   **Deprecated:** Validate a value against both the target_type (new).
 Puppet::Functions.create_function(:validate_legacy) do
-  # The function checks a value against both the target_type (new) and the previous_validation function (old).
+  # The function checks a value against both the target_type (new).
   # @param scope
   #   The main value that will be passed to the method
   # @param target_type
   # @param function_name
+  #   Unused
   # @param value
   # @param args
   #   Any additional values that are to be passed to the method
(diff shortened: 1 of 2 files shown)


david22swan (approved):
LGTM

Was just looking at this issue

PR #1356, opened 2023-05-16
Namespace Puppet 4.x functions

### LINKED ISSUE #1346: RFC: Convert all Puppet 4.x API functions to namespaced variants

smortex:
Fixes #1346

This PR also include:
* #1349


LukasAud:
Hey @smortex, we have finally have completed our maintenance on Stdlib and we are getting ready to cut a major release for the module. This change looks like it would fall under backwards-incompatible and thus, if not merged before the release, there is a high chance it will sit around for a good while before we merge it. 

Can you give us an update on the state of the PR and whether you are interested in continuing to work on it?

smortex:
@LukasAud as it is now it feels almost ready to me, but there are still a few points that may deserve attention.  They are listed in #1346, and that's why this is currently a draft.

chelnak (approved):

PR #1359, opened 2023-05-29
(CONT-1023) - Enhancing deferrable_epp to support nested hash

Ramesh7:
## Summary of Issue

Currently `stdlib::deferrable_epp` doesn't support well if the input variable is nested hash has `Deferrable` value .

## Change

Adding support to evaluate the nested hash values to add support for nested hash.

### DIFF
--- a/lib/puppet/functions/hash_values.rb
+++ b/lib/puppet/functions/hash_values.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+# This function will return list of Hash values, the return value will be Array
+# NOTE : This function is expecting only Hash and return value will be Array
+#
+# @example :
+# $hash = {
+#   "key1" => "value1",
+#   "key2" => { "key2.1" => "value2.1"}
+# }
+# $hash.hash_value
+#
+# Output : ["value1", "value2.1"]
+#
+Puppet::Functions.create_function(:hash_values) do
+  dispatch :hash_values do
+    param 'Hash', :options
+    return_type 'Array'
+  end
+
+  def hash_values(options)
+    options.each_with_object([]) do |(_k, v), values|
+      v.is_a?(Hash) ? values.concat(hash_values(v)) : (values << v)
+    end
+  end
+end
--- a/lib/puppet/functions/nested_values.rb
+++ b/lib/puppet/functions/nested_values.rb
@@ -8,19 +8,19 @@
 #   "key1" => "value1",
 #   "key2" => { "key2.1" => "value2.1"}
 # }
-# $hash.hash_value
+# $hash.nested_values
 #
 # Output : ["value1", "value2.1"]
 #
-Puppet::Functions.create_function(:hash_values) do
-  dispatch :hash_values do
+Puppet::Functions.create_function(:nested_values) do
+  dispatch :nested_values do
     param 'Hash', :options
     return_type 'Array'
   end
 
-  def hash_values(options)
+  def nested_values(options)
     options.each_with_object([]) do |(_k, v), values|
-      v.is_a?(Hash) ? values.concat(hash_values(v)) : (values << v)
+      v.is_a?(Hash) ? values.concat(nested_values(v)) : (values << v)
     end
   end
 end
--- a/spec/functions/hash_values_spec.rb
+++ b/spec/functions/hash_values_spec.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe 'hash_values' do
+  # please note that these tests are examples only
+  # you will need to replace the params and return value
+  # with your expectations
+  it { is_expected.to run.with_params({}).and_return([]) }
+  it { is_expected.to run.with_params({ 'key' => 'value' }).and_return(['value']) }
+  it { is_expected.to run.with_params({ 'key' => { 'key1' => 'value1', 'key2' => 'value2' } }).and_return(['value1', 'value2']) }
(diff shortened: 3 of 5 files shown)


binford2k (dismissed):

lib/puppet/functions/hash_values.rb
```diff
@@ -0,0 +1,26 @@
...
+# NOTE : This function is expecting only Hash and return value will be Array
+#
+# @example :
+# $hash = {
+#   "key1" => "value1",
+#   "key2" => { "key2.1" => "value2.1"}
+# }
+# $hash.hash_value
+#
+# Output : ["value1", "value2.1"]
```
smortex, line 13: :wave: As a Ruby developer, my feeling / experience is that the Puppet language quite match Ruby and my first reaction was that this new `hash_value()` function should be named `values` (to match `Hash#values`) and return `["value1", { "key2.1" => "value2.1"}]`.  Only then I could spot what this change was doing, and the function was precisely needing to work on the flattened hash values…

Naming things is hard, but IMHO we need to emphasis the special behavior of the function in its name, maybe adding something like `nested` or `recursive` in it?

Another possibility would be to add support for Hash to the `flatten` function, but this would break backwards compatibility :cold_face:.

spec/functions/hash_values_spec.rb
```diff
@@ -0,0 +1,14 @@
...
+
+require 'spec_helper'
+
+describe 'hash_values' do
+  # please note that these tests are examples only
+  # you will need to replace the params and return value
+  # with your expectations
+  it { is_expected.to run.with_params({}).and_return([]) }
+  it { is_expected.to run.with_params({ 'key' => 'value' }).and_return(['value']) }
+  it { is_expected.to run.with_params({ 'key' => { 'key1' => 'value1', 'key2' => 'value2' } }).and_return(['value1', 'value2']) }
```
smortex, line 11: Maybe also a test to verify the ordering does not change and the same keys at different levels are not merged / overwritten, e.g.

```
  it { is_expected.to run.with_params({ 'key1' => 'value1', 'key2' => { 'key1' => 'value21', 'key2' => 'value22' }, 'key3' => 'value3' }).and_return(['value1', 'value21', 'value22', 'value3']) }
```

Ramesh7: added spec, Thanks for your review comment.

lib/puppet/functions/nested_values.rb
```diff
@@ -0,0 +1,26 @@
...
+
+# This function will return list of Hash values, the return value will be Array
+# NOTE : This function is expecting only Hash and return value will be Array
+#
+# @example :
+# $hash = {
+#   "key1" => "value1",
+#   "key2" => { "key2.1" => "value2.1"}
+# }
+# $hash.hash_value
```
smortex, line 11: Comment needs updating:

```suggestion
# $hash.nested_values
```

smortex (commented):
This feels more more future proof to me, thanks! Only a minor fix and it's :100: to me!

lib/puppet/functions/nested_values.rb
Ramesh7: 🤦 how I missed it.

Ramesh7: done!!

smortex (approved):
LGTM!

lib/puppet/functions/nested_values.rb
```diff
@@ -0,0 +1,26 @@
...
+# @example :
+# $hash = {
+#   "key1" => "value1",
+#   "key2" => { "key2.1" => "value2.1"}
+# }
+# $hash.nested_values
+#
+# Output : ["value1", "value2.1"]
+#
+Puppet::Functions.create_function(:nested_values) do
```
alexjfisher, line 15: What happened to all new functions being namespaced??

spec/functions/nested_values_spec.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe 'nested_values' do
+  # please note that these tests are examples only
+  # you will need to replace the params and return value
+  # with your expectations
```
alexjfisher, line 8: These comments should not have made it in.

lib/puppet/functions/nested_values.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+# This function will return list of Hash values, the return value will be Array
+# NOTE : This function is expecting only Hash and return value will be Array
```
alexjfisher, line 4: These two lines could be reworded to remove the repetition about the return value.

lib/puppet/functions/nested_values.rb
alexjfisher: Suggest namespacing it _without_ creating a shim as this hasn't made it to a release yet.

smortex: Damned!  Yes, to all of this :sleepy:

spec/functions/nested_values_spec.rb
Ramesh7: addressed in https://github.com/puppetlabs/puppetlabs-stdlib/pull/1363.

lib/puppet/functions/nested_values.rb
Ramesh7: addressed [here](url https://github.com/puppetlabs/puppetlabs-stdlib/pull/1363)

PR #1363, opened 2023-05-31
(CONT-1023) - Addressing review comments

Ramesh7:
Addressing review comments of https://github.com/puppetlabs/puppetlabs-stdlib/pull/1359


alexjfisher:
Great!  It might also be worth adding `@param` tags and regenerating the REFERENCE.md to check it renders nicely.

lib/puppet/functions/stdlib/nested_values.rb
```diff
@@ -1,18 +1,17 @@
 # frozen_string_literal: true
 
 # This function will return list of Hash values, the return value will be Array
-# NOTE : This function is expecting only Hash and return value will be Array
 #
 # @example :
```
alexjfisher, line 5: I think this will render with the example having the literal name `:`

```suggestion
# @example Some text to describe the example
```

lib/puppet/functions/stdlib/nested_values.rb
```diff
@@ -1,18 +1,17 @@
 # frozen_string_literal: true
 
 # This function will return list of Hash values, the return value will be Array
```
alexjfisher, line 3: I think the `@summary` tag should go here somewhere.

lib/puppet/functions/stdlib/nested_values.rb
alexjfisher: It might also be in the wrong place.  Not sure.  Maybe needs to go directly above the dispatch (along with an `@param` tag and a `@return` tag)

alexjfisher: I always find I have to iterate a few times getting the puppet-strings stuff to correctly generate something nice.
Here's an [example](https://github.com/puppetlabs/puppetlabs-stdlib/blob/4120eb26fd51b6197528cf0016ecf3285a3819ba/lib/puppet/functions/stdlib/xml_encode.rb) I think has everything in the right place.

lib/puppet/functions/stdlib/nested_values.rb
```diff
@@ -1,18 +1,17 @@
...
 #
 # @example :
 # $hash = {
 #   "key1" => "value1",
 #   "key2" => { "key2.1" => "value2.1"}
 # }
-# $hash.nested_values
+# $hash.stdlib::nested_values
 #
 # Output : ["value1", "value2.1"]
```
alexjfisher, line 12: I think the whole line will be rendered inside of a puppet marked up code block, so you might want an additional `#` on this line??
```suggestion
# # Output : ["value1", "value2.1"]
```

lib/puppet/functions/stdlib/nested_values.rb
```diff
@@ -0,0 +1,27 @@
...
+#   $hash = {
+#     "key1" => "value1",
+#     "key2" => { "key2.1" => "value2.1"},
+#     "key3" => "value3"
+#   }
+#   $data = $hash.stdlib::nested_values
+#   #Output : ["value1", "value2.1", "value3"]
+Puppet::Functions.create_function(:'stdlib::nested_values') do
+  # @param options A (nested) hash
+  # @return [Data]
```
alexjfisher, line 16: `Data`? Shouldn't this be `Array`? (and I wonder if it's needed given the `return_type` is also specified in the dispatch.

```suggestion
  # @return All the values found in the input hash included those deeply nested.
```

I'm also wondering if this could/should be expanded to handle data in _arrays_ inside in the hash??

binford2k (dismissed):
Looks good. Adding the `@param` tags that I was lazy and forgot to do initially is a good idea also.

lib/puppet/functions/stdlib/nested_values.rb
```diff
@@ -0,0 +1,27 @@
...
+#     "key2" => { "key2.1" => "value2.1"},
+#     "key3" => "value3"
+#   }
+#   $data = $hash.stdlib::nested_values
+#   #Output : ["value1", "value2.1", "value3"]
+Puppet::Functions.create_function(:'stdlib::nested_values') do
+  # @param options A (nested) hash
+  # @return All the values found in the input hash included those deeply nested.
+  dispatch :nested_values do
+    param 'Hash', :options
```
alexjfisher, line 18: `options` is a bit of an odd name for this parameter??  Makes sense in the context of `deferrable_epp`, but without that context, not so much.  Maybe just call it `hash`???

alexjfisher:
@Ramesh7 Cool.  Thanks for the touchups.  I think this is good to go now.  Would you like to squash your commits first?

alexjfisher (approved):

PR #1366, opened 2023-06-02
Pass calling scope to `stdlib::ensure_packages` from shim

### LINKED ISSUE #1365: namespacing ensure_packages breaks a lot of modules
The recent release of stdlib 9 namespaced a lot of functions.
However, namespacing ensure_packages breaks a lot of existing puppet forge modules.

Other namespaced functions like merge and the dropped has_keys have good alternatives.
I don't think ensure_packages does. So I don't think its a good idea to namespace it if the non namespaced version generates
an error. 

So I think it would be better to namespace it, but leave the non namespaced version alone or at worst make  it a warning.
Possibly the same reasoning applies to ensure_resources. But that doesn't seem to be as much of a problem.

### LINKED ISSUE #1374: Non namespaced `ensure_packages` function behaviour changed in 9.0.0
## Describe the Bug
As originally discussed in #1365 the behaviour of `ensure_packages` is subtly different depending on whether the namespaced or non-namespaced version is called.

## Expected Behavior
Package resources created by the function should be identically 'contained' whether they were created via the namespaced or non-namespaced function shim.

## Additional Context
This issue supercedes #1365 where 2 separate issues are discussed and conflated.
A fix for this specific issue has been submitted in #1366

alexjfisher:
Make the `ensure_packages` shim an Internal function and pass scope to the namespaced version so as to not change the behaviour of where packages are contained.

When the function was first ported to the new API, it was discussed that the existing behaviour might not be 'correct', but changing it would be a breaking change that might have consequences for many users.

In namespacing the function in 9.0.0 we accidentally created a situation where the namespaced version worked as before, but the non-namespaced version, (the shim), now behaved differently.

Fixes #1374


bastelfreak (approved):

alexjfisher:
@bastelfreak Hold off on this.  I should make sure it actually works first! (which it doesn't!)

lib/puppet/functions/ensure_packages.rb
```diff
@@ -1,14 +1,13 @@
 # frozen_string_literal: true
 
-# THIS FILE WAS GENERATED BY `rake regenerate_unamespaced_shims`
-
 # @summary DEPRECATED.  Use the namespaced function [`stdlib::ensure_packages`](#stdlibensure_packages) instead.
```
smortex, line 3: Hum, maybe also change this line because it is what triggers the rake task that generated this code:
https://github.com/puppetlabs/puppetlabs-stdlib/blob/main/Rakefile#L95

```suggestion
# @summary DEPRECATED.  Use the scoped and namespaced function [`stdlib::ensure_packages`](#stdlibensure_packages) instead.
```

Very hackish :-/

Otherwise we can search for `Puppet::Functions\.create_function\(.*, Puppet::Functions::InternalFunction\)` and generate different code if we find it.  Also a bit hackish.

alexjfisher: It's more an internal detail.  Not sure we really need to spell it out to users.  Just not break things for them! :)

Are we likely to be running the rake task anymore?

danielparks:
This does not appear to fix the problem described in #1365 — the deprecation warning is treated as an error by default in Puppet 8. Here’s how I tested:

```
❯ puppet --version                                             
8.0.1
❯ cd $(mktemp -d)
❯ ls
❯ git clone git@github.com:puppetlabs/puppetlabs-stdlib.git stdlib
. . .
❯ cd stdlib           
❯ puppet apply --modulepath '..' -e 'ensure_packages("foobar")'
Error: Evaluation Error: Error while evaluating a Function Call, deprecation. ensure_packages. This function is deprecated, please use stdlib::ensure_packages instead. at ["unknown", 1]:["unknown", 0] (line: 1, column: 1) on node marlow.local
❯ gh pr checkout 1366                                             
. . .
Switched to branch 'issue1365'
❯ puppet apply --modulepath '..' -e 'ensure_packages("foobar")'
Error: Evaluation Error: Error while evaluating a Function Call, deprecation. ensure_packages. This function is deprecated, please use stdlib::ensure_packages instead. at ["unknown", 1]:["unknown", 0] (line: 1, column: 1) on node marlow.local
❯ git switch --detach v8.6.0                                      
HEAD is now at b4b5f0c5 Merge pull request #1313 from puppetlabs/v8.6.0_release_prep
❯ puppet apply --modulepath '..' -e 'ensure_packages("foobar")'
Notice: Compiled catalog for marlow.local in environment production in 0.02 seconds
Error: Mac OS X PKG DMGs must specify a package source.
Error: /Stage[main]/Main/Package[foobar]/ensure: change from 'absent' to 'present' failed: Mac OS X PKG DMGs must specify a package source.
Notice: Applied catalog in 0.01 seconds
❯ 
```

Note that the final error is expected since I’m running on macOS.

alexjfisher:
#1365 really should be two issues.  The only bit that is `ensure_packages` specific _is_ fixed by this change though??  Do you want to open a separate issue for whether the deprecate function should cause a compile failure when strict mode is on?  That's always been the case, but I suspect has largely gone unnoticed before as it defaulted to off.

lib/puppet/functions/ensure_packages.rb
alexjfisher: @smortex Now that I've updated the rake task to ignore this file, I've just got what you were getting at!

robertc99:
I can't believe theres been a new release of stdlib and this didnt get included.. Its a serious bug.

alexjfisher:
> I can't believe theres been a new release of stdlib and this didnt get included.. Its a serious bug.

cc @chelnak

jordanbreen28:
@alexjfisher Great catch, This looks good to me! 
Can you @ me directly when this is ready to be be merged.

alexjfisher:
> https://github.com/puppetlabs/puppetlabs-stdlib/issues/1365 really should be two issues.

I've split the original issue into #1374 and #1373 and updated this PRs commit and description to reference #1374

jordanbreen28 (approved):
Thanks again @alexjfisher!
Will merge on green tests

PR #1367, opened 2023-06-08
Fix `fqdn_rand_string` regression

alexjfisher:
The function use to be able to accept `undef` as its second argument.

eg.
```puppet
fqdn_rand_string(10, undef, 'custom seed')
```

Using `undef` to mean 'use the default' is much more conventional than `''` (empty string).

This commit restores the old interface broken in `9.0.0`

### DIFF
--- a/spec/functions/fqdn_rand_string_spec.rb
+++ b/spec/functions/fqdn_rand_string_spec.rb
@@ -14,13 +14,15 @@ describe 'stdlib::fqdn_rand_string' do
   it { is_expected.to run.with_params('string').and_raise_error(ArgumentError, %r{parameter 'length' expects an Integer\ value, got String}) }
   it { is_expected.to run.with_params([]).and_raise_error(ArgumentError, %r{parameter 'length' expects an Integer value, got Array}) }
   it { is_expected.to run.with_params({}).and_raise_error(ArgumentError, %r{parameter 'length' expects an Integer value, got Hash}) }
-  it { is_expected.to run.with_params(1, 1).and_raise_error(ArgumentError, %r{parameter 'charset' expects a String value, got Integer}) }
-  it { is_expected.to run.with_params(1, []).and_raise_error(ArgumentError, %r{parameter 'charset' expects a String value, got Array}) }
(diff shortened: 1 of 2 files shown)


spec/functions/fqdn_rand_string_spec.rb
```diff
@@ -14,11 +14,12 @@
...
-  it { is_expected.to run.with_params(1, []).and_raise_error(ArgumentError, %r{parameter 'charset' expects a String value, got Array}) }
-  it { is_expected.to run.with_params(1, {}).and_raise_error(ArgumentError, %r{parameter 'charset' expects a String value, got Hash}) }
+  it { is_expected.to run.with_params(1, 1).and_raise_error(ArgumentError, %r{parameter 'charset' expects a value of type Undef or String, got Integer}) }
+  it { is_expected.to run.with_params(1, []).and_raise_error(ArgumentError, %r{parameter 'charset' expects a value of type Undef or String, got Array}) }
+  it { is_expected.to run.with_params(1, {}).and_raise_error(ArgumentError, %r{parameter 'charset' expects a value of type Undef or String, got Hash}) }
   it { is_expected.to run.with_params('100').and_raise_error(ArgumentError, %r{parameter 'length' expects an Integer value, got String}) }
-  it { is_expected.to run.with_params(100, nil).and_raise_error(ArgumentError, %r{parameter 'charset' expects a String value, got Undef}) }
+  it { is_expected.to run.with_params(100, nil).and_return(default_charset) }
+  it { is_expected.to run.with_params(100, nil, 'MY_CUSTOM_SEED').and_return(default_charset) }
```
smortex, line 22: Maybe we should include these examples a few line bellow to assert that `nil` and `''` as the second parameter produce the same result?

smortex (dismissed):
LGTM!

spec/functions/fqdn_rand_string_spec.rb
alexjfisher: ok, can do.  Interestingly, `undef` and empty string use to produce the same result because of [this](https://github.com/voxpupuli/puppet-lint-undef_in_function-check#undef-passed-to-a-function).

> (Of Puppet 3 functions)
>
> When functions are called, undef is cast into an empty string. Passing undef to a function is most likely not what you mean to do.

That might explain the history of the way the function was originally coded.

PR #1369, opened 2023-06-13
(CONT-1029) - Adding generic function for deferred function

Ramesh7:
## Change

Adding new function which will accept `Any` type of argument and will return the same, this function will be leveraged for `Deferred` function where we need to defer sensitive variables to get pull at agent level.

### DIFF
--- a/lib/puppet/functions/stdlib/bounce.rb
+++ b/lib/puppet/functions/stdlib/bounce.rb
@@ -1,0 +1,32 @@
+# frozen_string_literal: true
+
+# @summary Method to return the passed values
+#   This function returns the passed value as parameters. If the passed argument is hash then it will return hash.
+#   It will require for deferred function to deferr the different data types like String, Array, Hash & etc.
+#
+# @example Example Usage:
+#   $str = "abc"
+#   $data = $str.stdlib::bounce
+#   #Output : "abc"
+#
+#   $hash = {
+#     "key1" => "value1",
+#   }
+#   $data = $hash.stdlib::bounce
+#   #Output : { "key1": "value1"}
+#
+#   $array = ["item1", "item2"]
+#   $data = $array.stdlib::bounce
+#   #Output : ["item1", "item2"]
+Puppet::Functions.create_function(:'stdlib::bounce') do
+  # @param String, Integer, Array, Hash
+  # @return The return value will be same as passed value.
+  dispatch :bounce do
+    param 'Any', :args
+    return_type 'Any'
+  end
+
+  def bounce(args)
+    args
+  end
+end
--- a/spec/functions/bounce_spec.rb
+++ b/spec/functions/bounce_spec.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe 'stdlib::bounce' do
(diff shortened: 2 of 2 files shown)


b4ldr:
@Ramesh7 are you able to provide more information on the specific use case.  to me both the implementation and the example indicate this function dose nothing.   

Further to this I'm unable to access CONT-1029 if its possible to make it public it would be useful.  if its sensitive perhaps we could at least give [CODEOWNERS](https://github.com/puppetlabs/puppetlabs-stdlib/blob/main/CODEOWNERS) access.

Ramesh7:
> @Ramesh7 are you able to provide more information on the specific use case. to me both the implementation and the example indicate this function does nothing.
> 
> Further to this I'm unable to access CONT-1029 if its possible to make it public it would be useful. if its sensitive perhaps we could at least give [CODEOWNERS](https://github.com/puppetlabs/puppetlabs-stdlib/blob/main/CODEOWNERS) access.

Yeah you got it right, this function is added for Deferred function where we used to defer different data types of values like String, Hash, Array & etc, so the intention here is to add method which just accept the value and return the same so that that can be leveraged by Deferred function so that we don't end-up duplicating with different modules if they has to deal with sensitive values to get defer. 
For string we do have `sprintf` but for other types of data types we don't have.
One of example is https://github.com/puppetlabs/puppetlabs-docker/pull/918/files where we have defined function `docker::env` which accepts array and return same but we want one generic function which can fits everywhere and we don't need to duplicate with each module where ever its require.

Ramesh7:
> > One of example is https://github.com/puppetlabs/puppetlabs-docker/pull/918/files where we have defined function docker::env which accepts array
> 
> But why? Sorry, I'm not getting at what you're trying to achieve here. How does replacing the value for the `environment` parameter at run time with the exact same value you had at compile time help? /me confused.

My intention here is to avoid duplication to each module which needs to implement the `Deferred` function and can use this generic function. Its not specific to `environment`  but its for all the values types which contain some sensitive values which could be Hash/Array/String or etc.
Right now we are duplicating (which means creating generic function at module level) for each modules which has sensitive variable in catalog. I have share few example in above comment which might help to get more context.

alexjfisher:
I've been trying to work out _why_ you are using deferred though...

You're not abusing the functionality added in Puppet 7.17 are you??

From https://dev.to/puppet/updating-puppet-modules-for-deferred-functions-5d6k
> As of Puppet 7.17, this is no longer a problem. Deferred functions are introspected and the return type they declare will be used for type matching. If the function doesn't explicitly declare a return type, Puppet will print a warning, but the compilation will succeed.

Ramesh7:
> I've been trying to work out _why_ you are using deferred though...
> 
> You're not abusing the functionality added in Puppet 7.17 are you??
> 
> From https://dev.to/puppet/updating-puppet-modules-for-deferred-functions-5d6k
> 
> > As of Puppet 7.17, this is no longer a problem. Deferred functions are introspected and the return type they declare will be used for type matching. If the function doesn't explicitly declare a return type, Puppet will print a warning, but the compilation will succeed.

Thanks @alexjfisher for the ref blog this is very helpful. Now I have got real usecase and its limitation of Deferred function.
As you mentioned this is abusing change in stdlib, hence closing the PR.

PR #1377, opened 2023-06-23
Ignore Puppet's `strict` setting when calling function without namespace

### LINKED ISSUE #1373: Calls to `Deprecation` function cause catalog compilation to fail if `strict` setting is set to `error`
In stdlib 9.0 many functions were namespaced and non-namespaced deprecated shims were introduced to maintain compatibility with modules that haven't yet switched over to using the namespaced versions.

This largely works as intended, but (originally in #1365) it's been noted that the `deprecation` function will actually fail a catalog compile when [`strict` is set to `error`](https://github.com/puppetlabs/puppetlabs-stdlib/blob/44b6065d9e43e7b5d2983c91f6753627ceed550b/lib/puppet/functions/deprecation.rb#L32-L33), and that this is also the default in the upcoming Puppet 8.

This makes it very difficult for users to upgrade to stdlib 9 unless they're prepared to turn the `strict` setting back to `off` or `warning` (and there are many other good reasons for setting `strict` to `error`).  Since lots of modules are now also being updated to use the namespaced versions (requiring stdlib 9.0), upgrading a single affected component module now means having to upgrade _all_ modules that use stdlib functions.

Possible solutions include:
* Still leaving the shims documented as deprecated, but removing the calls to the `deprecation` function
* Adjusting deprecation so that it never raises an error

(This issue replaces #1365 where 2 separate issues were being discussed)

alexjfisher:
Previously, when a user had the Puppet setting `strict` set to `error` (which is the default in Puppet 8), a call to one of stdlib's functions via the deprecated non-namespaced function would cause a hard failure instead of just logging a warning and calling the real namespaced function.

In this change, all of our shims have been updated to call `deprecation` with its new third parameter, (`use_strict_setting`), set to `false`. The non-namespaced versions will now only ever log warnings informing users to moved to the namespaced versions. It will not raise exceptions even if `strict` is set to `error`.

This change will make it much easier for users to migrate to stdlib 9 (and to upgrade modules that now depend on stdlib 9)

Fixes #1373

(Note, this PR description was updated after the `deprecation` function update was extracted into its own PR https://github.com/puppetlabs/puppetlabs-stdlib/pull/1378)

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -115,7 +115,7 @@ task :regenerate_unamespaced_shims do
           repeated_param 'Any', :args
         end
         def deprecation_gen(*args)
-          call_function('deprecation', '#{function_name}', 'This function is deprecated, please use stdlib::#{function_name} instead.')
+          call_function('deprecation', '#{function_name}', 'This function is deprecated, please use stdlib::#{function_name} instead.', false)
           call_function('stdlib::#{function_name}', *args)
         end
       end
--- a/lib/puppet/functions/batch_escape.rb
+++ b/lib/puppet/functions/batch_escape.rb
@@ -8,7 +8,7 @@ Puppet::Functions.create_function(:batch_escape) do
     repeated_param 'Any', :args
   end
   def deprecation_gen(*args)
-    call_function('deprecation', 'batch_escape', 'This function is deprecated, please use stdlib::batch_escape instead.')
+    call_function('deprecation', 'batch_escape', 'This function is deprecated, please use stdlib::batch_escape instead.', false)
     call_function('stdlib::batch_escape', *args)
   end
 end
--- a/lib/puppet/functions/ensure_packages.rb
+++ b/lib/puppet/functions/ensure_packages.rb
@@ -7,7 +7,7 @@ Puppet::Functions.create_function(:ensure_packages, Puppet::Functions::InternalF
     repeated_param 'Any', :args
   end
   def deprecation_gen(scope, *args)
-    call_function('deprecation', 'ensure_packages', 'This function is deprecated, please use stdlib::ensure_packages instead.')
+    call_function('deprecation', 'ensure_packages', 'This function is deprecated, please use stdlib::ensure_packages instead.', false)
     scope.call_function('stdlib::ensure_packages', args)
   end
 end
--- a/lib/puppet/functions/fqdn_rand_string.rb
+++ b/lib/puppet/functions/fqdn_rand_string.rb
@@ -8,7 +8,7 @@ Puppet::Functions.create_function(:fqdn_rand_string) do
     repeated_param 'Any', :args
   end
   def deprecation_gen(*args)
-    call_function('deprecation', 'fqdn_rand_string', 'This function is deprecated, please use stdlib::fqdn_rand_string instead.')
+    call_function('deprecation', 'fqdn_rand_string', 'This function is deprecated, please use stdlib::fqdn_rand_string instead.', false)
     call_function('stdlib::fqdn_rand_string', *args)
   end
 end
--- a/lib/puppet/functions/has_interface_with.rb
+++ b/lib/puppet/functions/has_interface_with.rb
@@ -8,7 +8,7 @@ Puppet::Functions.create_function(:has_interface_with) do
     repeated_param 'Any', :args
   end
   def deprecation_gen(*args)
-    call_function('deprecation', 'has_interface_with', 'This function is deprecated, please use stdlib::has_interface_with instead.')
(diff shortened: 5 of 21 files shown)


lib/puppet/functions/deprecation.rb
```diff
@@ -1,38 +1,38 @@
...
+  #   (outputs as an error message), :off (no message / error is displayed) and :warning
+  #   (default, outputs a warning).
   dispatch :deprecation do
     param 'String', :key
     param 'String', :message
+    optional_param 'Boolean', :use_strict_setting
   end
 
-  def deprecation(key, message)
+  def deprecation(key, message, use_strict_setting = true) # rubocop:disable Style/OptionalBooleanParameter
```
bastelfreak, line 19: can the default for `use_strict_setting` maybe come from an environment variable? At the moment CI jobs on Puppet 8 fail if they use the legacy functions and that's a nice way to identify them. It would be helpful to have an easy way to identify them in the future.

chelnak: Can we have both?

chelnak:
I don't feel strongly either way but I wonder if it would be cleaner to have the implementation and updates should be in separate PRs?

lib/puppet/functions/deprecation.rb
alexjfisher: Not sure.  We _could_ but I think I'd lean towards keeping it simple.  The existing environment variable code already looked hacky without adding yet more.  Functions whose behaviour changes based on something other than the function inputs seem like an antipattern to me.  Let's not make it worse?

> that's a nice way to identify them
But was it 'nice'?  The current behaviour has actually forced us to make all modules _depend_ on stdlib 9 in a very short space of time. That may have caused more harm than good.

Maybe a new puppet-lint plugin would be a better way to identify calls to non-namespaced functions?

jhoblitt: I'm pretty sure that the road to hell is paved with env vars...

chelnak: A new linter would certainly do the job.

alexjfisher: Not a blocker though, so maybe let this stew over the weekend and if there aren't any objections get it merged and a release out Monday??

chelnak: Sounds good to me.

Thanks for jumping on this!

jhoblitt (approved):
I think this is probably the least painful path forward from the current situation. I hope that in the future we will do a better job of **not** introducing a deprecation and the replacement facility in the same release. Otherwise, `strict = error` will quickly become meaningless and/or users will change to `string = warn`.

lib/puppet/functions/deprecation.rb
```diff
@@ -1,38 +1,38 @@
...
-    end
+
+    # Do nothing if using strict setting and strict is set to `off`
+    return if use_strict_setting && Puppet.settings[:strict] == :off
+
+    # Fail hard if using strict setting and strict is set to `error`
+    raise("deprecation. #{key}. #{message}") if use_strict_setting && Puppet.settings[:strict] == :error
+
+    # Otherwise raise a soft warning
+    # (unless the STDLIB_LOG_DEPRECATIONS has been set to `false`.  This is mainly for use in rspec-puppet testing to supress noise in logs)
```
ekohl, line 34:
```suggestion
    # (unless the STDLIB_LOG_DEPRECATIONS has been set to `false`.  This is mainly for use in rspec-puppet testing to suppress noise in logs)
```

ekohl (approved):
Makes sense to me.

lib/puppet/functions/deprecation.rb
alexjfisher: Fixed

chelnak:
How's everyone feeling about this today?

bastelfreak:
> I don't feel strongly either way but I wonder if it would be cleaner to have the implementation and updates should be in separate PRs?

 I agree with @chelnak . @alexjfisher what do you think?

alexjfisher:
@chelnak @bastelfreak I've split this into 2 commits and opened https://github.com/puppetlabs/puppetlabs-stdlib/pull/1378 with just the updates to the `deprecation` function.  If you approve and merge that one first, this one can go in second, and I'll update the description to match the single commit.

chelnak:
Thank you!

chelnak:
This PR should be ready for a rebase now!

chelnak (approved):

chelnak (approved):

chelnak (approved):

chelnak (approved):

PR #1378, opened 2023-06-27
Add `use_strict_setting` parameter to `deprecation` function

alexjfisher:
In this change, an optional third parameter is added to the `deprecation` function.  The default behaviour remains the same, but when the third parameter is set to `false`, Puppet's `strict` setting is ignored. This means that even if `strict` is set to `error`, no exception will be raised when this function is called, and a warning will be logged instead.

This is a prerequisite for #1373

### DIFF
--- a/lib/puppet/functions/deprecation.rb
+++ b/lib/puppet/functions/deprecation.rb
@@ -1,38 +1,38 @@
 # frozen_string_literal: true
 
-#   Function to print deprecation warnings, Logs a warning once for a given key.
-#
-# The uniqueness key - can appear once.
-# The msg is the message text including any positional information that is formatted by the
-# user/caller of the method.
-# It is affected by the puppet setting 'strict', which can be set to :error
-# (outputs as an error message), :off (no message / error is displayed) and :warning
-# (default, outputs a warning)  *Type*: String, String.
-#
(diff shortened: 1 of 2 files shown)


chelnak:
Are the acceptance tests expected?

alexjfisher:
> Are the acceptance tests expected

@chelnak Sorry, not following.  How do you mean?  Are you referring to the failing tests that github is highlighting?  (If so, I think those are all the tests marked as `pending` and unrelated)

chelnak:
Ha! I was trying to multitask and failed.

Yes I was referring to the failed test. It's unusual for a pending test to make the run red.

Can you just double check it?

alexjfisher:
> Ha! I was trying to multitask and failed.
> 
> Yes I was referring to the failed test. It's unusual for a pending test to make the run red.

Which failed test?  I can't see any that are red.

chelnak:
I think it was transient! I re-kicked the failed one and it's green now.

chelnak (approved):

alexjfisher:
> I think it was transient! I re-kicked the failed one and it's green now.

ah, thanks.  That makes more sense now.  Thought one of us was seeing things! :)

PR #1382, opened 2023-07-17
pdksync - (maint) - Replace legacy facts in spec/default_facts.yml

jordanbreen28:
(maint) - Replace legacy facts in spec/default_facts.yml
pdk version: `2.7.1`

### DIFF
--- a/spec/default_facts.yml
+++ b/spec/default_facts.yml
@@ -2,7 +2,8 @@
 #
 # Facts specified here will override the values provided by rspec-puppet-facts.
 ---
-ipaddress: "172.16.254.254"
-ipaddress6: "FE80:0000:0000:0000:AAAA:AAAA:AAAA"
+networking:
+    ip: "172.16.254.254"
+    ip6: "FE80:0000:0000:0000:AAAA:AAAA:AAAA"
+    mac: "AA:AA:AA:AA:AA:AA"
 is_pe: false
-macaddress: "AA:AA:AA:AA:AA:AA"


Ramesh7 (approved):

spec/default_facts.yml
```diff
@@ -2,7 +2,8 @@
 #
 # Facts specified here will override the values provided by rspec-puppet-facts.
 ---
-ipaddress: "172.16.254.254"
-ipaddress6: "FE80:0000:0000:0000:AAAA:AAAA:AAAA"
+networking:
+    ip: "172.16.254.254"
+    ip6: "FE80:0000:0000:0000:AAAA:AAAA:AAAA"
+    mac: "AA:AA:AA:AA:AA:AA"
 is_pe: false
```
ekohl, line 9: Isn't the `is_pe` fact dead?

spec/default_facts.yml
```diff
@@ -2,7 +2,8 @@
 #
 # Facts specified here will override the values provided by rspec-puppet-facts.
 ---
-ipaddress: "172.16.254.254"
-ipaddress6: "FE80:0000:0000:0000:AAAA:AAAA:AAAA"
+networking:
+    ip: "172.16.254.254"
```
ekohl, line 6: Why 4 indents, isn't 2 more common?

spec/default_facts.yml
```diff
@@ -2,7 +2,8 @@
 #
 # Facts specified here will override the values provided by rspec-puppet-facts.
 ---
-ipaddress: "172.16.254.254"
-ipaddress6: "FE80:0000:0000:0000:AAAA:AAAA:AAAA"
+networking:
```
ekohl, line 5: This is a bad idea. Looking at the code:
https://github.com/puppetlabs/puppetlabs-stdlib/blob/45a90b8d65c5600ad95f0ca394024f9c0851962d/spec/spec_helper.rb#L19-L32

That's a simple merge, not a deep merge. That means you'll break the networking fact since MANY facts will no longer be present.

ekohl (commented):
Can't we simply delete this whole default facts file? Are there really any tests that depend on it?

I also see this is part of a series, but I really think it's wrong and needs to be rectified.

spec/default_facts.yml
jordanbreen28: Good spot, this explains some of the failures we have seen.
We are going to have to opt with the `deep_merge` like you mentioned, I'll work to get this rolled out.

spec/default_facts.yml
jordanbreen28: hey @ekohl, this fact will be removed in the templates shortly.

spec/default_facts.yml
jordanbreen28: Doh!
Yes, my mistake.
Will update the templates so this is corrected in the next update.

spec/default_facts.yml
ekohl: Perhaps a good excuse to do it natively in rspec-puppet-facts? https://github.com/voxpupuli/voxpupuli-test#fact-handling is how we've done it in voxpupuli-test now ([implementation](https://github.com/voxpupuli/voxpupuli-test/blob/master/lib/voxpupuli/test/facts.rb)), but there's really no good reason (other than being able to roll it out quicker) that it should live in voxpupuli-test.

That said, I still think it would be ideal if you could avoid `default_facts.yml` altogether. A file you always sync to many repositories is something I'd consider an anti-pattern.

jordanbreen28: Totally agree with what you're saying..
I'll raise this with the team to triage and see if we can get it assigned out.

PR #1384, opened 2023-07-20
Add function stdlib::sort_by

XMol:
## Summary
Add new Ruby function `stdlib::sort_by()`, emulating native Ruby's `Enumerable.sort_by()` for Puppet.

## Additional Context
`stdlib::sort_by()` supports sorting Strings, Arrays and Hashes
- Arrays, since that is the most prominent use-case
- Strings, because it is a subset of Iterable, even though I cannot imagine any sensible proc besides `clone`/`copy`, which would be identical to basic `String::sort()`
- Hashes, because that was found to be missing in #1376, even though there likely is little demand for it

## Related Issues
#1376

## Checklist
- [x] 🟢 Spec tests included
- [ ] ⚪Acceptance tests are not changed, some of them fail even on the master branch, so unrelated to my additions
- [x] Manually verified with `puppet apply` version 7.24.0 on RHEL-7.9

### DIFF
--- a/lib/puppet/functions/stdlib/sort_by.rb
+++ b/lib/puppet/functions/stdlib/sort_by.rb
@@ -1,0 +1,49 @@
+# frozen_string_literal: true
+
+# @summary Sort an Array, Hash or String by mapping values through a given block.
+#
+# @example Sort local devices according to their used space.
+#   $facts['mountpoints'].stdlib::sort_by |$m| { $m.dig(1, 'used_bytes') }
+#
+Puppet::Functions.create_function(:'stdlib::sort_by') do
+  # @param ary The Array to sort.
+  # @param block The block for transforming elements of ary.
+  # @return [Array] Returns an ordered copy of ary.
+  dispatch :sort_by_array do
+    param 'Array', :ary
+    block_param 'Callable[1,1]', :block
+  end
+
+  # @param str The String to sort.
+  # @param block The block for transforming elements of str.
+  # @return [String] Returns an ordered copy of str.
+  dispatch :sort_by_string do
+    param 'String', :str
+    block_param 'Callable[1,1]', :block
+  end
+
+  # @param hsh The Hash to sort.
+  # @param block The block for transforming elements of hsh.
+  #              The block may have arity of one or two.
+  # @return [Hash] Returns an ordered copy of hsh.
+  dispatch :sort_by_hash do
+    param 'Hash', :hsh
+    block_param 'Variant[Callable[1,1], Callable[2,2]]', :block
+  end
+
+  def sort_by_iterable(iterable, &block)
+    Puppet::Pops::Types::Iterable.asserted_iterable(self, iterable).sort_by(&block)
+  end
+
+  def sort_by_array(ary, &block)
+    sort_by_iterable(ary, &block)
+  end
+
+  def sort_by_string(str, &block)
+    sort_by_iterable(str, &block).join
+  end
+
+  def sort_by_hash(hsh, &block)
+    sort_by_iterable(hsh, &block).to_h
+  end
+end
--- a/spec/functions/sort_by_spec.rb
+++ b/spec/functions/sort_by_spec.rb
@@ -1,0 +1,54 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe 'stdlib::sort_by' do
+  it { is_expected.not_to be_nil }
+
+  describe 'raise exception with inappropriate parameters' do
+    it { is_expected.to run.with_params.and_raise_error(ArgumentError, Regexp.new('expects 1 argument, got none')) }
+    it { is_expected.to run.with_params([]).and_raise_error(ArgumentError, Regexp.new('expects a block')) }
+    it { is_expected.to run.with_params(:undef).and_raise_error(ArgumentError, Regexp.new("rejected: parameter 'ary' expects an Array value, got Undef")) }
+    it { is_expected.to run.with_params(true).and_raise_error(ArgumentError, Regexp.new("rejected: parameter 'ary' expects an Array value, got Boolean")) }
+    it { is_expected.to run.with_params(1).and_raise_error(ArgumentError, Regexp.new("rejected: parameter 'ary' expects an Array value, got Integer")) }
+    it { is_expected.to run.with_params({}).with_lambda { 1 }.and_raise_error(ArgumentError, Regexp.new('block expects between 1 and 2 arguments, got none')) }
+  end
+
+  # Puppet's each iterator considers Integers, Strings, Arrays and Hashes to be Iterable.
+  unordered_array = ['The', 'quick', 'brown', 'fox', 'jumps', 'over', 'the', 'lazy', 'dog']
+  ordered_array = ['The', 'brown', 'dog', 'fox', 'jumps', 'lazy', 'over', 'quick', 'the']
+  unordered_hash = { 'The' => 'quick', 'brown' => 'fox', 'jumps' => 'over', 'the' => 'lazy', 'dog' => '.' }
+  ordered_hash = { 'dog' => '.', 'brown' => 'fox', 'the' => 'lazy', 'jumps' => 'over', 'The' => 'quick' }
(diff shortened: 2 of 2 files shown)


lib/puppet/functions/stdlib/sort_by.rb
```diff
@@ -0,0 +1,60 @@
...
+    block_param 'Callable[1,1]', :block
+  end
+
+  # @param hsh The Hash to sort.
+  # @param block The block for transforming elements of hsh.
+  # @return [Hash] Returns an ordered copy of hsh.
+  dispatch :sort_by_hash1 do
+    param 'Hash', :hsh
+    block_param 'Callable[1,1]', :block
+  end
```
smortex, line 31: I guess we should document that with this form, the block get passed an array of 2 items, the first being the key, and the second being the value?

lib/puppet/functions/stdlib/sort_by.rb
```diff
@@ -0,0 +1,60 @@
...
+
+  # @param hsh The Hash to sort.
+  # @param block The block for transforming elements of hsh.
+  # @return [Hash] Returns an ordered copy of hsh.
+  dispatch :sort_by_hash2 do
+    param 'Hash', :hsh
+    block_param 'Callable[2,2]', :block
+  end
```
smortex, line 39: Same here, we should document how it is different from the previous form by accepting a block that receive the key and the value as 2 arguments.

lib/puppet/functions/stdlib/sort_by.rb
```diff
@@ -0,0 +1,60 @@
...
+  def sort_by_string(str, &block)
+    sort_by_iterable(str, &block).join
+  end
+
+  def sort_by_hash1(hsh, &block)
+    sort_by_iterable(hsh, &block).to_h
+  end
+
+  def sort_by_hash2(hsh)
+    sort_by_iterable(hsh) { |e| yield(*e) }.to_h
```
smortex, line 58: `e` makes me think, I think this is more readable with explicit names:
```suggestion
    sort_by_iterable(hsh) { |key, value| yield(key, value) }.to_h
```

lib/puppet/functions/stdlib/sort_by.rb
```diff
@@ -0,0 +1,60 @@
...
+  def sort_by_array(ary, &block)
+    sort_by_iterable(ary, &block)
+  end
+
+  def sort_by_string(str, &block)
+    sort_by_iterable(str, &block).join
+  end
+
+  def sort_by_hash1(hsh, &block)
+    sort_by_iterable(hsh, &block).to_h
```
smortex, line 54: This `to_h` makes `sort_by` behave differently that Ruby's `sort_by` which return an array when sorting a Hash (you have to manually `Hash(...)` the result if you want a hash).

I wonder if we should match Ruby or not.  As a Puppet and Ruby developer, I trend to expect some consistency, but obviously the Puppet language is not Ruby so who am I to expect this :shrug:?  Maybe some people have an idea about this?

lib/puppet/functions/stdlib/sort_by.rb
XMol: `sort_by_hash2` is written this way, because Ruby's `sort_by` expects a block that has just one parameter. I thought, this construct was needed to align `stdlib::sort_by($hash) |$key, $value| { ... }` with that. Though it looks like I was mistaken and Ruby handles that transparently:
```ruby
hsh = { 'The' => 'quick', 'brown' => 'fox', 'jumps' => 'over', 'the' => 'lazy', 'dog' => '.' }
hsh.sort_by { |e| e[1] } # => [["dog", "."], ["brown", "fox"], ["the", "lazy"], ["jumps", "over"], ["The", "quick"]]
hsh.sort_by { |k, v| v } # => [["dog", "."], ["brown", "fox"], ["the", "lazy"], ["jumps", "over"], ["The", "quick"]]
```
My delusion was reinforced by `sort_by()` accepting blocks with arity 2 on Arrays and _not_ interpreting that as index-element-tuple, but rather element-nil-tuple. In other words, `sort_by()` does _not_ use `Enumerable.each_with_index` when appropriate.

XMol: On that note, [Puppet's `each()`](https://github.com/puppetlabs/puppet/blob/8.0.1_release/lib/puppet/functions/each.rb#L153) explicitly calls Ruby's `each_with_index()` in this case.

lib/puppet/functions/stdlib/sort_by.rb
XMol: Keeping consistency means that we can simplify the code enormously: only `sort_by_iterable()` is needed, if Ruby can handle blocks with two or a single parameter transparently for us. I went out of my way to distinct between Strings, Arrays and Hashes, precisely because _my_ expectation was that `sort_by()` should return the same data structure that it was fed. If we cannot find an objective argument for either interpretation, we will be unable to decide the "correct" resolution, I guess.

lib/puppet/functions/stdlib/sort_by.rb
XMol: With your other hints I decided to collapse `sort_by_hash1` and `sort_by_hash2`, plus explicitly stating that single and double parameter blocks are accepted.

lib/puppet/functions/stdlib/sort_by.rb
XMol: With the collapsed `sort_by_hash()`, this discussion can be resolved, too.

lib/puppet/functions/stdlib/sort_by.rb
smortex: > _my_ expectation was that `sort_by()` should return the same data structure that it was fed.

Good point. I think that help: Puppet can currently sort a String and return a String, something Ruby cannot do because `String#sort` does not exist.  So I am fine with returning a Hash when given a Hash because it makes more sense IMHO!

lib/puppet/functions/stdlib/sort_by.rb
```diff
@@ -0,0 +1,49 @@
...
+    block_param 'Callable[1,1]', :block
+  end
+
+  # @param hsh The Hash to sort.
+  # @param block The block for transforming elements of hsh.
+  #              The block may have arity of one or two.
+  # @return [Hash] Returns an ordered copy of hsh.
+  dispatch :sort_by_hash do
+    param 'Hash', :hsh
+    block_param 'Variant[Callable[1,1], Callable[2,2]]', :block
```
smortex, line 31: Is this different than:

```suggestion
    block_param 'Callable[1,2]', :block
```

smortex (approved):
This looks good, I am curious what other people think about it.

XMol:
Regarding the pending signature of the CLA agreement, I'm still waiting for a reply from our legal department... No estimated timeline, of course.

XMol:
Since our legal department has not replied to me in 3 months, I sign the CLA at my own risk. Stalling this MR for an unknown amount of time just because I'm trying to follow the rules is ridiculous.

XMol:
Hi project maintainers,

is anything else missing to merge this PR? I've signed the CLA months ago and at least one reviewer gave his approval (thanks for the support, @smortex).

Kind regards,  
Xavier.

bastelfreak:
@XMol can you please rebase against the main branch?

XMol:
> @XMol can you please rebase against the main branch?

Will do...

bastelfreak:
As I wrote, please do a *rebase*, not a merge.

XMol:
> As I wrote, please do a _rebase_, not a merge.

Sorry, I understood that, but I wasn't sure what GitHub would do when I press "sync fork". :-)

bastelfreak:
yeah the button is shit :(

XMol:
Guess we now need to wait for the two pipelines to eventually fail in about 4 hours? The Acceptance tests on Ubuntu-ARM seem to be broken...

bastelfreak (approved):

XMol:
Thanks for the prompt reaction, @bastelfreak.

PR #1386, opened 2023-07-25
Add stdlib::has_function

seanmil:
## Summary
Add new `stdlib::has_function` function to test for function availability.

## Additional Context
The old `is_function_available` function was deprecated for quite some time and removed.
It did not support testing for availability of modern functions, only 3.x style.  This new
function fully supports the various types of functions and restores the ability to
test if certain functions are available.

## Checklist
- [x] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.


smortex (approved):
Looks great!

PR #1392, opened 2023-08-28
Remove unused parser deprecation function

ekohl:
There is already a modern function API definition for the deprecation function, so this ends up being unused.

Fixes: 72d236595135 ("(MODULES-3529)add deprecation function")


bastelfreak (approved):

alexjfisher:
I wasn't 100% sure what happens when a legacy function calls `deprecation`.  Does it definitely find the api v4 version?

PR #1396, opened 2023-09-15
Fix fqdn_rotate determinism

DLDur:
This was changed in e2d8b18 and ever since fqdn_rotate has changed result on every successive run.

## Summary
Seed fqdn_rotate with the appropriate fact to restore consistency in its output.

## Additional Context
Issue was identifed after upgrading to stdlib 9.0.0  - files with contents depending on result of fqdn_rotate were refreshing every run.

## Related Issues (if any)
May need to review original reason for the original change in e2d8b18

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.
- [x ] Manually verified. (For example `puppet apply`)

### DIFF
--- a/lib/puppet/parser/functions/fqdn_rotate.rb
+++ b/lib/puppet/parser/functions/fqdn_rotate.rb
@@ -36,7 +36,7 @@ DOC
 
   elements = result.size
 
-  seed = Digest::MD5.hexdigest([lookupvar('::fqdn'), args].join(':')).hex
+  seed = Digest::MD5.hexdigest([self['facts']['networking']['fqdn'], args].join(':')).hex
   # deterministic_rand() was added in Puppet 3.2.0; reimplement if necessary
   if Puppet::Util.respond_to?(:deterministic_rand)
     offset = Puppet::Util.deterministic_rand(seed, elements).to_i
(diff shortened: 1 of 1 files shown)


lib/puppet/parser/functions/fqdn_rotate.rb
```diff
@@ -36,7 +36,7 @@
 
   elements = result.size
 
-  seed = Digest::MD5.hexdigest([lookupvar('facts'), args].join(':')).hex
+  seed = Digest::MD5.hexdigest([lookupvar('::fqdn'), args].join(':')).hex
```
ekohl, line 39: The legacy fact shouldn't be used anymore.

ekohl (changes requested):
I'd prefer https://github.com/puppetlabs/puppetlabs-stdlib/pull/1341 to be finished, but I'd accept a PR that only updates to the modern fqdn fact.

alexjfisher:
> I'd prefer #1341 to be finished, but I'd accept a PR that only updates to the modern fqdn fact.

let me see if I can fit this in.

alexjfisher:
Fixed in #1341

PR #1195, opened 2021-07-22
(MODULES-11126) Replacing URI.escape with URI::DEFAULT_PARSER

valleedelisle:
URI.escape was deprecated in ruby-2.7 and removed in ruby-3.x. This
change is breaking stdlib. This is described in detail here [1].

Replacing URI.escape with URI::DEFAULT_PARSER is working properly.

This is affecting openstack-tripleo [2]

[1] https://bugs.ruby-lang.org/issues/17309
[2] https://bugs.launchpad.net/tripleo/+bug/1937166

PR #1402, opened 2023-10-17
Revert "(CONT-801) Deprecate uriescape.rb"

### LINKED ISSUE #1401: uriescape function deprecated on P7 - no alternative on P8
## Describe the Bug
With puppet 8

```
# --- Caused by: ---
      # Puppet::ParseError:
      #   Puppet: This function is not available in Puppet 8. URI.escape no longer exists as of Ruby 3+.
      #   ./spec/fixtures/modules/stdlib/lib/puppet/parser/functions/uriescape.rb:23:in `block in <module:Functions>'
```

## Environment
- Puppet 8

I don't see any alternative function to use ?

traylenator:
## Summary

This reverts commit 799d608939b90effb7cfa7d0054a3985c78db982.

While uriescape was deprecated for Puppet 8 in https://github.com/puppetlabs/puppetlabs-stdlib/pull/1307 it was already fixed earlier for Puppet 8 and ruby 3 in https://github.com/puppetlabs/puppetlabs-stdlib/pull/1195

It is unclear to me why this function was deprecated.

## Additional Context
Function was deprecated on Puppet 8.

## Related Issues (if any)
* Fixes #1401

## Checklist
- [x ] 🟢 Spec tests.
- [x ] 🟢 Acceptance tests.
- [x ] Manually verified. (For example `puppet apply`)

### DIFF
--- a/spec/functions/uriescape_spec.rb
+++ b/spec/functions/uriescape_spec.rb
@@ -39,9 +39,5 @@ describe 'uriescape' do
       it { is_expected.to run.with_params(['one}', 'two']).and_return(['one%7D', 'two']) }
       it { is_expected.to run.with_params(['one}', 1, true, {}, 'two']).and_return(['one%7D', 1, true, {}, 'two']) }
     end
-  else
-    describe 'raising errors in Puppet 8' do
-      it { is_expected.to run.with_params([]).and_raise_error(Puppet::ParseError, %r{This function is not available in Puppet 8. URI.escape no longer exists as of Ruby 3+.}) }
-    end
   end
 end
--- a/lib/puppet/parser/functions/uriescape.rb
+++ b/lib/puppet/parser/functions/uriescape.rb
@@ -15,13 +15,8 @@ module Puppet::Parser::Functions
     @return [String]
       a string that contains the converted value
 
-    > **Note:**  **Deprecated:** Starting Puppet 8, our Ruby version has upgraded to 3.2.
-    Therefore, its no longer possible to call URI.escape as it was deprecated by 2.7 and removed completely by 3+.
-    This function should be removed once Puppet 7 is no longer supported.
   DOC
   ) do |arguments|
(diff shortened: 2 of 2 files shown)


traylenator:
CLA signed in agreement between CERN IT and Puppetlabs.

ekohl (changes requested):
In case that we do proceed with this: you need to update the comments on line 6 & 7.

On the question if we should proceed: I'm not sure if this is a good idea, I don't know. there's a good reason it was dropped in Ruby 3: the whole idea behind the function is sketchy. I'm more a fan of explicit URL construction where you pass the parameters unescaped and then it correctly constructs the URL.

spec/functions/uriescape_spec.rb
ekohl, line 1: Looks like lines 6 & 7 in this file also need to be dropped.

ekohl (changes requested):

traylenator:
Redundant comments removed.

I only use it one place to escape `PS Scavenge` and `PS MarkSweep` before we pass them as parameters to a
puppetserver beans endpoint.

`java.lang:name=PS%20Scavenge` 

when pulling metrics out of PuppetServer.

Does a url from component function exist ?

ekohl:
https://docs.knapsackpro.com/2020/uri-escape-is-obsolete-percent-encoding-your-query-string suggests `URI.encode_www_form_component` would be the proper replacement. It only escapes a URL parameter

jst-cyr:
@traylenator : Is this change still needed? The function was deprecated quite some time ago and now we're onto Puppet Core 9 with Ruby 4, I'm wondering if we can possibly close this out?

Just in case, did a rebase to double-check we are still green on Puppet 8 spec tests.

traylenator:
It's so long ago. If I ever cared about this I will have worked around it since them.

Thanks for chasing up this stuff .

Steve

PR #1403, opened 2023-10-18
pdksync - (CAT-1256) Roll out puppetlabs-puppet-lint and puppetlabs-rspec-puppet through spec_helper v7.0

GSPatton:
(CAT-1256) Roll out puppetlabs-puppet-lint and puppetlabs-rspec-puppet through spec_helper v7.0
pdk version: `2.7.1`


b4ldr (approved):

ekohl (changes requested):
You also depend on `voxpupuli-puppet-lint-plugins` which pulls in puppet-lint 4.x: https://rubygems.org/gems/voxpupuli-puppet-lint-plugins. So you will end up with **both** puppetlabs-puppet-lint and puppetlabs-rspec-puppet present.

PR #1404, opened 2023-10-23
(CAT-1521) - Adding RHEL-9-ARM in CI pipeline

Ramesh7:
## Summary

Adding RHEL-9-ARM in CI pipeline

## Additional Context
N/A

## Related Issues (if any)
N/A

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -16,7 +16,8 @@
       "operatingsystemrelease": [
         "7",
         "8",
-        "9"
+        "9",
+        "9-arm"
       ]
     },
     {


metadata.json
```diff
@@ -16,7 +16,8 @@
       "operatingsystemrelease": [
         "7",
         "8",
-        "9"
+        "9",
+        "9-arm"
```
ekohl, line 20: Is this valid? I always thought `operatingsystemrelease` had to match the `operatingsystemrelease` fact that Facter returned.

Ramesh7: sorry @ekohl, was not able to get much from your comment but I always went with https://github.com/puppetlabs/puppet_litmus/blob/main/exe/matrix_from_metadata_v2#L153 and based on that its expecting `operatingsystemrelease`.

ekohl: I think the spec is more important than what Litmus does. https://www.puppet.com/docs/puppet/7/modules_metadata.html#metadata-operating-system states:
> The operatingsystemrelease accepts an array of strings. The Forge displays these versions on module pages, and you can format them in whatever way makes sense for the operating system in question.

So that's not very specific, but it is important to note that the forge will list them. As a user I could think that only RHEL 9 supports ARM, but not RHEL 8 etc. That can be considered misleading.

I always interpreted it that the values it for `operatingsystemrelease` in `metadata.json` should match what Facter returned for `operatingsystemrelease` or `operatingsystemmajrelease` (or nowadays their modern equivalents).

Ramesh7: got it now, thanks for explanation. I will check with Agent team about factor values for  `operatingsystemrelease` or `operatingsystemmajrelease` and will update here.
Thanks @ekohl again!!

Ramesh7: Looks like @jordanbreen28 already handled this into litmus and we don't need this PR anymore.
Hence closing.

PR #1406, opened 2023-11-07
Correct casing of Stdlib::IP::Address

ekohl:
While Puppet doesn't appear to care about IP vs Ip, but Kafo does and can't find the correct type. This aligns the usage with the actual name in the file.

Fixes: fcbd4267fd83 ("Remove deprecated Stdlib::Compat::* data types")

### DIFF
--- a/types/host.pp
+++ b/types/host.pp
@@ -1,2 +1,2 @@
 # @summary Validate a host (FQDN or IP address)
-type Stdlib::Host = Variant[Stdlib::Fqdn, Stdlib::Ip::Address]
+type Stdlib::Host = Variant[Stdlib::Fqdn, Stdlib::IP::Address]
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -5619,7 +5619,7 @@ Alias of `Pattern[/(?i:\Ahttps?:\/\/.*\z)/]`
 
 Validate a host (FQDN or IP address)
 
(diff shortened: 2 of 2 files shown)


bastelfreak (approved):

types/host.pp
```diff
@@ -1,2 +1,2 @@
 # @summary Validate a host (FQDN or IP address)
-type Stdlib::Host = Variant[Stdlib::Fqdn, Stdlib::Ip::Address]
+type Stdlib::Host = Variant[Stdlib::Fqdn, Stdlib::IP::Address]
```
ekohl, line 2: This now matches the casing here: https://github.com/puppetlabs/puppetlabs-stdlib/blob/883ba57ada149b32deff4cd1ef51de3bb51a9b30/types/ip/address.pp#L2

ekohl:
I think the test failures are unrelated

types/host.pp
smortex: Interesting :thinking:.  For some reason I stick to `\A[[:upper:]][[:lower:]]*\z` but it seems having more capital letters is valid…  I guess what annoyed me is that I wanted to camel_case the type name for the file name as we do with Ruby but this is not supported…

smortex (approved):
LGTM!

bastelfreak:
restarted failed jobs.

PR #1415, opened 2023-12-23
Rewrite loadjson to use the modern function API

### LINKED ISSUE #1414: loadjson returns StringIO data which doesn't work in Puppet 8
## Describe the Bug
When using loadjson in a puppet manifest with small json files, the returning Data values come in as a StringIO instead of String value which Puppet 8 can't cast properly.

## Expected Behavior
Return a hash of Strings.

## Steps to Reproduce
1. Use loadjson as documented to pull in a json file that you want to convert to hash.
2. Run puppet agent and see the error.

## Environment
 - RedHat Enterprise Linux 8
 - Puppet 8.3.0

## Additional Context
I was able to fix this by editing the loadjson ruby code lines 56 and 63 changing from
JSON.parse(content) || args[1]
to
JSON.parse(content.string) || args[1]

ekohl:
## Summary

This also resolves a bug where JSON.parse returned a StringIO. To properly catch this, the tests are rewritten to avoid mocking where possible. Exceptions are with external URLs and where a failure is expected.

## Related Issues (if any)

Fixes #1414

## Checklist
- [x] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.


lib/puppet/functions/loadjson.rb
```diff
@@ -0,0 +1,58 @@
...
+# @summary
+#   Load a JSON file containing an array, string, or hash, and return the data
+#   in the corresponding native data type.
+#
+# @example Example Usage:
+#   $myhash = loadjson('/etc/puppet/data/myhash.json')
+#   $myhash = loadjson('https://example.local/my_hash.json')
+#   $myhash = loadjson('https://username:password@example.local/my_hash.json')
+#   $myhash = loadjson('no-file.json', {'default' => 'value'})
+Puppet::Functions.create_function(:loadjson) do
```
alexjfisher, line 12: At the same time, shouldn't we namespace this function and create a (deprecated) shim like we started with a lot of the other functions?

lib/puppet/functions/loadjson.rb
```diff
@@ -0,0 +1,58 @@
...
+    raise e unless default
+
+    default
+  end
+
+  def load_json_source(source)
+    if Puppet::Util::Package.versioncmp(Puppet.version, '8.0.0').negative?
+      PSON.load(source)
+    else
+      JSON.load(source)
```
alexjfisher, line 55: Rubocop is complaining here. Are you specifically using `JSON.load` over `JSON.parse` to address the bug you mentioned?

lib/puppet/functions/loadjson.rb
```diff
@@ -0,0 +1,58 @@
...
+    param 'String[1]', :path
+    optional_param 'Data', :default
+    return_type 'Data'
+  end
+
+  def loadjson(path, default = nil)
+    if path.start_with?('http://', 'https://')
+      require 'open-uri'
+      begin
+        content = URI.open(path) { |f| load_json_source(f) }
```
alexjfisher, line 32: What happened to all the basic authentication related code and tests?

lib/puppet/functions/loadjson.rb
ekohl: It's copied from the old code to remain compatible.

lib/puppet/functions/loadjson.rb
ekohl: Good point. I assumed it would just use the credentials from the URI, but it doesn't.

lib/puppet/functions/loadjson.rb
ekohl: I didn't consider this.

jst-cyr:
@ekohl : I rebased this PR to get it on latest. I'm sorry it's been so long since anybody has looked at this. Do you feel this is still a PR that would be valuable to add to stdlib?

jst-cyr:
Closing this PR for now until a need rises for this in the future at which point we can re-open!

Issue #1260, opened 2022-07-26
Convert the Legacy Puppet functions to the Modern format


david22swan:
Since the release of the Modern Puppet function format several of the functions and all new ones to be added have been written in it, however the vast majority of the original functions remain in the Legacy format.

While they are still functional and will not currently cause any errors in their use, they are still outdated and using a format that has been fazed out, potential causing potential problems down the line.

As such it would be best to update them now before any issues can arise.

b4ldr:
What's your view on if they should be name spaced with ``stdlib::`.  My personal view is that stdlib is a bit more special and unique then most modules as such we could drop the requirement to namespace all (new or refactored) functions

ekohl:
Somehow this doesn't link back to https://github.com/puppetlabs/puppetlabs-stdlib/pull/1259 but that's where this came up. In https://github.com/puppetlabs/puppetlabs-stdlib/pull/1259#issuecomment-1197839997 I stated pretty much the same as @b4ldr that I do feel stdlib is special and can be in the global namespace.

barriserloth:
As of v9.0.0, all deprecated functions should be removed or updated, please reopen this if there are still any left hanging around. Thank you!

PR #1417, opened 2024-02-07
Deprecate `time` function

alexjfisher:
`time` was both an old API ruby function, and not needed since Puppet 4.

In this commit, the `parser` function is removed and replaced by a non-namespaced APIv4 function. This is done to preserve compatibility before the function is removed completely in a later release.

The ruby function calls a Puppet language function which implements the existing behaviour using core Puppet functionality.

The original function did something weird with an optional `timezone` parameter. Nobody really remembers why. It appears that very early versions of Ruby may have been buggy and needed this as a workaround.

To be clear, seconds since the epoch are the same _everywhere_. Timezone is not relevant.

## Additional Context

Also see https://github.com/puppetlabs/puppetlabs-stdlib/commit/ad4ca4cc3486b0eaaa813595907521ce35c970ce
This also relates to https://github.com/puppetlabs/puppetlabs-stdlib/issues/1260 as it removes another function completely from `lib/puppet/parser/functions`.


bastelfreak (approved):

smortex (approved):
LGTM!

ekohl:
> To be clear, seconds since the epoch are the same _everywhere_. Timezone is not relevant.

I'm not entirely sure this is true. I think epoch has a timezone, so the way it could work is that you set the local timezone (like `Europe/Amsterdam`). Then when you get the current time would convert it to Epoch's timezone. Then convert that time to seconds. That's the only way that I can think of that the timezone parameter makes sense, but it is confusing to say the least (as all timezones are).

ekohl (approved):
:+1: to deprecating this.

PR #1418, opened 2024-02-14
PoC/RFC: Support sensitive values in `to_json_pretty`

alexjfisher:
The idea is being able to do something like this, (where just the call to `to_json_pretty` is explicitly `Deferred` - needed because `node_encrypt` returns a `Deferred`)

```puppet
file { '/etc/my_secret.json':
  content => Deferred('to_json_pretty',[{
    username => 'myuser',
    password => lookup('my_eyaml_secret').node_encrypt::secret,
  }]),
}
```

... instead of having to also explicitly defer `unwrap` and `Sensitive` and end up with a huge mess similar to...

```puppet
file { '/etc/my_secret.json':
  content => Deferred('Sensitive', [
    Deferred('to_json_pretty',[{
      username => 'myuser',
      password => Deferred('unwrap', [lookup('my_eyaml_secret').node_encrypt::secret]),
    }])
  ]),
}
```

The thought behind `rewrap_sensitive_data` is it makes it easy to extend this functionality into other similar functions, (`to_yaml`, `to_toml` etc.)

Later, we might consider adding a `deferrable_to_XXX` functions to further simplify this sort of use-case.

No spec tests yet.  Just wanted to get some early feedback before taking it further.


lib/puppet/functions/stdlib/to_json_pretty.rb
```diff
@@ -67,8 +67,10 @@ def to_json_pretty(data, skip_undef = false, opts = nil)
     end
 
     data = data.compact if skip_undef && (data.is_a?(Array) || Hash)
-    # Call ::JSON to ensure it references the JSON library from Ruby's standard library
-    # instead of a random JSON namespace that might be in scope due to user code.
-    JSON.pretty_generate(data, opts) << "\n"
+    call_function('stdlib::rewrap_sensitive_data', data) do |unwrapped_data|
+      # Call ::JSON to ensure it references the JSON library from Ruby's standard library
+      # instead of a random JSON namespace that might be in scope due to user code.
+      ::JSON.pretty_generate(unwrapped_data, opts) << "\n"
```
alexjfisher, line 73: Leading colons put back in to match existing comment.  See https://github.com/puppetlabs/puppetlabs-stdlib/pull/1307#discussion_r1489325529

alexjfisher:
@binford2k Would be interested to hear your thoughts around this.  I dunno if I've come up with a great or terrible idea! :)

LukasAud:
I like the idea behind this. Looks like a nice QoL update to clean up functions and make things a bit more readable 👍 

Unfortunately, I can't tell right away if there is any limitations or issues with this approach but I'll try to shine some light on this PR and see if it catches some eyes.

lib/puppet/functions/stdlib/rewrap_sensitive_data.rb
```diff
@@ -0,0 +1,55 @@
...
+      result
+    end
+  end
+
+  def deep_unwrap(obj)
+    case obj
+    when Hash
+      obj.each_with_object({}) do |(key, value), result|
+        result[key] = deep_unwrap(value)
+      end
```
ekohl, line 45: Probably unlikely that you want to unwrap keys, so you can also use [transform_values](https://ruby-doc.org/core-2.7.1/Hash.html#method-i-transform_values) for shorter code:
```suggestion
      obj.transform_values { |value| deep_unwrap(value) }
```

lib/puppet/functions/stdlib/rewrap_sensitive_data.rb
```diff
@@ -0,0 +1,55 @@
...
+    optional_block_param 'Callable[Any]', :block
+    return_type 'Any'
+  end
+
+  def rewrap_sensitive_data(data)
+    @contains_sensitive = false
+
+    unwrapped = deep_unwrap(data)
+
+    result = block_given? ? yield(unwrapped) : unwrapped
```
ekohl, line 31: Should the yielded `unwrapped` actually return a `Sensitive`. Or in other words, should it be `yield(wrapped)` instead?

lib/puppet/functions/stdlib/rewrap_sensitive_data.rb
alexjfisher: Thanks.  If anyone ever does come up with a situation where they've ended up with a sensitive(string) as a hash key we can always change it back and add this functionality.

lib/puppet/functions/stdlib/rewrap_sensitive_data.rb
alexjfisher: No, I think this is correct. The block, (if provided), needs to be handed the unwrapped data so that it can mutate it as desired (eg. serialising to yaml or json), before this function then wraps the result back up with sensitive.

I've added the first spec tests that hopefully illustrate this a bit better.

lib/puppet/functions/stdlib/rewrap_sensitive_data.rb
alexjfisher: I've had a rethink! :)  I agree it's probably unlikely to come across sensitive keys, but not impossible or inherently wrong. It's also easy enough to code for, so I've added a test and now handle this case.

binford2k:
I really like the rewrap-with-a-block idea. I hate that we have to do it. 🤣

b4ldr (approved):
lgtm

PR #1419, opened 2024-03-01
Addition of stdlib::getpwnam to return `/etc/passwd` entry.

traylenator:
## Summary
Given a username `stdlib::getpwnam` returns the password entry for that user.

Example

```puppet
$passwd_entry = stdlib::getpwnam('root')
```

then `$passwd_entry` would contain:

```puppet
{
  'name'=>'root',
  'passwd'=>'x',
  'uid'=>0,
  'gid'=>0,
  'gecos'=>'root',
  'dir'=>'/root',
  'shell'=>'/bin/bash',
}
```

## Checklist
- [x] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.

```puppet
$passwd = stdlib::getpwnam('steve')
notify{"${passwd}":}
notify{"${passwd['uid']}":}
```
results in:

```
Notice: {name => steve, passwd => x, uid => 1000, gid => 1000, gecos => Steve, dir => /home/steve, shell => /bin/bash}
Notice: /Stage[main]/Main/Notify[{name => steve, passwd => x, uid => 1000, gid => 1000, gecos => Steve, dir => /home/steve, shell => /bin/bash}]/message: defined 'message' as '{name => steve, passwd => x, uid => 1000, gid => 1000, gecos => Steve Traylen, dir => /home/steve, shell => /bin/bash}'
Notice: 1000
Notice: /Stage[main]/Main/Notify[1000]/message: defined 'message' as '1000'
```


bastelfreak (approved):

lib/puppet/functions/stdlib/getpwnam.rb
```diff
@@ -0,0 +1,27 @@
...
+  #
+  # @return
+  #   [Hash] For example {"name"=>"root", "passwd"=>"x", "uid"=>0, "gid"=>0, "gecos"=>"root", "dir"=>"/root", "shell"=>"/bin/bash"}
+  dispatch :getpwnam do
+    param 'String', :user
+    return_type 'Hash'
+  end
+
+  def getpwnam(user)
+    Etc.getpwnam(user).to_h.first(7).map { |k, v| { k.to_s => v } }.reduce(:merge)
```
ekohl, line 25: If the user isn't found this will probably end up being `nil.to_h`. Also, why do you even need to do the map? Isn't this sufficient?
```suggestion
    Etc.getpwnam(user)&.to_h
```

I'd expect Puppet to translate the symbols to strings.

ekohl (commented):
Perhaps a note about Windows (in)compatibility?

lib/puppet/functions/stdlib/getpwnam.rb
traylenator: Without just `Etc.getpwnam(user).to_h` you get a Tuple rather than a Hash back - changing the return to Tuple 
shows this:
```
       expected stdlib::getpwnam("steve") to have returned {"name"=>"steve", "passwd"=>"x", "uid"=>1000, "gid"=>1001, "gecos"=>"Steve", "dir"=>"/home/steve", "shell"=>"/bin/fish"} instead of [[:name, "steve"], [:passwd, "x"], [:uid, 1000], [:gid, 1001], [:gecos, "Steve"], [:dir, "/home/steve"], [:shell, "/bin/fish"]]
```
which does not seem much good.

For the no such user it fails in Etc.getpwnam itself.

```
ArgumentError(can't find user for notexisting)
```
which seemed reasonable to me.

traylenator:
> Perhaps a note about Windows (in)compatibility?

Added a note about POSIX not windows. .. I dont have a windows machine to hand to see what it does.

lib/puppet/functions/stdlib/getpwnam.rb
ekohl: > With just `Etc.getpwnam(user).to_h` you get a Tuple rather than a Hash back - changing the return to Tuple

This is odd, perhaps Ruby version dependent?

```console
$ ruby --version
ruby 3.2.2 (2023-03-30 revision e51014f9c0) [x86_64-linux]
$ ruby -retc -e "puts Etc.getpwnam('ekohl').to_h"
{:name=>"ekohl", :passwd=>"x", :uid=>1000, :gid=>1000, :gecos=>"ekohl", :dir=>"/home/ekohl", :shell=>"/bin/bash"}
```

> which seemed reasonable to me.

Sounds fair to me. Is there a `@raises`  tag in puppet-strings to document this?

ekohl:
> Added a note about POSIX not windows. .. I dont have a windows machine to hand to see what it does.

Me neither, but quoting https://ruby-doc.org/core-3.1.2/Etc.html

> The [Etc](https://ruby-doc.org/core-3.1.2/Etc.html) module provides access to information typically stored in files in the /etc directory on Unix systems.

Reading that, I think the behavior is completely undefined on Windows so I'd say it's just incompatible as you did now.

lib/puppet/functions/stdlib/getpwnam.rb
traylenator: Added the `@raises`

With `Etc.getpwnam(user).to_h` :

* ruby 3.2.2 and puppet 8.5.0    (should switch to 3.3.X I think)
* ruby 2.7.7 and puppet 7.29.0 

I get 
```
[[:name, "steve"], [:passwd, "x"], [:uid, 1000], [:gid, 1001], [:gecos, "Steve"], [:dir, "/home/steve"], [:shell, "/bin/fish"]]
````
for both.

**Comment by traylenator** [comment]
Interesting I added to the CI :
* Adding a user
* Using that user in Deferred call to exec .

On a random VM here with just an  apply the ordering is fine.
In github CI it seems the require from the exec to the user was not respected.

**Comment by traylenator** [comment]
> Interesting I added to the CI :
> 
> * Adding a user
> * Using that user in Deferred call to exec .
> 
> On a random VM here with just an apply the ordering is fine. In github CI it seems the require from the exec to the user was not respected.

It's a lack of https://puppet.atlassian.net/browse/PUP-11526 on Puppet 7 causing problems. Trying to mark it pending for 7 but totally failing. :cry:

File: spec/acceptance/getpwnam_spec.rb [hunk 2]
```diff
@@ -0,0 +1,107 @@
...
+       command     => 'echo $MY_XDG_RUNTIME_DIR > /tmp/testx-xdg-dir',
+       path        => $facts['path'],
+       creates     => '/tmp/testx-xdg-dir',
+       require     => User['testx'],
+     }
+   MANIFEST
+    end
+
+    it 'works idempotently with no errors' do
+      pending('puppet7 having https://puppet.atlassian.net/browse/PUP-11526') if Puppet.version.to_f < 8 && (os[:family] != 'windows')
```
bastelfreak, line 91: as a hack; create the user with one apply, then use the rest of the code  in another apply?

traylenator: Morning - yeah I have done that above - works fine.
Just failing with either `Puppet.version` or `fact('puppetversion')`.
I could just remove that whole test but nice to have..

traylenator:
Given the situation with Puppet 7 does it make more sense to return `undef` if user is missing.

It will much be easier to handle.

lib/puppet/functions/stdlib/getpwnam.rb
smortex: Hum, I checked on a bunch of versions of ruby and `Etc.getpwnam('romain').to_h` return a Hash for all of them (2.5, 2.7, 3.1 and 3.2).  What you get in your last message seems to be the output of `to_a` applied on a Hash :thinking:.

Also why only keep the 7 first entries? (also `#first` on a Hash return an Array :wink:) All OS do not return the same information, so I think it make sense to keep it unchanged and allow the end-user to consume what they want from the hash?

The suggested change by @ekohl seems to really be what I would expect, can you please double-check this and if you reproduce the issue show us the failing code so that we can better understand what is going on?

spec/acceptance/getpwnam_spec.rb
```diff
@@ -0,0 +1,109 @@
...
+      # Test the use case for systemd module
+      user{ 'testu':
+        ensure => present,
+        uid    => 250,
+      }
+    MANIFEST
+    end
+
+    it 'works idempotently with no errors' do
+      apply_manifest(pp, catch_failures: true) unless os[:family] == 'windows'
```
smortex, line 24: Instead of repeating the condition (`unless os[:family] == 'windows'`) on each line…

spec/acceptance/getpwnam_spec.rb
```diff
@@ -0,0 +1,109 @@
+# frozen_string_literal: true
+
+require 'spec_helper_acceptance'
+
+describe 'function stdlib::getpwnam' do
```
smortex, line 5: Better make the whole test conditional?

```suggestion
describe 'function stdlib::getpwnam', unless os[:family] == 'windows' do
```

spec/functions/getpwnam_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+      {
+        'name' => 'steve',
+        'passwd' => 'x',
+        'uid' => 1000,
+        'gid' => 1001,
+        'gecos' => 'Steve',
+        'dir' => '/home/steve',
+        'shell' => '/bin/fish'
+      },
+    )
```
smortex, line 31: Not sure if mocking is the best strategy.  I would only have a unit test (and skip it on windows) that get info for a user we are sure will exist (either we get the name of the current user or we look for `root`) and check that the returned hash include `{name: 'root'}`

traylenator:
I started adding this as it was needed for https://github.com/voxpupuli/puppet-systemd/pull/443

however the final solution used completely avoids the need for this.

Close this unless someone thinks it's really worth having.

traylenator:
This was needed to get the uid for a user to find the path `$XDG_RUNTIME_DIR=/run/user/<uid>`

This is now avoided by running an exec wrapped in `systemd-run` as per https://github.com/voxpupuli/puppet-systemd/pull/443

This function is no longer required and given its a bit of a non-trivial one lets not add it.

PR #1420, opened 2024-03-05
Use the read method on the return of open_url to actually get the con…

gerases:
## Summary
Use the read method on the return of open_url to actually get the content. Otherwise the following error is returned by puppetserver:

```
Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Error while evaluating a Function Call,
 no implicit conversion of Tempfile into String (file: <file_path>, line: 8, column: 15) on node <node_name>
```

### DIFF
--- a/lib/puppet/parser/functions/loadjson.rb
+++ b/lib/puppet/parser/functions/loadjson.rb
@@ -44,7 +44,7 @@ module Puppet::Parser::Functions
           url = args[0]
         end
         begin
-          contents = OpenURI.open_uri(url, http_options)
+          contents = OpenURI.open_uri(url, http_options).read
         rescue OpenURI::HTTPError => e
           res = e.io
           warning("Can't load '#{url}' HTTP Error Code: '#{res.status[0]}'")
--- a/spec/functions/loadjson_spec.rb
+++ b/spec/functions/loadjson_spec.rb
@@ -55,7 +55,6 @@ describe 'loadjson' do
         allow(File).to receive(:exist?).and_call_original
         allow(File).to receive(:exist?).with(filename).and_return(true).once
         allow(File).to receive(:read).with(filename).and_return(json).once
-        allow(File).to receive(:read).with(filename).and_return(json).once
         if Puppet::PUPPETVERSION[0].to_i < 8
           allow(PSON).to receive(:load).with(json).and_return(data).once
         else
@@ -98,7 +97,7 @@ describe 'loadjson' do
       let(:json) { '{"key":"value", {"ķęŷ":"νậŀųề" }, {"キー":"値" }' }
 
       it {
-        expect(OpenURI).to receive(:open_uri).with(filename, {}).and_return(json)
+        expect(OpenURI).to receive(:open_uri).with(filename, {}).and_return(StringIO.new(json))
         if Puppet::PUPPETVERSION[0].to_i < 8
           expect(PSON).to receive(:load).with(json).and_return(data).once
         else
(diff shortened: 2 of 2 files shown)


smortex (changes requested):
The test suite needs to be adjusted accordingly.  Currently it assumes `OpenURI.open_uri` return a `String` but in reality it return a `StringIO` or a `Tempfile` depending on the response. And both have a `read` method.

I am not found of doubles, but something like:

```diff
diff --git a/spec/functions/loadjson_spec.rb b/spec/functions/loadjson_spec.rb
index 17d01e45..0022ebfa 100644
--- a/spec/functions/loadjson_spec.rb
+++ b/spec/functions/loadjson_spec.rb
@@ -98,7 +98,9 @@ describe 'loadjson' do
       let(:json) { '{"key":"value", {"ķęŷ":"νậŀųề" }, {"キー":"値" }' }
 
       it {
-        expect(OpenURI).to receive(:open_uri).with(filename, {}).and_return(json)
+        data = double
+        expect(data).to receive(:read).and_return(json)
+        expect(OpenURI).to receive(:open_uri).with(filename, {}).and_return(data)
         if Puppet::PUPPETVERSION[0].to_i < 8
           expect(PSON).to receive(:load).with(json).and_return(data).once
         else
```

Run `bundle exec rake spec` to run the test suite on your machine.  Failures are reported here:

* `./spec/functions/loadjson_spec.rb:100`
* `./spec/functions/loadjson_spec.rb:120`
* `./spec/functions/loadjson_spec.rb:140`
* `./spec/functions/loadjson_spec.rb:157`

smortex:
> Yep, will do within a few days, thank you.
Thanks!

> Do you agree it's a bug though?
Yup!

gerases:
@smortex, I decided to use StringIO in the tests. I'm not sure how to restart the build though.

gerases:
Also I see there's no way to test using a docker image, which is a bit of a hassle. Is there a plan to add that similar to other repos of vox pupuli? By the way, why "pupuli" (which is incorrect) and not "populi"?

bastelfreak:
@gerases puppetlabs/stdlib uses Litmus to spin up SUTs (systems under test), whereas Vox Pupuli uses Beaker. But as Beaker, Litmus supports Docker (and Podman I think?) Containers for SUTs. I just don't know how to use Litmus.

And Vox Pupuli is correct, it's just not 'people', but 'puppet': https://latin-dictionary.net/search/english/puppet

gerases:
> And Vox Pupuli is correct, it's just not 'people', but 'puppet': https://latin-dictionary.net/search/english/puppet

Oh, fantastic then. Didn't think of "pupulus". Thanks.

smortex (approved):
LGTM!

gerases:
@bastelfreak  Could this be merged?

bastelfreak:
I'm a bit hesitant on merging because I don't know if this change breaks the setup for other users. I need to test it and I currently don't have the time.

ekohl:
I'm surprised I didn't link it, but https://github.com/puppetlabs/puppetlabs-stdlib/pull/1415 was my attempt to address it.

jst-cyr:
I did a rebase to run this through the tests again with the latest main updates (similar to #1415).  It looks like there was a general consensus that this was a bug but that folks were unsure how much this might break setups. 

@gerases : Is this still a need in your setup?

jst-cyr:
Closing this PR for now until a need rises for this in the future at which point we can re-open!

gerases:
Sorry missed your ping, I'm not sure how much it is a problem, was just trying to contribute, i think it's pretty safe to merge, but who am i to say :)

jst-cyr:
Hey, no worries, @gerases ! I don't mind reopening if this is still something that people need resolved. I do see #1460 also open that seems to have some similar overlap. Do you mind taking a look at #1460 and seeing if we could centralize on that PR as the place for us to address this issue? Between @smortex and yourself, we might be able to get a solid fix here between both your attempts.

PR #1422, opened 2024-03-26
Allow usage of file templates with stdlib::manage

tuxmea:
This one only covers epp templates.

## Summary
Users want to make use of stdlib::manage also with templates.
As hiera can not evaluate functions, we must implement this as Puppet code.

## Checklist
- [x] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.

### DIFF
--- a/manifests/manage.pp
+++ b/manifests/manage.pp
@@ -56,7 +56,7 @@ class stdlib::manage (
       case $type {
         'file': {
           if 'template' in $attributes and 'content' in $attributes {
-            fail("You can not set content and template fir file ${title}")
+            fail("You can not set 'content' and 'template' for file ${title}")
           }
           if 'template' in $attributes {
             $content = epp($attributes['template'])
--- a/manifests/manage.pp
+++ b/manifests/manage.pp
@@ -41,7 +41,13 @@
 #         notify: 'Service[sshd]'
 #       '/etc/motd':
 #         ensure: 'file'
-#         template: 'profile/motd.epp'
+#         epp:
+#           template: 'profile/motd.epp'
+#           context: {}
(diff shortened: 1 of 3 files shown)


rwaffen (approved):

templates/manage_spec.epp
```diff
@@ -0,0 +1 @@
+I am a template
```
bastelfreak, line 1: I'm not a big fan of having a template, just for the tests. But I also don't have a better idea right now :thinking:

tuxmea: Replaced with a function mock/stub

bastelfreak (dismissed):

manifests/manage.pp
```diff
@@ -46,7 +53,27 @@
...
   $create_resources.each |$type, $resources| {
     $resources.each |$title, $attributes| {
-      create_resources($type, { $title => $attributes })
+      case $type {
+        'file': {
+          if 'template' in $attributes and 'content' in $attributes {
+            fail("You can not set 'content' and 'template' for file ${title}")
+          }
+          if 'template' in $attributes {
+            $content = epp($attributes['template'])
```
ekohl, line 62: Should this respect the file extension? I can also imagine you want to pass `epp` to map to the `epp()` function and `template` to map to the `template()` function.

Other thought: you can provide context as well. This may be needed for built in templates that have parameter validation.
```yaml
stdlib::manage::create_resources
  file:
    '/etc/motd':
      epp:
        template: 'profile/motd.epp'
        context: {}
```

bastelfreak (approved):

bastelfreak (approved):

PR #1425, opened 2024-04-15
Parse deferred templates non deferred first

### LINKED ISSUE #513: Use of deferred function fails due to complex types in epp
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: 7.24.0
- Module version:

## How to reproduce (e.g Puppet code you use)

Needs to be run on a puppet server and distinct agent setup, i.e not rspec or puppet apply.

```puppet
redis::instance{String($port):
  service_enable => true,
  service_ensure => 'running',
  requirepass    => Deferred('secret::get',['redis_password']),
  port           => $port,                                      
  bind           => [], # all interfaces
}
```

## What are you seeing

It compiles fine but during the stdlib::deferable_epp part  in the apply

```
Error: Failed to apply catalog: Evaluation Error: Resource type not found: Stdlib::Absolutepath (file: inlined-epp-text, line: 3, column: 3)

```

## What behaviour did you expect instead

Should be possible to set password as a deferred to be late evaluated on the agent however it fails due to lack of rich
data types being available on the agent nodes.

## Any additional information you'd like to impart
The only way I know to "fix" this is to strip back the data types in the template to core data types e..g

`Stdlib::Absolutepath` -> `String[1]`

would be very happy to hear if other options are available.

traylenator:
## Summary
Parse templates with deferred values twice so complex data types can be used.

## Additional Context

Currently it is not possible to have a template `file.epp`

```puppet
<%- |
Stdlib::Port $port,
String[1] $password,
| %>
port <%= $port %>
password <%= $password %>

```

and run

```puppet
file{'/tmp/junk':
  content => stdlib::deferrable_epp(
      'module/file.epp',
      { 'port' => '1234', pass => Deferred('secrets::get',['mysecret'])}
   ),
}
```
since the deferred template substitution  will fail:
```
Error: Failed to apply catalog: Evaluation Error: Resource type not found: Stdlib::Port (file: inlined-epp-text, line: 2, column: 3)
```
due to `Stdlib::Port` not being available on the agent node.

This change now parses the EPP twice. The first non deferred parse will reduce the template to:

```puppet
port = 1234
password <%= $password %>
```

and this simpler template will be parsed in deferred mode.

Note the original template type for password must accept the intermediate generated value of `<%= $password %>` which is typically case for a secret password.  If the deferred value is more complicated then the argument `preparse` can be
set `false` to not attempt the epp substitution of non deferred values.

Provide a detailed description of all the changes present in this pull request.

## Related Issues (if any)
Fixes https://github.com/voxpupuli/puppet-redis/issues/513

## Checklist
- [X] 🟢 Spec tests.
- [x] 🟢 Acceptance tests.
- [X] Manually verified. Verified on a separate master and agent with puppet-redis module as per bug.

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -3173,18 +3173,38 @@ Type: Puppet Language
 This function returns either a rendered template or a deferred function to render at runtime.
 If any of the values in the variables hash are deferred, then the template will be deferred.
 
-Note: this function requires all parameters to be explicitly passed in. It cannot expect to
-use facts, class variables, and other variables in scope. This is because when deferred, we
-have to explicitly pass the entire scope to the client.
+Note: In the case where at least some of the values are deferred and preparse is `true` the template
+      is parsed twice:
+ The first parse will evalute any parameters in the template that do not have deferred values.
+ The second parse will run deferred and evaluate only the remaining deferred parameters. Consequently
+ any parameters to be deferred must accept a String[1] in original template so as to accept the value
+ "<%= $variable_with_deferred_value %>" on the first parse.
 
-#### `stdlib::deferrable_epp(String $template, Hash $variables)`
+ @param template template location - identical to epp function template location.
+ @param variables parameters to pass into the template - some of which may have deferred values.
+ @param preparse
+     If `true` the epp template will be parsed twice, once normally and then a second time deferred.
+     It may be nescessary to set `preparse` `false` when deferred values are somethig other than
+     a string
+
+#### `stdlib::deferrable_epp(String $template, Hash $variables, Boolean $preparse = true)`
 
 This function returns either a rendered template or a deferred function to render at runtime.
 If any of the values in the variables hash are deferred, then the template will be deferred.
 
-Note: this function requires all parameters to be explicitly passed in. It cannot expect to
-use facts, class variables, and other variables in scope. This is because when deferred, we
-have to explicitly pass the entire scope to the client.
+Note: In the case where at least some of the values are deferred and preparse is `true` the template
+      is parsed twice:
+ The first parse will evalute any parameters in the template that do not have deferred values.
+ The second parse will run deferred and evaluate only the remaining deferred parameters. Consequently
+ any parameters to be deferred must accept a String[1] in original template so as to accept the value
+ "<%= $variable_with_deferred_value %>" on the first parse.
+
+ @param template template location - identical to epp function template location.
+ @param variables parameters to pass into the template - some of which may have deferred values.
+ @param preparse
+     If `true` the epp template will be parsed twice, once normally and then a second time deferred.
+     It may be nescessary to set `preparse` `false` when deferred values are somethig other than
+     a string
 
 Returns: `Variant[String, Sensitive[String], Deferred]`
 
@@ -3197,6 +3217,12 @@ Data type: `String`
 ##### `variables`
 
 Data type: `Hash`
+
+
+
+##### `preparse`
+
+Data type: `Boolean`
 
 
 
--- a/functions/deferrable_epp.pp
+++ b/functions/deferrable_epp.pp
@@ -1,15 +1,39 @@
 # This function returns either a rendered template or a deferred function to render at runtime.
 # If any of the values in the variables hash are deferred, then the template will be deferred.
 #
-# Note: this function requires all parameters to be explicitly passed in. It cannot expect to
-# use facts, class variables, and other variables in scope. This is because when deferred, we
-# have to explicitly pass the entire scope to the client.
+# Note: In the case where at least some of the values are deferred and preparse is `true` the template
+#       is parsed twice:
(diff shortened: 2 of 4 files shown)


traylenator:
Very open to idea that this new feature should be behind a parameter:
```puppet
file{'/tmp/junk':
  content => stdlib::deferrable_epp(
      'module/file.epp',
      { 'port' => '1234', pass => Deferred('secrets::get',['mysecret'])},
      prerender => true,
   ),
}
```
or something?

REFERENCE.md
```diff
@@ -3173,18 +3173,22 @@ Type: Puppet Language
...
 If any of the values in the variables hash are deferred, then the template will be deferred.
 
-Note: this function requires all parameters to be explicitly passed in. It cannot expect to
-use facts, class variables, and other variables in scope. This is because when deferred, we
-have to explicitly pass the entire scope to the client.
+Note: In the case where at least some of the values are deferred the template is parsed twice:
+ The first parse will evalute any parameters in the template that do not have deferred values.
+ The second parse will run deferred and evaluate only the remaining deferred parameters. Consequently
+ any parameters to be deferred must accept a String[1] in original template so as to accept the value
+ "<%= $variable_with_deferred_value %>" on the fist parse.
```
alexjfisher, line 3180:
```suggestion
 "<%= $variable_with_deferred_value %>" on the first parse.
```

functions/deferrable_epp.pp
```diff
@@ -1,15 +1,25 @@
...
 # If any of the values in the variables hash are deferred, then the template will be deferred.
 #
-# Note: this function requires all parameters to be explicitly passed in. It cannot expect to
-# use facts, class variables, and other variables in scope. This is because when deferred, we
-# have to explicitly pass the entire scope to the client.
+# Note: In the case where at least some of the values are deferred the template is parsed twice:
+#  The first parse will evalute any parameters in the template that do not have deferred values.
+#  The second parse will run deferred and evaluate only the remaining deferred parameters. Consequently
+#  any parameters to be deferred must accept a String[1] in original template so as to accept the value
+#  "<%= $variable_with_deferred_value %>" on the fist parse.
```
alexjfisher, line 8:
```suggestion
#  "<%= $variable_with_deferred_value %>" on the first parse.
```

traylenator:
> this might be necessary, but maybe the default can be the new behaviour?

Yes someone using will a secret array type, while rare I expect, will have to do something. Set new `preparse=>false` or 
update their template to accept a string also. 

Have added `preparse` => `true` as a default.

alexjfisher:
> > this might be necessary, but maybe the default can be the new behaviour?
> 
> Yes someone using a secret array type, while rare I expect, will have to do something. Set new `preparse=>false` or update their template to accept a string also.
> 
> Have added `preparse` => `true` as a default.

That works for me.  (I think more likely than a non string type is someone using `Pattern` for cases where passwords eg. don't accept special characters.)

traylenator:
As this is it's backwards incompatible.
Changing to default false it would be compatible.

I think true is better.

Depends a bit how near we are to a major bump.

functions/deferrable_epp.pp
```diff
@@ -1,15 +1,39 @@
...
+#  @param variables parameters to pass into the template - some of which may have deferred values.
+#  @param preparse
+#      If `true` the epp template will be parsed twice, once normally and then a second time deferred.
+#      It may be nescessary to set `preparse` `false` when deferred values are somethig other than
+#      a string
+#
+function stdlib::deferrable_epp(String $template, Hash $variables, Boolean $preparse = true) >> Variant[String, Sensitive[String], Deferred] {
   if $variables.stdlib::nested_values.any |$value| { $value.is_a(Deferred) } {
+    if $preparse {
+      $_variables_escaped = $variables.map | $_var , $_value | {
```
ekohl, line 21: Like https://github.com/voxpupuli/puppet-redis/pull/515#discussion_r1568490875 this has the problem that `nested_values` does recurse into hashes while `$variables.map` doesn't. I think the inner part is essentially a Puppet native version of Ruby's `Hash#transform_values`. Perhaps what's needed is a `deep_transform_values`.

traylenator: So:
```
stdlib::deferrable_epp(template, { 'data' => { 'a' => Deferred(....) } })
```

Used to be supported and now it would  not be with this patch as it is.

ekohl: Yes. It would be good to somehow cover this in a test case.

jst-cyr:
@traylenator : It looks like there are some unresolved feedback items on this PR. Do you still need this change? Or should we close out this PR?

traylenator:
I think this is still useful.

I'll take a look in the next few days and remind myself.

jst-cyr:
Thanks @traylenator ! Just let me know when things are ready and I'll do the first pass on acceptance testing.

PR #1431, opened 2024-06-12
Revert "pdksync - (ITHELP-87329) - replace pull_request_target with pull_request"

binford2k:
Reverts puppetlabs/puppetlabs-stdlib#1429

the labeler needs this event and uses it safely


bastelfreak:
Any chance that the ITHELP project will be made public? I think it's always confusing when public code references private tickets.

malikparvez (approved):

PR #1436, opened 2024-07-28
stdlib::parsehocon: Support reading files

bastelfreak:
## Summary

This makes it possible to also parse hocon files. I'm a bit unsure if we should extend the current function or add a new load like `stdlib::loadhocon`. I'm fine with both. Let me know what you prefer and I will add some tests for it.

## Additional Context
- [ ] Root cause and the steps to reproduce. (If applicable)
- [ ] Thought process behind the implementation.

## Related Issues (if any)
Mention any related issues or pull requests.

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.

### DIFF
--- a/lib/puppet/functions/stdlib/parsehocon.rb
+++ b/lib/puppet/functions/stdlib/parsehocon.rb
@@ -4,6 +4,10 @@
 #   This function accepts HOCON as a string and converts it into the correct
 #   Puppet structure
 #
+# @param hocon_string can be an actual string of data or a path to a Hocon config file
+#
+# @param default content that will be returned in case the string isn't parseable
+#
 # @example How to parse hocon
 #   $data = stdlib::parsehocon("{any valid hocon: string}")
 #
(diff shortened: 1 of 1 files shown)


lib/puppet/functions/stdlib/parsehocon.rb
```diff
@@ -17,16 +21,18 @@
   end
 
   def parsehocon(hocon_string, default = :no_default_provided)
-    require 'hocon/config_factory'
-
-    begin
+    if File.exist? hocon_string
```
ekohl, line 24: I'm not a fan of this, because this can also run on the server. It becomes unpredictable.

Isn't there a way to read a file in puppet? I know [`file()`](https://github.com/puppetlabs/puppet/blob/main/lib/puppet/parser/functions/file.rb) can be used in some cases. Its docs mention it can also read absolute files, so I'd expect this to work:
```puppet
stdlib::parsehocon(file('/path/to/file'))
```

ekohl (changes requested):

lib/puppet/functions/stdlib/parsehocon.rb
bastelfreak: yes the idea was to read a file from a PE installation on the puppetserver. Your solution should work. I think I will propose this function to the hocon module in case that's still maintained.

jst-cyr:
Running CI tests for this PR here: https://github.com/puppetlabs/puppetlabs-stdlib/pull/1490

jst-cyr:
**Acceptance test results:** https://github.com/puppetlabs/puppetlabs-stdlib/actions/runs/32538712281

Acceptance tests are all green, except for SLES 12 which fails on provisioning just like in main. 

Next step is an engineering review when the team is available.

PR #1448, opened 2025-02-11
Fix `has_ip_address` and `has_ip_network` functions

### LINKED ISSUE #1447: The has_ip_address function no longer works
## Describe the Bug
When performing a Puppet agent run, the following is reported. 
"Error while evaluating a Function Call, Undefined variable 'interfaces'"
The location described by the error refers to a line that runs the "has_ip_address" function. Previously this worked on PE 2021 and Puppet Agent 7.30.0

## Expected Behavior
The function should check if an IP address is on the local server and return true or false.

## Steps to Reproduce
puppet agent -t

## Environment
  "name": "puppetlabs-stdlib",
  "version": "9.7.0",
  "os": RedHat 7.9
  "puppet enterprise": 2023.8.1
  "puppet agent": 8.10.0

## Additional Context
I'm guessing this is because the "interfaces" fact no longer exists, as confirmed by the "facter -p" command. OpenSource Puppet also refers to "interfaces" as a legacy fact - https://www.puppet.com/docs/puppet/7/core_facts.htm

alexjfisher:
The functions were using the old legacy function API.  This meant that when they called `has_interface_with`, they were calling the legacy implementation of that function and not the v4 API version introduced in bc218f01497da494a499a4ab5d5fadb0ac278a37.

Only in the v4 API implementation was the modern `networking` structured fact being used.  The old `parser/functions/has_interface_with.rb` version still used legacy facts that are now not included in Puppet 8 by default.

In this commit, we replace the `has_ip_address` and `has_ip_network` functions with namespaced Puppet language functions, (these functions are simple enough to not need ruby). Non-namespaced versions are added (but marked as deprecated) in `lib/puppet/functions`.

The old implementations are removed completely. This is _almost_ certainly not going to be a breaking change for anyone. (Only other legacy functions which in turn call these functions could be affected).

Fixes #1447


functions/has_ip_address.pp
```diff
@@ -0,0 +1,10 @@
+# @summary Returns true if the client has the requested IPv4 address on some interface.
+#
+# @param ip_address
+#   The IPv4 address you want to check the existence of
+# @return [Boolean] Returns `true` if the requested IP address exists on any interface.
+function stdlib::has_ip_address(
+  Stdlib::IP::Address::V4::Nosubnet $ip_address,
```
bastelfreak, line 7: Should we mark the non namesaced functions as deprecated?

functions/has_ip_address.pp
```diff
@@ -0,0 +1,10 @@
+# @summary Returns true if the client has the requested IPv4 address on some interface.
+#
+# @param ip_address
```
bastelfreak, line 3: Should we also support ipv6?

bastelfreak: Ah right, that would require updating `stdlib::has_interface_with()` to work with IpAddr objects

b4ldr (approved):

Henrik-Hansson:
I think the acceptance test breaks due to https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/754.

PR #1460, opened 2025-12-19
Fix loading JSON from URL

smortex:
Starting with Puppet 8, stdlib use the `JSON` class instead of the `PSON` class that was used before to load JSON content downloaded from an URL.

The `PSON` class allowed to pass an `IO` object and `#read` its content for parsing, but the `JSON` class mandates passing a `String`.

The `loadjson()` functions, when passed an URL, download the document with `OpenURI.open_uri` (which return a `StringIO` object) and try to parse it.  On Puppet 7, this works because `PSON` accept `IO` objects, but with Puppet 8 an exception is raised:

```
Error: Evaluation Error: Error while evaluating a Function Call, no implicit conversion of StringIO into String
```

As the test suite inadequately mocked `OpenURI.open_uri` to return a `String`, this was not catch.

Fix this by adjusting the return value in the test suite to match the actual behavior of OpenURI and calling `#read` on the `IO` object returned by `OpenURI.open_uri`.

While here, add an acceptance tests to check `loadjson` when using an URL.


bastelfreak (dismissed):

Heidistein:
Greetings!
We just migrated to openvox8, finally. This now bites us in the behind. Any chance this gets merged?
The proposed change works as expected.

jst-cyr:
@Heidistein : I think this branch is a little out of date, so it needs some rebasing and retesting, but we can take a look for sure. There is quite the backlog for reviewing so it might be some time before it gets merged up, depending on engineering availability.

smortex:
Force pushed one more time to rebase on top of main.
The CI infrastructure seems to still be unable to provision some machines to run acceptance tests.

jst-cyr:
Thank you @smortex ! I've done a comparison of the failing systems against latest nightly, and it does look like the failures here are mostly consistent with the infrastructure issues in the nightly builds.

**Possible issue (not in Nightly)**
* Debian 12 failure looks like a real issue. Rspec failure "loading JSON from an URL would apply manifest":

`Error: Evaluation Error: Error while evaluating a Function Call, no implicit conversion of nil into String (file: /root/manifest_20260909_4629_usl702.pp, line: 1, column: 9) on node 87c9f043a385`

**Not an issue**
* Debian 11 is failing to provision. We're seeing this in other modules as well and in the nightly, something that the team has to look into, but it's not caused by this PR.
* Alma Linux 8 failed with a network connectivity issue when running the test against Puppet Core 9. I re-ran it and it went green.

PR #1461, opened 2026-01-13
pw_hash: Use JVM-wide mutex to prevent race conditions

seqizz:
## Summary

On parallel runs on a single puppetserver (debian bookworm, v: 7.9.5-2+deb12u1) with 8 workers, we randomly get `Error while evaluating a Function Call, system does not support enhanced salts` even though it works most of the time.

## Additional Context

Apparently this was a race condition, so this is an attempt to add a mutex for not colliding calculations.

Disclaimer: Since I am not a fluent ruby/java person and I realize this is a security-sensitive file to touch; I'd love to see a thorough review from someone more familiar with it. This is just an attempt to share a solution which worked in our environment.

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.

### DIFF
--- a/lib/puppet/parser/functions/pw_hash.rb
+++ b/lib/puppet/parser/functions/pw_hash.rb
@@ -3,6 +3,10 @@
 #  Please note: This function is an implementation of a Ruby class and as such may not be entirely UTF8 compatible.
 #  To ensure compatibility please use this function with Ruby 2.4.0 or greater - https://bugs.ruby-lang.org/issues/10085.
 #
+
+# Mutex for MRI Ruby - native crypt(3) is not thread-safe
+PW_HASH_CRYPT_LOCK = Mutex.new unless RUBY_PLATFORM == 'java'
+
 Puppet::Parser::Functions.newfunction(:pw_hash, type: :rvalue, arity: 3, doc: <<-DOC
   @summary
     Hashes a password using the crypt function. Provides a hash usable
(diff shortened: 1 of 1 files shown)


jst-cyr:
Thank you for this contribution @seqizz ! I agree that this change likely needs a very thorough review by the engineering crew and run it through some extra tests.

I know you mentioned you manually verified in your environment. Is that something that can be captured as a test that can be run here?

seqizz:
Hi again,

Thanks for checking! I have prepared [a ruby script](https://paste.gurkan.in/pw_has_compare.txt) to reproduce the problem, didn't take too long to see:

Pre-patched version:
```
~> puppetserver ruby pw_hash_compare.rb MODULE_ROOT=/etc/puppet/code/modules-versioned/stdlib/32ca52a1d6b8a9659c6ec69ebcb8f756a85634c3
...[default startup outputs]...

module:    32ca52a1d6b8a9659c6ec69ebcb8f756a85634c3
file md5:  694e6c8b92abed82289fd1f09b6235be
platform:  java
call:      pw_hash("password", "bcrypt", "05$salt.salt.salt.salt.sa")
reference: "$2b$05$salt.salt.salt.salt.sO5QUgeeLRANZyvfNiKJW5amLo3cVD8nW"
load:      10 threads x 20 iterations

wrong-hash: 51/200
errored:    146/200
  sample: wrong hash: "$2b$05$salt.salt.salt.salt.sOjkc.luAb5We4crI/SH0KENHz1sAGti6"

FAIL: 197/200 concurrent calls corrupted!
```

After change:
```
~> puppetserver ruby pw_hash_compare.rb MODULE_ROOT=/etc/puppet/code/modules-versioned/stdlib/10808dc1a39801a6a8bb2587b909d69da31ae559
...[default startup outputs]...

module:    10808dc1a39801a6a8bb2587b909d69da31ae559
file md5:  f8f4fbcc212cd4c2f39e4202de21f213
platform:  java
call:      pw_hash("password", "bcrypt", "05$salt.salt.salt.salt.sa")
reference: "$2b$05$salt.salt.salt.salt.sO5QUgeeLRANZyvfNiKJW5amLo3cVD8nW"
load:      10 threads x 20 iterations

wrong-hash: 0/200
errored:    0/200

OK: 200 concurrent calls, all correct. No corruption found.
```

Yet I am not sure if we can inject this to this repo, I couldn't find a way to run this on my local machine without puppetserver installed. Hints appreciated, if the testing method looks plausible :grimacing:

PR #1470, opened 2026-04-24
Add option to provide inline epp templates

jcpunk:
## Summary
Provide a way to specify inline epp template.

Example use case:
I want to add a snippet like "`profiles: <%= $profiles.join(',') %>`". It feels a bit silly to build and deploy a separate template file when I just want such a simple template.

## Additional Context
- [x] Root cause and the steps to reproduce. (If applicable)
- [x] Thought process behind the implementation.

## Related Issues (if any)
None

## Checklist
- [x] 🟢 Spec tests.
- [x] 🟢 Acceptance tests.

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -280,6 +280,13 @@ class { 'stdlib::manage':
           'template'      => 'profile/motd.epp',
         }
       },
+      '/etc/inline'       => {
+        'ensure'          => 'file',
+        'epp'             => {
+          'inline'        => 'Hello <%= $name %>',
+          'context'       => { 'name' => 'world' },
+        }
+      },
       '/etc/information'  => {
         'ensure'          => 'file',
         'erb'             => {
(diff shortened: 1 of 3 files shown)


jst-cyr:
@jcpunk : Thanks for the PR! This looks like it could be helpful but am I correct in understanding this is a Developer Experience type of issue, correct? There is currently a way to accomplish this, but it is cumbersome for very simple changes?

I'm helping out with prioritization of PR reviews, so this type of info can help guide where it winds up sitting in the list.

My initial scan on the changes look like this is helpful and a minimal change, but it does seem like this type of change could introduce Hiera doing more code execution directly without proper validation.

Have you investigated an alternative that could avoid full code injection? Or what other safety/security considerations have been put in to ensure this is a safe introduction?

jcpunk:
This is indeed a developer experience issue. There is a silly workaround of making a "epptemplates" module I'm using to accomplish the same thing, but that results in having a mix of templates that don't have any obvious relationships.

I'm not worried for my workflows about any code injections, these are rendering templates and I'm not sure `epp` can actually execute things?

Issue #1478, opened 2026-07-16
`validate_legacy` fails catalog compilation under Puppet 8 strict deprecation handling


alex-harvey-z3q:
# `validate_legacy` fails catalog compilation under Puppet 8 strict deprecation handling

## Summary

`validate_legacy()` still calls stdlib's `deprecation()` helper directly. Under Puppet 8 / OpenVox 8, where `strict=error` is the default behaviour, this deprecation call can raise a hard evaluation error before `validate_legacy()` performs its actual type assertion.

This appears to be an uncovered `validate_legacy()` instance of the deprecation/Puppet 8 compatibility issue tracked in #1391.

Related issues and PRs:

- #1391: The `deprecation` function and Puppet 8 are not compatible
- #1373: Calls to `Deprecation` function cause catalog compilation to fail if `strict` setting is set to `error`
- #1377: Ignore Puppet's `strict` setting when calling function without namespace
- #1378: Add `use_strict_setting` parameter to `deprecation` function

The previous fixes addressed namespaced/non-namespaced function shims, but `validate_legacy()` still has a direct `deprecation()` call and can still fail.

## Affected Code

Current `validate_legacy()` calls `deprecation()` before doing the assertion:

```ruby
def validate_legacy(_scope, target_type, _function_name, value, *_prev_args)
  call_function('deprecation', 'validate_legacy', 'This method is deprecated, please use Puppet data types to validate parameters')
  if assert_type(target_type, value)
    # "Silently" passes
  else
    inferred_type = Puppet::Pops::Types::TypeCalculator.infer_set(value)
    error_msg = Puppet::Pops::Types::TypeMismatchDescriber.new.describe_mismatch("validate_legacy(#{target_type}, ...)", target_type, inferred_type)
    call_function('fail', error_msg)
  end
end
```

## Expected Behaviour

`validate_legacy()` should preserve its legacy compatibility behaviour:

- valid values should pass silently
- invalid values should fail with the existing type mismatch error
- the fact that `validate_legacy()` is deprecated should not itself make catalog compilation fail under Puppet 8

## Actual Behaviour

The direct call to stdlib `deprecation()` can raise under Puppet 8 strict deprecation handling, so otherwise valid catalogs fail before `validate_legacy()` evaluates the supplied value.

PR #1479, opened 2026-07-16
Issue #1478 Avoid validate_legacy deprecation failure on Puppet 8

alex-harvey-z3q:
## Summary

Fixes the deprecated `validate_legacy()` function so that it emits a deprecation warning, as intended, and no longer lets Puppet's global `strict` setting turn that warning into a catalog failure.

Note that the stdlib `deprecation` helper accepts an optional fourth argument named `use_strict_setting`. It defaults to `true`, which means the helper follows `Puppet.settings[:strict]`:

- `:warning` logs a warning
- `:error` raises an error
- `:off` suppresses the message

This PR changes the `validate_legacy()` call to pass `false` for that argument:

```ruby
call_function('deprecation', 'validate_legacy', 'This method is deprecated, please use Puppet data types to validate parameters', false)
```

With `use_strict_setting` set to `false`, `validate_legacy()` still calls the deprecation helper and still emits the warning, but the warning is always handled as a soft deprecation warning instead of being promoted to an error under strict mode.

That preserves the migration signal for module maintainers while avoiding a hard failure before `validate_legacy()` can perform its type assertion.

## Existing stdlib precedent

This matches the pattern already used by stdlib's generated unnamespaced compatibility shims. Those functions are deprecated, emit a deprecation warning, and then delegate to the namespaced replacement while passing `false` to `deprecation()` so the compatibility path remains usable under strict mode.

Existing examples include:

- [`merge`](https://github.com/puppetlabs/puppetlabs-stdlib/blob/main/lib/puppet/functions/merge.rb#L12)
- [`ensure_packages`](https://github.com/puppetlabs/puppetlabs-stdlib/blob/main/lib/puppet/functions/ensure_packages.rb#L10)
- [`batch_escape`](https://github.com/puppetlabs/puppetlabs-stdlib/blob/main/lib/puppet/functions/batch_escape.rb#L11)
- [`fqdn_rand_string`](https://github.com/puppetlabs/puppetlabs-stdlib/blob/main/lib/puppet/functions/fqdn_rand_string.rb#L11)
- [`fqdn_rotate`](https://github.com/puppetlabs/puppetlabs-stdlib/blob/main/lib/puppet/functions/fqdn_rotate.rb#L11)
- [`has_interface_with`](https://github.com/puppetlabs/puppetlabs-stdlib/blob/main/lib/puppet/functions/has_interface_with.rb#L11)
- [`has_ip_address`](https://github.com/puppetlabs/puppetlabs-stdlib/blob/main/lib/puppet/functions/has_ip_address.rb#L11)
- [`has_ip_network`](https://github.com/puppetlabs/puppetlabs-stdlib/blob/main/lib/puppet/functions/has_ip_network.rb#L11)
- [`os_version_gte`](https://github.com/puppetlabs/puppetlabs-stdlib/blob/main/lib/puppet/functions/os_version_gte.rb#L11)
- [`parsehocon`](https://github.com/puppetlabs/puppetlabs-stdlib/blob/main/lib/puppet/functions/parsehocon.rb#L11)
- [`powershell_escape`](https://github.com/puppetlabs/puppetlabs-stdlib/blob/main/lib/puppet/functions/powershell_escape.rb#L11)
- [`seeded_rand`](https://github.com/puppetlabs/puppetlabs-stdlib/blob/main/lib/puppet/functions/seeded_rand.rb#L11)
- [`seeded_rand_string`](https://github.com/puppetlabs/puppetlabs-stdlib/blob/main/lib/puppet/functions/seeded_rand_string.rb#L11)
- [`shell_escape`](https://github.com/puppetlabs/puppetlabs-stdlib/blob/main/lib/puppet/functions/shell_escape.rb#L11)
- [`to_json`](https://github.com/puppetlabs/puppetlabs-stdlib/blob/main/lib/puppet/functions/to_json.rb#L11)
- [`to_json_pretty`](https://github.com/puppetlabs/puppetlabs-stdlib/blob/main/lib/puppet/functions/to_json_pretty.rb#L11)
- [`to_python`](https://github.com/puppetlabs/puppetlabs-stdlib/blob/main/lib/puppet/functions/to_python.rb#L11)
- [`to_ruby`](https://github.com/puppetlabs/puppetlabs-stdlib/blob/main/lib/puppet/functions/to_ruby.rb#L11)
- [`to_toml`](https://github.com/puppetlabs/puppetlabs-stdlib/blob/main/lib/puppet/functions/to_toml.rb#L11)
- [`to_yaml`](https://github.com/puppetlabs/puppetlabs-stdlib/blob/main/lib/puppet/functions/to_yaml.rb#L11)
- [`type_of`](https://github.com/puppetlabs/puppetlabs-stdlib/blob/main/lib/puppet/functions/type_of.rb#L11)
- [`validate_domain_name`](https://github.com/puppetlabs/puppetlabs-stdlib/blob/main/lib/puppet/functions/validate_domain_name.rb#L11)
- [`validate_email_address`](https://github.com/puppetlabs/puppetlabs-stdlib/blob/main/lib/puppet/functions/validate_email_address.rb#L11)

The specs have also been updated to assert that `validate_legacy()` continues to call the deprecation helper with `use_strict_setting` set to `false`.

## Checklist

- [x] Spec tests.
- [ ] Acceptance tests.
- [x] Manually verified.

### DIFF
--- a/lib/puppet/functions/validate_legacy.rb
+++ b/lib/puppet/functions/validate_legacy.rb
@@ -52,6 +52,7 @@ Puppet::Functions.create_function(:validate_legacy) do
   end
 
   def validate_legacy(_scope, target_type, _function_name, value, *_prev_args)
+    call_function('deprecation', 'validate_legacy', 'This method is deprecated, please use Puppet data types to validate parameters', false)
     if assert_type(target_type, value)
       # "Silently" passes
     else
--- a/lib/puppet/functions/validate_legacy.rb
+++ b/lib/puppet/functions/validate_legacy.rb
@@ -52,7 +52,6 @@ Puppet::Functions.create_function(:validate_legacy) do
   end
 
   def validate_legacy(_scope, target_type, _function_name, value, *_prev_args)
-    call_function('deprecation', 'validate_legacy', 'This method is deprecated, please use Puppet data types to validate parameters')
     if assert_type(target_type, value)
       # "Silently" passes
     else
--- a/spec/functions/validate_legacy_spec.rb
+++ b/spec/functions/validate_legacy_spec.rb
@@ -8,8 +8,8 @@ if Puppet::Util::Package.versioncmp(Puppet.version, '4.4.0') >= 0
     it { is_expected.to run.with_params.and_raise_error(ArgumentError) }
 
     describe 'when passing the type assertion' do
-      it 'passes without a deprecation warning' do
-        expect(subject.func).not_to receive(:call_function).with('deprecation', anything, anything)
+      it 'passes with a deprecation warning that does not honour strict mode' do
+        expect(subject.func).to receive(:call_function).with('deprecation', 'validate_legacy', include('deprecated'), false).once
         expect(scope).not_to receive(:function_validate_foo)
         expect(Puppet).not_to receive(:notice)
         expect(subject).to run.with_params('Integer', 'validate_foo', 5)
@@ -18,7 +18,7 @@ if Puppet::Util::Package.versioncmp(Puppet.version, '4.4.0') >= 0
 
     describe 'when failing the type assertion' do
       it 'fails with a helpful message' do
-        expect(subject.func).not_to receive(:call_function).with('deprecation', anything, anything)
+        expect(subject.func).to receive(:call_function).with('deprecation', 'validate_legacy', include('deprecated'), false).once
         expect(scope).not_to receive(:function_validate_foo)
         expect(subject.func).to receive(:call_function).with('fail', 'validate_legacy(Integer, ...) expects an Integer value, got String').once
         expect(subject).to run.with_params('Integer', 'validate_foo', '5')
@@ -27,7 +27,7 @@ if Puppet::Util::Package.versioncmp(Puppet.version, '4.4.0') >= 0
 
     describe 'when passing in undef' do
       it 'passes without failure' do
-        expect(subject.func).not_to receive(:call_function).with('deprecation', anything, anything)
+        expect(subject.func).to receive(:call_function).with('deprecation', 'validate_legacy', include('deprecated'), false).once
         expect(scope).not_to receive(:function_validate_foo)
         expect(Puppet).not_to receive(:notice)
         expect(subject).to run.with_params('Optional[Integer]', 'validate_foo', :undef)
@@ -35,8 +35,8 @@ if Puppet::Util::Package.versioncmp(Puppet.version, '4.4.0') >= 0
     end
 
     describe 'when passing in multiple arguments' do
-      it 'passes without a deprecation message' do
-        expect(subject.func).not_to receive(:call_function).with('deprecation', anything, anything)
+      it 'passes with a deprecation message that does not honour strict mode' do
+        expect(subject.func).to receive(:call_function).with('deprecation', 'validate_legacy', include('deprecated'), false).once
         expect(scope).not_to receive(:function_validate_foo)
         expect(Puppet).not_to receive(:notice)
         expect(subject).to run.with_params('Optional[Integer]', 'validate_foo', :undef, 1, 'foo')
--- a/spec/functions/validate_legacy_spec.rb
+++ b/spec/functions/validate_legacy_spec.rb
@@ -8,8 +8,8 @@ if Puppet::Util::Package.versioncmp(Puppet.version, '4.4.0') >= 0
     it { is_expected.to run.with_params.and_raise_error(ArgumentError) }
 
     describe 'when passing the type assertion' do
-      it 'passes with a deprecation warning' do
-        expect(subject.func).to receive(:call_function).with('deprecation', 'validate_legacy', include('deprecated')).once
+      it 'passes without a deprecation warning' do
+        expect(subject.func).not_to receive(:call_function).with('deprecation', anything, anything)
         expect(scope).not_to receive(:function_validate_foo)
         expect(Puppet).not_to receive(:notice)
         expect(subject).to run.with_params('Integer', 'validate_foo', 5)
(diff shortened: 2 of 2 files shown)


alex-harvey-z3q:
@CLAassistant Note that I changed my GitHub handle from alexharv074 to alex-harvey-z3q which could explain why it shows me as having not accepted the CLA. I've accepted again.

alex-harvey-z3q:
The failing build does not appear related to my PR.

jst-cyr:
@alex-harvey-z3q : Thanks for putting in this PR. If I'm reading your request correctly, you have a scenario where you are running on strict mode so the deprecation on validate_legacy is causing a full error on other modules.

Have you identified which module(s) in your dependency tree are still calling validate_legacy (or the older validate_* functions) directly? If it's something you maintain, updating those modules seems like the more durable fix than patching stdlib's runtime behavior. If it's a third-party module you don't control, perhaps we can get PRs in on those modules to fix their use of a deprecated function.

`stdlib` has had this method deprecated for three years, and the fix here would actually remove deprecation warnings for non-strict (not only strict) mode meaning it would lose the messaging to module maintainers that they need to update.

alex-harvey-z3q:
@jst-cyr Yes I see your point. 

How is this updated version of the PR that keeps the `validate_legacy` deprecation warning, but passes `false` for the strict-setting flag when calling `deprecation`. I think that matches the pattern already used?

E.g. [lib/puppet/functions/validate_legacy.rb (line 55)](https://github.com/puppetlabs/puppetlabs-stdlib/blob/main/lib/puppet/functions/validate_legacy.rb#L55)

I will also trace the dependency tree that is still calling `validate_legacy`/older `validate_*` functions, but hopefully this change should avoid breaking strict-mode Puppet 8 consumers while those downstream fixes make their way through?

smortex:
Given these functions where deprecated in 9.0.0, I would have expected them to be removed in 10.0.0 (but they have not).  I really do _not_ think that making them less "harmful" is a move in the right direction.

alex-harvey-z3q:
> Given these functions where deprecated in 9.0.0, I would have expected them to be removed in 10.0.0 (but they have not). I really do _not_ think that making them less "harmful" is a move in the right direction.

I have different expectations because this is enterprise software. I would expect enterprise software not to have bugs in it, and I would expect the maintainers of enterprise software to understand that enterprises move slowly.

alexjfisher:
> Given these functions where deprecated in 9.0.0, I would have expected them to be removed in 10.0.0 (but they have not). I really do _not_ think that making them less "harmful" is a move in the right direction.

By removing the function, you make it **harder** for people to upgrade to newer stdlib.
By not accepting changes like this, you make it harder for people to upgrade Puppet, (or make them disable strict mode at the expense of not getting its benefits elsewhere).

cvquesty (approved):
LGTM

cvquesty (commented):
I'm with Alex fisher... we do need to make attempts to keep the barrier lower and the file changes nominal if we can. I think it's important to realize, as mentioned, that we're all volunteers but at the same time this project exists because the puppet folks couldn't keep momentum nor get to the really necessary updates and changes in a timely fashion largely due to blast radius, being overwhelmed when there are indeed several changes that need to happen, thus they did nothing (or next to it...or moved painfully slow)

The conversation around appropriateness for a purpose is important, though. we can't prefer one over the other, and have to strike a medium balance here in that yes, we need to see these as enterprise software but at the same time, there is a mechanism to duplicate and expand on forge modules that are poor or not meeting one's needs.  The validate functions have been gone for quite some time now (3 years), but this represents 10 releases.  I think they can safely be let go now.

jst-cyr:
My apologies, been out of office and this conversation took off and I'm getting caught up here. I see an approval on the changes here, but the conversation seems to indicate that we SHOULDN'T be removing deprecation warnings.

From my view, I agree that we should keep these deprecated and then remove them completely in the next major version (it was not done in stdlib 10 to keep things stable for that release). 

So it seems to me that the correct fix for the situation that @alex-harvey-z3q finds themself into is updating the references to the old validate functions instead of changing stdlib to avoid strict mode errors?

alex-harvey-z3q:
@jst-cyr Thanks, and we already agreed that we should not remove the deprecation warning.

That is not what the updated PR is doing, and that is not what the conversation above is about. The PR still calls the stdlib `deprecation` helper, so the warning remains visible to module maintainers. The change is only to pass `false` for the helper's `use_strict_setting` argument, so this particular compatibility function continues to warn without Puppet strict mode promoting the warning to a hard catalog failure.

I have updated the PR description to make that clearer and added links to all of the (many) existing stdlib compatibility shims that already use the same `deprecation(..., false)` pattern.

jst-cyr:
@alex-harvey-z3q : Thank you, I definitely misread the flow of the conversation while trying to catch up. I see now the changes have been updated to match with other deprecations in the module instead of the original changes that had proposed removing the deprecation warning altogether.

I had already added the PR into the list for engineering review, but given that the change has been reduced and scope is minimal and resolves a bug I've boosted it a bit versus some other changes in the backlog.

alex-harvey-z3q:
Thank you I appreciate it @jst-cyr !

jst-cyr:
CI test run created: https://github.com/puppetlabs/puppetlabs-stdlib/pull/1488

jst-cyr:
Acceptance test results: https://github.com/puppetlabs/puppetlabs-stdlib/actions/runs/31842149553?pr=1488

All platforms pass except SLES-12, which fails during environment provisioning (`PuppetLitmus::RakeHelper::LitmusTimeoutError` which seems to be an SSH connection refused/timeout to the test host). This is not related to your change. The same failure occurs on the current nightly build against main, so it's a pre-existing infrastructure issue, not something introduced by this PR.

PR still needs engineering review, but acceptance tests seem good.

PR #1484, opened 2026-07-30
feat: add support for ubuntu2604

Tonguechaude:
## Summary
This PR add support for ubuntu26.04 

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.


jst-cyr:
@Tonguechaude : I've let the CI spec tests run and it looks good. We are working through publishing support for Ubuntu 26.04 on server and agent right now so this is helpful. We'll need engineering to run some acceptance testing of the module against 26.04 before it can be approved as supported, though.

PR #1488, opened 2026-08-14
Issue #1478 Avoid validate_legacy deprecation failure on Puppet 8

jst-cyr:
This PR is created to execute tests for the original PR (https://github.com/puppetlabs/puppetlabs-stdlib/pull/1479)


jst-cyr:
All platforms pass except SLES-12, which fails during environment provisioning (`PuppetLitmus::RakeHelper::LitmusTimeoutError` which seems to be an SSH connection refused/timeout to the test host). This is not related to your change. The same failure occurs on the current nightly build against main, so it's a pre-existing infrastructure issue, not something introduced by this PR.

Feedback posted on the original PR. Closing this temporary test branch.

PR #1490, opened 2026-08-21
stdlib::parsehocon: Support reading files

jst-cyr:
This PR is created to execute tests for the original PR (https://github.com/puppetlabs/puppetlabs-stdlib/pull/1436)


jst-cyr:
Acceptance tests are all green, except for SLES 12 which fails on provisioning just like in main. Will provide feedback on original PR.

